PR #103, opened 2014-07-21
clarify primary sequence, startJoin and endJoin

richarddurbin:
Change variantId in VariantJoinLocation to allow it to be null.
Primary variants must have null startJoin.variantId and endJoin.variantId, and non-primary variants must have non-null startId.variantId.  
Non-primary variants can have null endJoin.variantId if they have dangling ends.  Should we allow this?
This supports natural traversal of the graph by following startJoins as you come across them passing along variants, but not endJoins (requires searching for all the startJoins incident to a parental sequence).
A couple of other minor formatting changes.


benedictpaten:
As I understand it, this is now redundant to the new way of specifying positions and joins. I'm closing this as a cleanup, unless people object.

PR #188, opened 2014-11-11
Renamed `genes` object to more appropriate `genomicFeatures`

buske:
The `genes` object is intended to represent candidate genes or variants, making it a bit of a misnomer.

A number of alternatives were discussed in the Google doc:
* `variants`
* `variantGenes`
* `genetics`
* `sequenceFeatures`
* `genomicFeatures` (most popular)

I also fixed a potentially very-confusing typo in the description of the start and end positions (which are of the variant, not the gene). This is an uncorrected relic from before the first commit and is unfortunately incorrect in the v1 version.

### DIFF
--- a/src/main/resources/avro/wip/matchmaker.avdl
+++ b/src/main/resources/avro/wip/matchmaker.avdl
@@ -109,8 +109,8 @@ record Feature {
   union { null, string } ageOfOnset = null;
 }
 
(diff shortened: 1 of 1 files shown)


buske:
We've decided that candidate genes and variants will be separated, making this PR unnecessary.

PR #189, opened 2014-11-11
Variation Reference Methods

adamnovak:
Pulls some features of both the References and Variants data models into the existing `variationReference.avdl`, and adds a `variationreferencemethods.avdl` to define an API for it.


adamnovak:
Closing this since we have #238 in now.

PR #198, opened 2014-12-04
Graph Adaptations of Variants, References, and Reads APIs

adamnovak:
This pull request seeks to merge a first shot at adapting the Variants, References, and Reads APIs to a graph reverence into the *`graph` branch*, which is not the main master branch. They will sit there until they are ready to replace the current non-graph-capable API.

This PR also includes some commits that have been merged into master since the graph branch was created, but you can ignore those.

The biggest changes are in the Variants API, which has been converted to a monoallelic model. Additionally, the References API now has support for `Reference`s branching off of other `Reference`s in the same `ReferenceSet`, and the Reads API has support for reads that are aligned to multiple `Reference`s without being chimeric (because they follow adjacencies in the graph of `Reference`s).

Closes #189.


lh3:
How could we get the genotype of a sample? The `Genotype` object only keeps the `totalCopyNumber` but not the copy number of each allele. `Call` keeps the copy number of one allele but is not linked to `Genotype`.

src/main/resources/avro/reads.avdl
```diff
@@ -177,7 +177,38 @@ record LinearAlignment {
...
+Some especially large deletions could be represented just as well as a large
+deletion in the CIGAR string, or as a chimeric alignment.
+*/
+record GraphAlignment {
+  /** The `Path` against which the read is aligned */
+  Path path;
+
+  /**
+  The mapping quality of this alignment. Represents how likely
+  the read maps to this position as opposed to other locations.
```
jeromekelleher: Can we document the scale for this quality value please?

src/main/resources/avro/references.avdl
```diff
@@ -1,24 +1,102 @@
 @namespace("org.ga4gh")
 
+/**
+We represent reference genomes, or collections therof, by sequence graphs.
+
+A sequence graph is made progressively joining new sequence pieces, which we
```
jeromekelleher: insert 'by' after 'made'.

src/main/resources/avro/variants.avdl
```diff
@@ -94,48 +147,44 @@ record CallSet {
...
 
   /**
-  The genotype of this variant call. Each value represents either the value
-  of the referenceBases field or is a 1-based index into alternateBases.
-  If a variant had a referenceBases field of "T", an alternateBases
-  value of ["A", "C"], and the genotype was [2, 1], that would mean the call
-  represented the heterozygous value "CA" for this variant. If the genotype
-  was instead [0, 1] the represented value would be "TA". Ordering of the
-  genotype values is important if the phaseset field is present.
+  What is the ID of this `Genotype`? Unique within the repository.
```
jeromekelleher: It would be better to follow the existing pattern for documenting these attributes. i.e. `/** The ID of this Genotype. Unique within the repository. */`. I find it confusing to state the documentation for these attributes in the form of questions. Please rephrase these as declarative statements. (See above the documentation for `Segment.start`)

src/main/resources/avro/variants.avdl
```diff
@@ -94,48 +147,44 @@ record CallSet {
...
+  What `Variant` are we genotyping?
   */
-  array<int> genotype = [];
+  string variantId;
 
   /**
-  If this field is present, this variant call's genotype ordering implies
-  the phase of the bases and is consistent with any other variant calls on
-  the same contig which have the same phaseset value.
+  What is the total copy number that all `Call`s in this `Variant` should sum
```
jeromekelleher: This is too informal for an API specification. Please rephrase.

src/main/resources/avro/variants.avdl
```diff
@@ -178,42 +232,181 @@ record Variant {
...
+  */
+  union { null, int } copyNumber = null;
+
+  /**
+  Expected value of copyNumber.
+  */
+  union { null, float } dosage = null;
+
+  /**
+  Probability that copy number is k: 0..n
```
jeromekelleher: Please clarify the semantics of this field.

src/main/resources/avro/variants.avdl
```diff
@@ -178,42 +232,181 @@ record Variant {
...
+
+  /**
+  Number of reads supporting this allele (DP in VCF).
+  */
+  union { null, int } readCount = null;
+
+  /**
+  Potential other evidence for the call.
+  */
+  map<string> evidence = {};
```
jeromekelleher: What is the expected use for this field?

src/main/resources/avro/variants.avdl
```diff
@@ -178,42 +232,181 @@ record Variant {
...
+(in general at some copy number) or absent in a sample. Very often it will just
+be a single segment covering the entirety of a novel `Reference`, but it is also
+common for it to be a `Segment` of a `Reference` not overlapped by another
+`Allele`'s `Segment`s, or in general to be any contiguous path through the
+sequence graph of `Reference`s.
+
+`Allele`s belong to `VariantSet`s.
+
+Remember that `Path`s cannot follow adjacencies not represented in the
+`Reference`s they use. If you want an `Allele` that spans a novel deletion, say,
```
jeromekelleher: Again, I think this is too informal for an API specification. I would say, "Note that `Path`s cannot follow adjacencies not represented in the `Reference`s they use. If  an `Allele` that spans a novel deletion is required, for example, a new `Reference` must be included in the `VariantSet` providing that adjacency, which this `Allele` can then follow."

src/main/resources/avro/variants.avdl
```diff
@@ -178,42 +232,181 @@ record Variant {
...
+  Path path;
 }
 
+/**
+The `Call`s in a `CallSet` are not explicitly ordered and oriented. This
+reflects typically the data we have about them. In many cases we can infer order
+and orientation from the reference graph on which the `Allele`s are defined, for
+example for a sequence of simple variants (SNPs, indels etc) along a reference.
+But in the general case with multiple copies of `Allele`s and rearrangements
+this is not the case. For when we do want to explicitly state the order and
```
jeromekelleher: Delete "For" here, and insert a comma after "Alleles".

src/main/resources/avro/common.avdl
```diff
@@ -45,6 +53,38 @@ record Position {
...
+  Position start;
+
+  /**
+  How many bases are covered by this segment?
+  */
+  long length;
+}
+
+/**
+A `Path` is a series of ordered positions in a set of `Reference`s. In general any
```
jeromekelleher: "A `Path` is an ordered list of `Segment's." would be more accurate here.

src/main/resources/avro/references.avdl
```diff
@@ -27,8 +105,22 @@ record Reference {
...
+  May be null only for a primary sequence that is not joined onto any other.
+  Note that the referenceId for start and end may not be the same.
+  */
+  union { null, ReferenceJoinLocation } startJoin, endJoin;
+
+  /**
+  The sequence to insert between startJoin and endJoin. Must be set if sourceURI
+  is null.
+  */
+  union { null, string } sequence = null;
```
jeromekelleher: I think this model for getting the sequence of a `Reference` has some issues. As I understand it, here either we have the sequence embedded in the `Reference` object (which might be many megabytes) or we provide a URI to the source FASTA file. Presumably we provide this URI when the reference sequence is too long to embed within the `Reference` itself? This creates all sorts of problems in my view: do we expect the client to download and parse the FASTA file out-of-band? 

I think a better approach would be to remove both `sequence` and `sourceURI` from this object, and to provide a `GET` method `getSequence(referenceId)`, which returns the sequence as a string. Problems associated with long strings can be handled using HTTP Streaming, rather than complicating the protocol with paging code for the sequences. (Any thoughts here @calbach and @cassiedoll?)

jeromekelleher:
I'm not sure if `Reference` and `ReferenceSet` have been completely translated into the Graph world. For example, what does `isDerived` mean in the context of a general `Reference` (i.e., a string of sequence grafted into the graph somewhere)? Also, I don't think `sourceURI` has any real meaning for a complex graph ReferenceSet, since it won't have been built from a FASTA file.

src/main/resources/avro/variants.avdl
```diff
@@ -61,6 +94,15 @@ record VariantSet {
...
+calls.
+*/
+enum CallSetType {
+  GENOTYPE,
+  HAPLOTYPE
+}
+
+/**
 A `CallSet` is a collection of variant calls for a particular sample.
 It belongs to a `VariantSet`. This is equivalent to one column in VCF.
```
jeromekelleher: Does a `CallSet` belong to one `VariantSet` or many? The comment here implies one, but the `variantSetIds` list below implies many. I realise this is inherited from the main branch and we may not want to change too much to make transitioning between APIs as simple as possible. It would be good to clarify this relationship as far as the Graph based model is concerned here in a comment though.

jeromekelleher:
This is great progress, thanks for all the work Adam. I'm happy with most of the changes, but I wonder if we can tidy up the relationship between what we'd think of as a 'reference' now and what a reference will be under the graph model. Here is an alternative suggestion for a subset of the class hierarchy:

- A `ReferenceSet` is a set of `Reference` objects, representing a canonical reference, e.g., GRCh38 (same as current definition).
- A `Reference` is a sequence graph providing a canonical description of the known variation in a single genomic region (for example, human chromosome 1). The sequence graph is composed of a set of `Segment`s (but does not have the IDs of these segments embedded within it). For a linear reference, there is exactly one `Segment`.
- A `Segment` is a continuous chunk of sequence, which joins with other Segments to form a graph (equivalent to your current `Reference` object).

Under this model, there's a clear mapping between the old and the new when it comes to `Reference` objects. By using the more generic name `Segment` (or `Fragment` or whatever we end up using) to refer to the building blocks of the graph, I think we make the constructions for `Allele`s a little clearer (I find the idea of adding a new piece of `Reference` for each novel allele a bit strange). 

Any thoughts?

src/main/resources/avro/variants.avdl
adamnovak: Which field in particular? The evidence field?

src/main/resources/avro/variants.avdl
adamnovak: Which field here?

src/main/resources/avro/reads.avdl
adamnovak: It's supposed to be the same as for a linear alignment: the phred-scale probability that the mapping is wrong. However, we do need to work out what exactly we mean by "wrong" in both cases.

adamnovak:
@lh3 the `Genotype` can store genotypeLikelihoods if you have them. Otherwise you have to get all the `Call`s in the `CallSet` linked to the `Variant` you want a genotype for. (In other words, a `Genotype` and the appropriate `Call`s are linked by association to the same `Variant` and `CallSet` combination.)

src/main/resources/avro/reads.avdl
adamnovak: I would leave it as is for now, make an issue/PR to fix the lack of documentation in LinearAlignment on the main branch, and then merge those changes into the graph branch when I catch it up with master. I'm shooting for a minimal pull request at the moment.

adamnovak: I've made this its own issue: #229

src/main/resources/avro/references.avdl
adamnovak: I agree that the URL-to-FASTA approach for transferring long sequences to the client leaves a lot to be desired. However, I don't think it's a problem unique to the graph implementation.

I think we should work on this problem against the main branch, and use the string-or-FASTA setup on the graph branch (selecting strings for short sequences and FASTA URLs for those that are too long) until `getSequence` is ready. I've made issue #230 for this.

src/main/resources/avro/variants.avdl
adamnovak: A `CallSet` can span `VariantSet`s, for talking about situations in which you have fed the same data to different callers and gotten different results.

I'll change the comment about it "belonging" to a `VariantSet` to something more correct, especially given the existence of two `CallSetType`s.

adamnovak:
> I'm not sure if `Reference` and `ReferenceSet` have been completely translated into the Graph world. For example, what does `isDerived` mean in the context of a general `Reference` (i.e., a string of sequence grafted into the graph somewhere)? Also, I don't think `sourceURI` has any real meaning for a complex graph ReferenceSet, since it won't have been built from a FASTA file.

You are right that `isDerived` is going to be a bit useless for many `Reference`s in the graph model. However, I'm not sure exactly what it's useful for in the main branch either. What are some examples of derived references? If a reference is derived, where do we find the URI and accession numbers of the derived sequence versus those of its parent? I think we should document a use case for it on the main branch before we decide whether we need it on `Reference`s in a graph, and I've created issue #231 to collect opinions.

As for `sourceURI`, I'll change its definition slightly so that it gives the URI from which the sequence can be obtained by the client. This gives it a meaningful interpretation for `Reference`s that weren't created by just importing a FASTA file, and provides an easy way to work around paging issues with long `sequence` strings. It also means that the client is guaranteed to be able to actually obtain the sequence of a `Reference`, which doesn't seem to be the case on the master branch: just because the API server originally fetched a FASTA from somewhere doesn't mean it's still there.

src/main/resources/avro/references.avdl
calbach: There is an existing and already defined API for getting bases (`ListReferenceBases` method). Agreed that changing that protocol is an entirely separate discussion so #230 is a good place to have that discussion; I don't think it's relevant to this PR.

This method should be used instead of embedding a potentially very long sequence here (or requiring a user to fetch a FASTA file). It may also make sense to eventually augment that method for the Graph API; seems it would be much more convenient if you could fetch a "GraphRange" via ListReferenceBases, in addition to a linear range. Else resolving the reference bases for a graph path may take several serial RPCs.

fnothaft: > There is an existing and already defined API for getting bases (ListReferenceBases method). Agreed that changing that protocol is an entirely separate discussion so #230 is a good place to have that discussion; I don't think it's relevant to this PR.

+1

adamnovak:
> This is great progress, thanks for all the work Adam. I'm happy with most of the changes, but I wonder if we can tidy up the relationship between what we'd think of as a 'reference' now and what a reference will be under the graph model. Here is an alternative suggestion for a subset of the class hierarchy:

> - A `ReferenceSet` is a set of `Reference` objects, representing a canonical reference, e.g., GRCh38 (same as current definition).
> - A `Reference` is a sequence graph providing a canonical description of the known variation in a single genomic region (for example, human chromosome 1). The sequence graph is composed of a set of `Segment`s (but does not have the IDs of these segments embedded within it). For a linear reference, there is exactly one `Segment`.
> - A `Segment` is a continuous chunk of sequence, which joins with other Segments to form a graph (equivalent to your current `Reference` object).

> Under this model, there's a clear mapping between the old and the new when it comes to `Reference` objects. By using the more generic name `Segment` (or `Fragment` or whatever we end up using) to refer to the building blocks of the graph, I think we make the constructions for `Allele`s a little clearer (I find the idea of adding a new piece of `Reference` for each novel allele a bit strange). 

> Any thoughts?

Basically, this sounds like dividing the genome into per-chromosome boxes called `Reference`s, in which `Segment`s are linked together into one graph per box. Is that right?

This would work, and depending on how the meaning of `isDerived` gets settled and the eventual (non)implementation of `getSequence` it might be the best way to go to strip superfluous fields that sequence in a `VariantSet` will never need. But it presents a couple of problems:

- How do we deal with variants that connect chromosomes?
- In a similar vein, what about things like the pseudoautosomal region on chrX and chrY?
- Can `Reference`s overlap and describe the same genomic region? Can a chromosome be split into multiple `Reference`s (maybe one for the PAR and one for the rest of Y)? If so, how are they connected?
- Does every piece of sequence in a `Reference` have to come from the same FASTA and have the same set of accessions? If not, do we have to put those fields on the `Segment`/`Fragment` again? What data would that leave in `Reference`?
- Would a `Reference` make more sense as a `Fragment` wrapper?

I'd favor leaving the overall model as is; it's the minimal change we need to enable graphs to be represented. Each existing `Reference` can just stay as is, with no parent and nothing attached. If we solve these problems and determine that we really have some useless-for-variants fields on `Reference` as implemented, we can do another refactoring.

src/main/resources/avro/variants.avdl
adamnovak: I don't think `evidence` actually does anything useful that can't be rolled into the `info` map. We could separate certain keys out into `evidence` in our spec, but at that point we would probably just want to standardize them as fields on `Call`.

I'll just cut `evidence`, and any other information you need to bring along, whether it is evidence for the call or not, will go in `info`.

src/main/resources/avro/references.avdl
adamnovak: Sorry, I kind of forgot about that method.

It seems like the Right Thing to do would be to have `sourceURI` retain its current information-only semantics. Clients should not ever go download that FASTA in the normal course of using the API. (Should we put a comment on it to that effect?)

I don't think adding a new `sequence` field is a good idea either, because we want one correct way to get the sequence (through the method call). I think extending the method to take a Path as input might be a good idea, as it would push some of the IO-intensive walking of the graph from the client to the server, but it would be a bit hard to square with keeping the method as a GET request.

So I'll just knock off the `sequence` field and note that the sequences of `Reference`s, even tiny 1-base ones used to specify SNPs, should be obtained by calls to `ListReferenceBases`?

calbach: +1

I agree that `sourceURI` should stay; perhaps a "for provenance" addendum to the comment would dissuade someone from assuming it should be used to fetch bases.

lh3:
> @lh3 the Genotype can store genotypeLikelihoods if you have them. Otherwise you have to get all the Calls in the CallSet linked to the Variant you want a genotype for. (In other words, a Genotype and the appropriate Calls are linked by association to the same Variant and CallSet combination.)

@adamnovak Thanks. I see what you mean now. I am trying to walk through a query "get the genotype of sample NA12878 at position chr1:1000". Is that something like this:

```javascript
// get callSet; suppose we know there is only one result
// Maybe we should also allow to search multiple samples in one go?
callSet = searchCallSets({"variantSetIds":["foo"],"name":"NA12878"}).callSets[0];
// get variant. Could we get two variants at the same position? Let's assume not as it is too complex.
variant = searchVariants({"referenceName":"chr1","start":9999,"end":1000,"callSetId":callSet.id}).variants[0];
// get genotype. Does the line below always return one result?
genotype = searchGenotypes({"callSetIds":[callSet.id],"variantIds":[variant.id]}).genotypes[0];
// get calls; could be multiple. Does the list also include the reference allele/call?
calls = searchCalls({"callSetIds":[callSet.id],"variantIds":[variant.id]}).calls;
// get allele sequences.
alleleIds = [];
for (call in calls) alleleIds.push(call.alleleId);
// I guess `SearchAllelesRequest` is missing an `alleleIds` request field?
// Is it always true that `calls.length == alleles.length`?
alleles = searchAlleles({"alleleIds":alleleIds}).alleles;
// get the genotype excluding uncalled alleles.
// Suppose we have a function `getAlleleSequenceById()` that returns the allele sequence.
finalGenotype = []; copyNumber = 0;
for (call in calls) {
  copyNumber += call.copyNumber;
  for (var c = 0; c < call.copyNumber; ++c)
    finalGenotype.push(getAlleleSequenceById(alleles, call.alleleId));
}
assert(copyNumber == genotype.totalCopyNumber); // this is a constraint
print(finalGenotype.join("/"));
```
There are a few additional questions which are also inlined in the code:

1. In a `VariantSet`, do we always get one `Variant` at each position? If we could get two variants, V1 and V2, I guess one of them will have totalCopyNumber==0. Right? Is it possible for two overlapping variants to share a same `Allele`?

2. When we get a list of alleles/calls, does it include the reference allele?

3. Does a 3-tuple (variantSetId,callSetId,variantId) unambiguously identify a `Genotype` object (i.e. two `Genotype` objects cannot have the same 3-tuple)? Does (callSetId,alleleId) unambiguously identify a `Call` object (i.e. for the same sample, two `Calls` cannot have the same `alleleId`)?

4. Should we allow to search multiple sample names in `SearchCallSetsRequest`? Currently we can only search one sample at a time. In `SearchAllelesRequest`, should we allow to retrieve `Allele` objects by their IDs? Currently there is no "alleleIds" field.

5. Is there a convenient way to get the allele sequence? For now, we need to go through `Allele`=>`Path`=>`Segment`=>`Position`=>`Reference` to get an allele sequence.

Finally, `Genotype.genotypeLikelihoods[]` is different from genotypes. We need single/multi-sample calling to move from likelihoods to genotypes.

lh3:
Here is another approach towards a monolithic allele model that is closer to VCF and the v0.5 APIs. Just for thought.
```cpp
record VariantRow { array<Genotype> genotypes; } // A variant row made of per-sample genotypes
record Genotype { int totCpyNum; array<Call> calls; } // A genotype made of per-allele calls
record Call { string callSetId; Path allele; int copyNumber; } // A call is a piece of sequence and its copy number in a sample
record AlleleRow {
  array<Call> calls; // An allele row made of per-sample calls (by definition)
  array<Genotype> genotypes; // optionally allow to get the associated genotypes (for response)
}
```
A `VariantRow` here is equivalent to a VCF line and a `Variant` in the v0.5 APIs. The difference is that we separate `Genotype` and per-allele `Call`. We could traverse all variants like what we do with VCF and v0.5 APIs. We could also traverse each allele with an `AlleleRow` iterator. We could remove the `genotypes` field, but it might be handy for some access patterns.

This model removes a lot of IDs and imposes an explicit hierarchy between objects that is not so obvious by linking objects with IDs. "Objects of objects" also reduces frequent ID lookups and probably helps performance.

lh3:
I was mostly talking about the monolithic model above. I realize that I also have a question on the graph model itself. Currently, the actual allele sequence is a [`Path` object](https://github.com/adamnovak/schemas/blob/4c4d273207e1d8c943bc3cd5ff1b477f895eb273/src/main/resources/avro/variants.avdl#L363), [which is](https://github.com/adamnovak/schemas/blob/4c4d273207e1d8c943bc3cd5ff1b477f895eb273/src/main/resources/avro/common.avdl#L77) a list of regions (called `Segment` in the schema) on the [`Reference`](https://github.com/adamnovak/schemas/blob/4c4d273207e1d8c943bc3cd5ff1b477f895eb273/src/main/resources/avro/common.avdl#L77) sequence. It seems that we cannot add an arbitrary allele sequence without adding a new `Reference`. Is it true? If it is, does this mean that we need to add ~30 million 1bp `Reference` objects to keep the ~30 million 1000g SNPs? Thanks.

src/main/resources/avro/reads.avdl
jeromekelleher: Makes sense, I agree with tactic of keeping the PR minimal.

src/main/resources/avro/references.avdl
jeromekelleher: This is a good plan - `ListReferenceBases` should be able to take a `Path` as an argument, which would give us the required semantics I think. +1 for discussing the issues around `ListReferenceBases` in #230.

I'm still not sure I see the point in `sourceURI`, but this applies equally to the main branch, so this should be brought up elsewhere.

jeromekelleher:
> Basically, this sounds like dividing the genome into per-chromosome boxes called  Reference s, in which  Segment s are linked together into one graph per box. Is that right?

That's it exactly. Isn't this what we do now (but the graph is a very simple one)?

>This would work, and depending on how the meaning of  isDerived  gets settled and the eventual (non)implementation of  getSequence  it might be the best way to go to strip superfluous fields that sequence in a  VariantSet  will never need. But it presents a couple of problems:
>How do we deal with variants that connect chromosomes?

I don't think there's a problem with this in principle. Both graphs are composed of (what I'll call for arguments sake here) `Fragment`s. We make a new `Fragment`, which happens to have a `Fragment` from one graph at one join and a `Fragment` from the other in the other join. We don't update the _graph_ here, because we're creating an allele, not updating the references.

>In a similar vein, what about things like the pseudoautosomal region on chrX and chrY?
Can  Reference s overlap and describe the same genomic region? Can a chromosome be split into multiple  Reference s (maybe one for the PAR and one for the rest of Y)? If so, how are they connected?

Well, under the definition the graph is just a set of `Fragments`, so there's no reason `Fragments` can't be members of several different graphs (== `Reference`). There's issues with where we define the set of `Fragments` that a graph is composed of, but maybe an API getter with paging is the way to do this? If we attach a `referenceId` attribute, then we can only belong to one `Reference`, and we'd also have to have an `alleleId` if the `Fragment` is used to define an `Allele`.

>Does every piece of sequence in a  Reference  have to come from the same FASTA and have the same set of accessions? If not, do we have to put those fields on the  Segment / Fragment  again? What data would that leave in  Reference ?

No, I would make `Fragment` as lightweight as possible. As @lh3 pointed out, we'll need an awful lot of them and they'll need to be very small to the feasible. A `Fragment` should just be an ID and two `Join`s, basically. The `Reference` is the container object, to which we attach all of the metadata and provenance information.

> I'd favor leaving the overall model as is; it's the minimal change we need to enable graphs to be represented. Each existing  Reference  can just stay as is, with no parent and nothing attached. If we solve these problems and determine that we really have some useless-for-variants fields on  Reference  as implemented, we can do another refactoring.

I'm not sure I agree. Under the model I've proposed, we leave `ReferenceSet` and `Reference` exactly as they are. We just add in a new concept for how we get the sequence bases for the reference via a `Fragment` or a `Path` through multiple fragments. Existing implementations can easily be updated to support this, even if they don't support a graph reference.

lh3:
@jeromekelleher Say we have the GRCh38 chromosomal assembly and one ~3Mbp HLA haplotype. What will your graph look like? Will adding another ~3Mbp HLA haplotype change the existing fragments in the old graph?

src/main/resources/avro/common.avdl
```diff
@@ -45,6 +53,37 @@ record Position {
...
+through the sequence graph of `Reference`s, with no novel adjacencies, can be
+represented by a `Path`.
+*/
+record Path {
+  /**
+  Require that segments abut in order, changing between `Reference`s only where they actually connect.
+  Also require that segments are maximal, i.e. that
+  consecutive elements of the array cannot be replaced by a single element.
+  */
+  array<Segment> segments;
```
fnothaft: Nit: might be a good idea to provide a default value, e.g. `array<Segment> segments = [];`

richarddurbin:
This is about the nomenclature issue.
I agree that Reference is awkward, and would be better applied to a (standard) set of sequence pieces.
There is a problem with Fragment for the name of the basic sequence piece, because it is the standard term for the piece of DNA which is put into a sequencing machine.  Read pairs come one at each end of a fragment.
I proposed on today's phone call to use the term Segment for the basic sequence piece, then switch what is now called Segment, which is an interval on the piece, to be called a SubSegment.  Adam (I think) suggests Slice, or perhaps SegmentSlice.
An alternative for the main piece is Contig.  I quite like this also.  It is an assembly term, but the concept there, of a simple linear piece of DNA, corresponds reasonably well to what we are discussing.  Would we then leave Segment as the term for an interval on a Contig? That seems OK to me.
So, I vote +1 for Contig for the primary piece,  and Segment as an interval on that.
I vote 0 for the Segment as the piece, and SegmentSlice as the interval.
And for completeness I vote -1 for Reference as the piece, and Segment as the interval (current position).
Can we get consensus?

src/main/resources/avro/common.avdl
adamnovak: I'll do that.

benedictpaten:
Adam and I were discussing. We like:

(a) Segment for what is currently called a "Reference"
(b) Subsegment for what is currently called "Segment"
(c) Creating an extra object "Reference" that would wrap a Segment (old
name Reference), and be used for describing a Segment in the ReferenceSet,
with associated meta-data, like an accession or extra flags (see
"isDerived" flag issue).

I'm also open to the word Contig in place of Segment or Reference (and
Segment as an interval, as Richard suggests), but a contig in an assembly
is assumed to be a contiguous piece of DNA without scaffold gaps - and some
of the bits of the primary reference are definitely scaffolds, not contigs.
So I'm 0 on that proposal, on reflection.

benedictpaten:
On Wed, Jan 21, 2015 at 7:21 AM, Jerome Kelleher <notifications@github.com>
wrote:

> Basically, this sounds like dividing the genome into per-chromosome boxes
> called Reference s, in which Segment s are linked together into one graph
> per box. Is that right?
>
> That's it exactly. Isn't this what we do now (but the graph is a very
> simple one)?
>

I'm -1 on this, because I think we won't end up with nice separable
connected components by chromosome, because of inter-chromosomal
rearrangements. I think in a subsequent pull-request beyond this minimal
pull request it would make sense to define the notion of a subset of a
reference set (a kind of subgraph), which would allow one to talk about a
subgraph in the context of the large coordinate system - but I think this
should be done in a secondary pull request.

benedictpaten:
On Tue, Jan 20, 2015 at 9:16 PM, Heng Li <notifications@github.com> wrote:

> Here is another approach towards a monolithic allele model that is closer
> to VCF and the v0.5 APIs. Just for thought.
>
> record VariantRow { array<Genotype> genotypes; } // A variant row made of per-sample genotypes
> record Genotype { int totCpyNum; array<Call> calls; } // A genotype made of per-allele calls
> record Call { string callSetId; Path allele; int copyNumber; } // A call is a piece of sequence and its copy number in a sample
> record AlleleRow {
>   array<Call> calls; // An allele row made of per-sample calls (by definition)
>   array<Genotype> genotypes; // optionally allow to get the associated genotypes (for response)
> }
>
> A VariantRow here is equivalent to a VCF line and a Variant in the v0.5
> APIs. The difference is that we separate Genotype and per-allele Call. We
> could traverse all variants like what we do with VCF and v0.5 APIs. We
> could also traverse each allele with an AlleleRow iterator. We could
> remove the genotypes field, but it might be handy for some access
> patterns.
>
> This model removes a lot of IDs and imposes an explicit hierarchy between
> objects that is not so obvious by linking objects with IDs. "Objects of
> objects" also reduces frequent ID lookups and probably helps performance.
>

Hi Heng, this is cool. It's great you're thinking about how to use the API!
However, it seems to me that these are convenience objects that could be
added on top of the initial minimal pull request - i.e. they don't replace
objects in the core pull request. To me they should be added in a
subsequent pull request?

fnothaft:
> (a) Segment for what is currently called a "Reference"
(b) Subsegment for what is currently called "Segment"
(c) Creating an extra object "Reference" that would wrap a Segment (old
name Reference), and be used for describing a Segment in the ReferenceSet,
with associated meta-data, like an accession or extra flags (see
"isDerived" flag issue).

I am +1 on `Reference` -> `Segment`. I'm -0 on `Segment` -> `Subsegment`. I feel like `Segment`->`Subsegment` denotes a subset relationship, e.g., a subset of a set yields another set, or a substring of a string yields another string. I'd prefer a name that was more explicit about the fact that the object just contains indices into a `Segment`/`Reference`; possibly a `SegmentSlice` or `SegmentRange`?

> I'm also open to the word Contig in place of Segment or Reference (and
Segment as an interval, as Richard suggests), but a contig in an assembly
is assumed to be a contiguous piece of DNA without scaffold gaps - and some
of the bits of the primary reference are definitely scaffolds, not contigs.
So I'm 0 on that proposal, on reflection.

As the person who suggested `Contig`, I agree, it would be a leaky abstraction. I'd also prefer to not overload current terms, as that is confusing. I motion to retract the `Contig` proposal. ;)

lh3:
Above other things, I am most interested in my question: will we need ~30 million 1bp Reference/Segment/Contig/Subsegment objects to describe the ~30 million 1000g SNPs?

benedictpaten:
Yes, absolutely, we will. That doesn't mean the back-end has to store it
that way, but yes. Do you see this is a big deal. Assuming each such record
is less than some small number of bytes I don't see this getting that big?

richarddurbin:
Yes, we will need 30 million 1bp Reference/Segment/Contig/Subsegment objects.

I don't see that as a problem - we need some sort of object to represent the sequence difference.

I do see that we need to clean things up so that one does not need to make long series of calls to get
at very simple information.

Perhaps a solution to the debate about SubSegment or Slice is to actually have the subsegment be a segment, as
Frank wants.  This removes a layer of the structure, because then a Path can be an Array of Segments (as it is now).
Remember that we need the interval to describe, for example, reference alleles.

So, a Segment would then be an atomic basic segment, with its own DNA, or a subsegment defined as an interval on a Segment.

Actually, the extreme form of this would be to also swallow up the Path object, and allow a Segment to also alternatively be an
array of consecutive Segments.   This would remove yet another layer of the structure, and make simple Alleles much easier to 
describe, as just a Segment, plus whatever additional information is wanted.

Richard

On 21 Jan 2015, at 19:29, Heng Li <notifications@github.com> wrote:

> Above other things, I am most interested in my question: will we need ~30 million 1bp Reference/Segment/Contig/Subsegment objects to describe the ~30 million 1000g SNPs?
> 
> —
> 




-- 

adamnovak:
I've put up what things would look like after the `Segment` -> `Subsegment` and `Reference` -> `Reference`/`Segment` changes.

I'm also interested in perhaps unifying `Subsegment` and `Segment`; at one point I wanted a `Segment` to be just a couple of `SegmentJoinLocation`s and a `Subsegment`, but then I realized that that would let you start a `Segment`'s coordinates at somewhere other than 0.

selewis:
I was just skimming through this thread and thus may have missed something,
but the use of allele and genotype is confusing. As diploid organisms at
any single position an individual may have two different alleles. Is the
assumption that an AlleleRow is describing just one of the chromosome? And
the genotype is the genotype of the chromosome and not of the individual
(who may be either hetero or homozygous at that position)?

Apologies if this was part of an earlier thread, but this at first reading
appears to be an unusual definition/use of the term genotype which more
conventionally refers to an individual as a whole.

src/main/resources/avro/references.avdl
```diff
@@ -1,24 +1,144 @@
...
+  The `Segment`s onto which this `Segment` is joined are called its "parents".
+  */
+  union { null, SegmentJoinLocation } startJoin, endJoin;
+
+  /**
+  ID of the `VariantSet` that the `Segment` belongs to, if any.
+
+  If null, this `Segment` has an associated `Reference`.
+  */
+  union { null, string } variantSetId = null;
```
lh3: Could a segment be shared across multiple VariantSets? This seems one of the strongholds of the unary model.

lh3:
@selewis `Call` gives the state of one chromosome. `Genotype` combines `Calls` and describes two chromosomes in case of a diploid locus, just like what we normally mean.

src/main/resources/avro/references.avdl
adamnovak: A Segment can't be shared across VariantSets unless it gets promoted into a Reference in a ReferenceSet, as it is now.

I don't like having that field on Segment at all, since it's a bit of a backwards dependency. I think the right way to go is to just have API methods that let you get the Segments a VariantSet has, and not have a reference back to the VariantSet in the Segment. We've already dropped the use case of being a storage format anyway.

adamnovak:
To elaborate, a `Call` can also give the total copy number in a sample if it's in a `GENOTYPE` `CallSet`, as opposed to a `HAPLOTYPE` `CallSet`.

src/main/resources/avro/references.avdl
lh3: By removing this field, we could have SNPs from two VariantSets (e.g. UK10K and 1000g) linked to the same `Segment`. This would seem nice. What backward dependency are you thinking of?

adamnovak:
I've done some thinking about the relationship between `Segment` and `Subsegment`, and as @richarddurbin suggested I think that they really ought to be one object. I've done that, and I've moved a few things into common.avdl to make it work.

Instead of being identified objects, `Segment`s now select subranges of sequences, with `startJoin` and `endJoin` fields to specify where those sequences are joined at their ends, if applicable. Sequences are not a new Avro type, but just strings with IDs that can be paged through with an API call.

You can specify a sequence graph as a set of `Segment`s selecting the entirety of the involved sequences and giving all the joins, and you can also specify a `Path` as an ordered list of `Segment`s (carrying only the joins followed along the path).

Each sequence can have one `Reference`, which contains a `Segment` covering the whole sequence and attaching it to other sequences with `Reference`s. A sequence's `Reference` is addressable by the sequence ID.

A `VariantSet` comes with new sequences for the new bases it needs in its `Allele`s. I scrapped the idea of having any object referencing the `VariantSet` for this. Instead I now have an API call you can do on a `VariantSet` to page through a set of `Segment`s describing the new pieces of sequence graph that it needs on top of its `ReferenceSet`. The `Allele`s will then be `Path`s through the sequence graph formed by the sequences with `Reference`s in the `ReferenceSet`, and those described in this set of `Segment`s from the `VariantSet`

I think this cleans things up tremendously. We now have just three types (`Side`, `Position`, and `Segment`) that you can use to describe a sequence graph. There's a natural correspondence between a sequence graph and a `Path` embedded in one. There's a single API endpoint for getting sequence base information. There are no more `Reference`s in the Variants API (and no more ugly ID reference to a dependent API to support it).

What do people think?

src/main/resources/avro/reads.avdl
```diff
@@ -177,7 +177,38 @@ record LinearAlignment {
...
+Note that `Path`s are restricted to following reference adjacencies.
+If you want to align your read following adjacencies that are
+not present in the `ReferenceSet` to which you are aligning, you have a
+"chimeric" alignment, and should use multiple `ReadAlignment`s and the
+supplementaryAlignment flag instead of a single `GraphAlignment`.
+
+Some especially large deletions could be represented just as well as a large
+deletion in the CIGAR string, or as a chimeric alignment.
+*/
+record GraphAlignment {
```
calbach: The current PR allows one to define a GraphAlignment which takes a Path through a non-reference (variant) segment. Is this desired behavior, or an unintended side-effect of using Segment for both References and Variants?

From my reading of the paper, and the comments in this PR, my understanding is that in a Graph-based model, Reads (for example) would still be aligned to a reference set, and not to a reference set + variant set(s).

benedictpaten: On Wed, Jan 21, 2015 at 9:30 PM, CH Albach <notifications@github.com> wrote:

> In src/main/resources/avro/reads.avdl
> <https://github.com/ga4gh/schemas/pull/198#discussion_r23356303>:
>
> > @@ -177,7 +177,38 @@ record LinearAlignment {
> >  }
> >
> >  /**
> > -Each read alignment describes a linear alignment with additional information
> > +A string-to-reference-graph alignment can be represented by one CIGAR string and
> > +one `Path` through multiple `Reference`s, against which the CIGAR string is
> > +interpreted.
> > +
> > +Note that `Path`s are restricted to following reference adjacencies.
> > +If you want to align your read following adjacencies that are
> > +not present in the `ReferenceSet` to which you are aligning, you have a
> > +"chimeric" alignment, and should use multiple `ReadAlignment`s and the
> > +supplementaryAlignment flag instead of a single `GraphAlignment`.
> > +
> > +Some especially large deletions could be represented just as well as a large
> > +deletion in the CIGAR string, or as a chimeric alignment.
> > +*/
> > +record GraphAlignment {
>
> The current PR allows one to define a GraphAlignment which takes a Path
> through a non-reference (variant) segment. Is this desired behavior, or an
> unintended side-effect of using Segment for both References and Variants
>
From my reading of the paper, and the comments in this PR, my understanding
> is that in a Graph-based model, Reads (for example) would still be aligned
> to a reference set, and not to a reference set + variant set(s).
>

Good question. I can imagine there are times when you want to align a read
to a segment not in the reference set - for example for reads that are
otherwise not mappable, and where the extra segment serves the purpose of
extending the reference. IMO, either we should (1) make it a lightweight
operation to extend a reference set to include such extra segments, (2)
permit the case you point out, where you can align to a segment in a
variant set - which is a bit odd (I think I'm -1 on this), or (3) (probably
preferred, so 1+) allow extra segments to be included with the read
alignments, as we do for variant sets, but for reads.

benedictpaten:
On Wed, Jan 21, 2015 at 5:40 PM, adamnovak <notifications@github.com> wrote:

> I've done some thinking about the relationship between Segment and
> Subsegment, and as @richarddurbin <https://github.com/richarddurbin>
> suggested I think that they really ought to be one object. I've done that,
> and I've moved a few things into common.avdl to make it work.
>
> Instead of being identified objects, Segments now select subranges of
> sequences, with startJoin and endJoin fields to specify where those
> sequences are joined at their ends, if applicable. Sequences are not a new
> Avro type, but just strings with IDs that can be paged through with an API
> call.
>
> You can specify a sequence graph as a set of Segments selecting the
> entirety of the involved sequences and giving all the joins, and you can
> also specify a Path as an ordered list of Segments (carrying only the
> joins followed along the path).
>
> Each sequence can have one Reference, which contains a Segment covering
> the whole sequence and attaching it to other sequences with References. A
> sequence's Reference is addressable by the sequence ID.
>

Hi Adam, I thought we agreed it would be good to allow a reference segment
to not necessarily cover the entirety of the underlying sequence - for
example when specifying a reference set for a subregion, like the MHC or
one of the other HGVM pilot regions?

jeromekelleher:
Boring git engineering stuff: I think we need to rebase off upstream/master again Adam. Unfortunately, there's been some structural changes that have happened since you made this branch (#209 made some new files and added namespaces). So, some things like `GAException` moving into methods.avdl have happened. This is going to be tedious to resolve, but it's only going to get worse the longer we leave it, unfortunately.

src/main/resources/avro/common.avdl
```diff
@@ -17,31 +29,88 @@ error GAException {
...
-// TODO: Add support here for universally unique base IDs.
 record Position {
   /**
-  The name of the reference (or, more technically, the scaffold) in whatever
-  reference set is being used. Does not generally include a "chr" prefix, so for
-  example "X" would be used for the X chromosome.
+  The ID of the sequence on which the `Position` is located.
   */
-  string referenceName;
+  string id;
```
jeromekelleher: Do we  need this back-reference? We're just embedding these into `Segment` objects, and there's no reason to look them up independently. `Position` should be as lightweight as possible, so I would vote to remove this.

We're running into major problems with backwards compatibility here though. By changing the `Position` type, the old protocol won't work any more. Perhaps we should leave  `Position` alone and use a new `Location` object? We could then slowly deprecate the old reference interface?

src/main/resources/avro/common.avdl
```diff
@@ -17,31 +29,88 @@ error GAException {
...
+*/
+record Path {
+  /**
+  We require that each adjacent pair of `Segment`s in a `Path` be related by a
+  join of one onto the end of the other that is present in the underlying
+  sequence graph. We also require that the whole array describes a single linear
+  path.
   */
-  boolean reverseStrand;
+  array<Segment> segments = [];
```
jeromekelleher: What if our `Path` consists of 10^6 segments? In general, our graphs are probably going to become quite complicated, and we want to have a way of traversing the whole graph, so I don't think embedding the `Segments` in an object is the way to go. What about providing a paged API call to get the segments corresponding to a path (as defined here), rather than an explicit type? We are going to want a fairly flexible API for graph traversal in any case, so I don't think there's much value in having a `Path` object.

jeromekelleher:
> @jeromekelleher Say we have the GRCh38 chromosomal assembly and one ~3Mbp HLA haplotype. What will your graph look like? Will adding another ~3Mbp HLA haplotype change the existing fragments in the old graph?

I think the graph will look like @richarddurbin's basic definition - we're really just working out exactly how we should translate that into objects and API calls. So, to answer your question, the graph for your example would consist of three `Segments`. Adding new `Segments` into the graph should not modify the old segments in any way - we just put in a new Segment, which knows where it starts and finishes.

src/main/resources/avro/referencemethods.avdl
```diff
@@ -166,17 +172,17 @@ Reference getReference(
     */
     string id) throws GAException;
 
-/****************  /references/{id}/bases  *******************/
```
jeromekelleher: I think removing this API call will cause major compatibility headaches as we try to merge the graph capable API back into the main repo. I would suggest we have a new interface that we use for getting the sequence for a segment/set of segments, and keep the `ListReferenceBases` interface intact. We can then aim for a slow deprecation process in favour of the graph API (which will natively support a linear reference).

jeromekelleher:
I like the economy of removing some of the objects, and I think we're approaching something minimal that we can work to improve in subsequent PRs. There's a few major outstanding issues for me though:

1. I disagree with a `Reference` corresponding to exactly one `Segment`. Surely a `Reference` must correspond to _set_ of segments? I completely agree that we might not have nice cleanly defined references corresponding to chromosomes, but surely there must be _some_ disjoint subsets of the graph that we are happy are really disjoint? If we map a `Reference` to a set of `Segments` then we can reuse the existing API without modification. We just add a new API call `getSegments(referenceId)` or something similar, and we then have the information we need. All of the existing references API can be left as-is, and we can then slowly deprecate it, rather than insisting on a sharp transition to the graph enabled world.

2. Following on from this point, I think we should seek to edit the existing types and interfaces as little as possible, so that we can have the two APIs coexisting. I think this would ease adoption and make it easier for people to transition to the new API. It will also free us up to experiment more with the new types and calls, as we don't have to worry so much about how this will affect existing users.

lh3:
> Surely a `Reference` must correspond to *set* of segments?

Do you mean `Reference` or `ReferenceSet`? A `Reference` corresponds to one chromosome in the linear view. A `ReferenceSet` corresponds to a set of chromosomes.

jeromekelleher:
> Do you mean Reference or ReferenceSet? A Reference corresponds to one chromosome in the linear view. A ReferenceSet corresponds to a set of chromosomes.

A Reference. I'm arguing that we should define a reference as a set of Segments (which in the linear case will be a singleton set), describing the sequence graph for a particular region. @benedictpaten's point about us not being able to cleanly define a set of Segments (or a single connected graph) as corresponding to only one chromosome is a good one, and this will cause difficulties. However, I think this is less problematic than mapping today's concept of a reference onto a single Segment.

src/main/resources/avro/reads.avdl
adamnovak: Yeah, you aren't supposed to have a `Segment` on a sequence with no `Reference` in the middle of a `Path` in a `GraphAlignment`. The `ReadGroup` is supposed to be aligned against just a `ReferenceSet`. If we want to support aligning to non-reference sequences, we'll need to add a method in the reads API to let `ReadGroup` carry augmentations for the reference just like `VariantSet` does.

adamnovak:
> Hi Adam, I thought we agreed it would be good to allow a reference segment to not necessarily cover the entirety of the underlying sequence - for example when specifying a reference set for a subregion, like the MHC or one of the other HGVM pilot regions?

You're right, I forgot that use case. I've added a comment explicitly allowing it.

src/main/resources/avro/common.avdl
adamnovak: `Position`s absolutely need to have IDs for the sequences they are on. Otherwise you would just have an offset and a side, without knowing what piece of sequence that offset was supposed to be into. The `Segment` object doesn't actually itself store the ID of the sequence it is talking about; it relies on its start `Position` to hold it.

Is backwards compatibility a concern for us beyond how much work it takes to write code for the new version? I was assuming we would do what most web APIs do and have a server present each supported version of the API behind a different endpoint: you would have http://ga4gh.org/api/v0.5/referencesets/{id} and http://ga4gh.org/api/v0.6/referencesets/{id} and so on for each version of the API that your server supports.

If we want to have clients that expect the old API hitting against servers that present the new API, they'll choke as soon as they encounter a `GraphAlignment` on the reads side, and the Variants side has been substantially redesigned. Backwards compatibility on the References I think is also broken by the refactoring I did to make a `Reference` use a `Segment`, so I'm not sure that keeping compatibility at the level of `Position`s is really worth it.

Would we prefer a sort of hybrid API where all the old endpoints and objects are available, but there are new versions of everything under different names?

src/main/resources/avro/referencemethods.avdl
adamnovak: That call is still there, sort of: it's been re-named to the `/sequence/{id}` endpoint and the `getSequence()` method. I could have a copy under the old name though. Would that make sense?

adamnovak:
@jeromekelleher It sounds like what you want might be better implemented as composable `ReferenceSet`s? The current everyday concept of a "reference" is something like GRCh38, which would be implemented as a `ReferenceSet` in our API. If you want to use a smaller part of GRCh38 for some analysis of just a single chromosome or region, you would implement that as a `ReferenceSet` as well.

What we call a `Reference` in our API has a 1:1 correspondence with FASTA records in GRCh38, and a FASTA record is a single sequence, which can be described with a single `Segment`.

Maybe these names are confusing, and we should have `Reference` as the object that `ReadGroup`s are mapped to, and `ReferenceSegment` as the annotated sequence? But then that's another refactoring that would be completely incompatible with the current non-graph API.

src/main/resources/avro/common.avdl
adamnovak: On the one hand, it's a good point that we might want a really long `Path` for some `Allele`, and it might be a bit of a pain to download it all at once.

On the other hand, I anticipate the vast majority of `Allele`s being SNPs, with paths of only a few `Segment`s. Do we want to force an additional round trip to the server for every `Path`?

Either way, we can probably make that change after this gets accepted; it seems fairly straightforward.

adamnovak:
This isn't supposed to be closed; I pushed the wrong branch to the wrong place while bringing the graph branch even with master (so I could have my rebased commits show as a diff against master) and Github helpfully closed it. I'll see if I can get it opened again.

src/main/resources/avro/common.avdl
```diff
@@ -2,6 +2,18 @@
 /**
 This protocol defines common types used in the other GA4GH protocols. It does
 not have any methods; it is merely a library of types.
+
+Many of these types are related to definign sequence graphs. A sequence graph is
```
ryan-williams: *defining :)

adamnovak: Alright, I've fixed that in my branch. Thanks!

adamnovak:
OK, #233 is going to replace this thread. Sorry again for closing it prematurely.

haussler:
There is a lot of detail on this thread. But the one take home for me was
Adam's comment:

"just three types (Side, Position, and Segment)"

Simplicity, economy, and beauty matter a lot. It strikes me that as a pure
form, these three are the fundamental elements of these graphs.

-D

PR #204, opened 2014-12-05
Better array semantics for search requests

adamnovak:
This takes the changes I made in #198 to arrays on search request objects, and pulls them out as their own pull request. A lot of them had comments reading "If specified..." or "If present...", but were not actually optional (i.e. not unioned with null). I changed these to "If nonempty" instead. If you pass an empty array, it means you don't want to use that particular filter in your query.


adamnovak:
I've rebased this against master, I think it is probably mergable.

PR #206, opened 2014-12-05
Dataset enumeration

adamnovak:
Adds the ability to enumerate the datasets that are available, making data discovery and getting an overview of a repository easier.

Includes changes extracted from #198.


adamnovak:
This is now squashed and rebased against master. Can I get another +1 from somewhere? Or some comments as to why we don't want this?

pgrosu:
+1, we definitely need it.

PR #207, opened 2014-12-05
Adding ability to search references by name.

adamnovak:
If we end up with many alt haplotype `Reference`s in a `ReferenceSet`, we might want to pull down, say, the primary path X chromosome named "X" without enumerating through the whole `ReferenceSet`.


adamnovak:
This has been mergable for a while, I'm going to tag it.

adamnovak:
I've also rebased on master; nothing conflicted.

PR #208, opened 2014-12-05
Look up CallSets by sample ID

adamnovak:
Allowing search by sample ID, and reworking some comments.

### DIFF
--- a/src/main/resources/avro/variantmethods.avdl
+++ b/src/main/resources/avro/variantmethods.avdl
@@ -326,6 +326,11 @@ record SearchCallSetsRequest {
   union { null, string } name = null;
 
   /**
+  If nonempty, return only `CallSet`s with one of these sampleIds.
+  */
+  array<string> sampleIds = [];
+
+  /**
   Specifies the maximum number of results to return in a single page.
   If unspecified, a system default will be used.
   */


adamnovak:
OK, I have brought this up to speed with master and pared it down to just the `CallSet` change: you can now look up `CallSet`s by sample ID.

pgrosu:
Nice!  Thank you for this @adamnovak :)

jeromekelleher:
Can you give a quick example of the use case please @adamnovak ? How does this interact semantically with the other fields in the request? Should this be unioned with null so it can be omitted?

adamnovak:
The use case is to look up CallSets by sample ID; if you went over to the metadata API and found a sample you care about, you will need to be able to get genotype information for it.

I meant to have semantics where an empty array doesn't apply the filter; I forgot to specify that in the comment and will fix that momentarily.

adamnovak:
OK, it now describes the proper syntax for not using this field.

adamnovak:
Sorry, now it *actually* actually describes the proper syntax for not using the field; I had amended my commit without adding the file.

jeromekelleher:
Thanks @adamnovak, +1.

david4096:
@adamnovak can we close given https://github.com/ga4gh/schemas/pull/636?

Issue #219, opened 2015-01-12
GAVariant missing "quality" and "filter"


jrobinso:
Hi,  the v0.5.1 GAVariant specification does not list the "filter" and "quality" fields,  which are part of the VCF specification.   However, variants I fetch from https://www.googleapis.com/genomics/v1beta2/variants/search do include them.   Is this a mistake in the specification,  or are these fields no longer required?

johntbates:
Hi, Jim.

You're correct: the GAVariant schema lacks those fields, despite the fact that the Google v1beta2 implementation returns them.  This appears to be an oversight in the specification and I'll be submitting a pull request to address this.  Thank you for pointing it out!

-John Bates

skeenan:
This has been dormant since January. Closing in 2 days.

jeromekelleher:
There is still a PR open (#220) for this, so I don't think we should close it. Perhaps we should ping those involved and try to get the PR updated?

skeenan:
Thanks @jeromekelleher I've pinged @johntbates in https://github.com/ga4gh/schemas/pull/220

david4096:
Closed with https://github.com/ga4gh/ga4gh-schemas/pull/826

PR #220, opened 2015-01-12
Added quality and filter fields to GAVariant.

johntbates:
#219 pointed out we were missing two fields that were present in the VCF specification.

### DIFF
--- a/src/main/resources/avro/variants.avdl
+++ b/src/main/resources/avro/variants.avdl
@@ -204,6 +204,18 @@ record Variant {
   /** The bases that appear instead of the reference bases. */
   array<string> alternateBases = [];
 
+  /** 
+  A measure of how likely this variant is to be real.
+  A higher value is better.
+  */
+  union { null, double } quality = null;
+
+  /** 
+  A list of filters (normally quality filters) this variant has failed.
+  PASS indicates this variant has passed all filters.
+  */
+  array<string> filter = [];
+
   /**
   A map of additional variant information.
   */
--- a/src/main/resources/avro/variants.avdl
+++ b/src/main/resources/avro/variants.avdl
@@ -205,8 +205,9 @@ record Variant {
   array<string> alternateBases = [];
 
   /** 
-  A measure of how likely this variant is to be real.
-  A higher value is better.
+  Phred-scaled quality score for the assertion made in ALT.
+  If ALT is ‘.’ (no variant) then this is −10log10 prob(variant), and if ALT is 
+  not ‘.’ this is −10log10 prob(no variant).
   */
   union { null, double } quality = null;
 


richarddurbin:
Yes, very good to add these. 
I wonder whether filter should be renamed to filtersFailed, and an empty filtersFailed list should imply PASS.
Also, I think that the description for quality should be tightened up.  I suggest to add something like:

Standardly, the value is -10 log_10 p, where p is the probability of the Variant being a real genetic variant given the data, under a probablistic model for the data given the variant and a population genetic model for genetic variation.

haussler:
+1

johntbates:
I generally agree, but according to the VCF 4.2 specification, a single value of "PASS" indicates that all filters have passed - see https://samtools.github.io/hts-specs/VCFv4.2.pdf sec 1.4.1 Fixed fields, heading 7. 

I'll go ahead and update the quality field comment.

selewis:
+1

calbach:
@richarddurbin's proposal for `filtersFailed` is much much cleaner than the existing semantics (sentinel value of single element array containing magic "PASS" constant). But perhaps it makes sense to keep parity with VCF for now, since I suspect all current 0.5.1 implementations use VCF-semantics, and work this change into a future version.

Perhaps there could be a follow-up PR which switches the name and semantics and the discussion can be separated?

johntbates:
@calbach I agree that the @richarddurbin's proposal is cleaner than the VCF-parity version, but as far as I know, all existing implementations report the sentinel value, if they report filter values at all.

adamnovak:
I would be in favor of dropping the PASS. The translation back and forth to
VCF is very clear, and we don't need to support grepping over all the
records for PASSing ones, or have any potential problems with empty fields.

jeromekelleher:
I think @richarddurbin's `filtersFailed` is much better, and trivial to derive from VCF. Existing implementations are for version ~0.5, which we've made lots of breaking changes to, so we shouldn't be too worried about maintaining compatibility with these.

adamnovak:
I think there's consensus around the `filtersFailed` implementation. @johntbates do you want to change this over to use that syntax?

skeenan:
@johntbates do want to change over to filtersFailed as per https://github.com/ga4gh/schemas/pull/220#issuecomment-69623353

fnothaft:
As an enhancement, it might be better to have both `filtersFailed` and `filtersPassed` arrays. I like the `filtersFailed` semantics, but if you have an empty `filtersFailed` array, you can't tell whether you passed _all_ filters, or didn't apply _any_ filters.

kozbo:
Please see #656 to close this

david4096:
Closing in favor of #656

PR #233, opened 2015-01-23
Graph Adaptations of Variants, References, and Reads APIs

adamnovak:
This PR replaces #198, which I incorrectly (and permanently) closed.

I've re-based all my changes against this morning's master, and addressed some typos that were pointed out to me there.

There seemed to be a couple of threads of discussion running when the other PR got closed:

1. Backwards compatibility: this PR doesn't really have it, because it refactors some things (particularly the Variants API). Is that a problem"?

2. Aligning reads to non-reference sequences: @calbach brought this up, and I'm prohibiting it for now. It would be nice to have a way to augment `ReferenceSet`s with additional sequence, but that should be a new PR after this one gets accepted.

Anything I missed, or any new issues?


jeromekelleher:
The argument for backwards compatibility has a large bearing on how we proceed, so I guess we should resolve this issue first. I'm not a fan of working on a branch like this --- as we've already seen, keeping up to date with the main branch is difficult and time consuming. I think it'll be very hard not to end up with two completely different protocols, with two independent implementations. I don't want to have to work with two different server implementations, with two different client libraries. Git simplifies this a lot of course, but it would still be a major headache.

Ideally, as an implementer, I want _one_ protocol that is versioned.  We can add new things in to the protocol in a new version, and we can mark interfaces as deprecated if we don't like them any more. Maybe we can remove things that were marked as deprecated (say) two versions ago, if we're sure that the majority of people have moved over the using the new recommended interface. It's probably a bit early in our design phase for the protocol to be thinking like this, but we will have to start soon. 

If it were at all possible to bring our graph concepts into the current protocol _without_ breaking it, I think this would be a huge gain, and worth the cost of perhaps not having exactly the abstractions that we might ultimately want  (but we can seek to improve over time, in non disruptive way). 

One might argue that we have made lots of breaking changes since v0.5.1, so there's no harm in adding some more. These have all been pretty minor up to now though, and can mostly be dealt with mechanically. The changes proposed above are deep semantic changes to existing objects and interfaces, which would be very disruptive to both server implementers and client side code.

In short, I would like us to try to add the minimal graph concepts into the existing API, without forking. We can add new types, and add fields into existing types, but not delete any old ones. Implementers would be free to ignore the graph extensions and incrementally update their servers to keep up with new protocol versions. Isn't it worth at least _trying_ to do this, and reverting to the parallel-protocols model if/when we show it can't be done, or the compromises are too great?

adamnovak:
I think if we are going to make breaking changes, we should make them as soon as possible, so that less code has to be rewritten.

It was my impression we had a consensus around developing an optimal, non-backwards-compatible API in a graph branch, and moving it over to the master branch when we were ready to release a new major version of the API with graph support.

What you are proposing---making a few changes to enable graphs while not breaking existing code---sounds like it ought to be a pull request against master instead of against a topic branch, and like it would be suitable for release in a minor version.

I don't in principle see why we can't do both: we can consider this PR (which isn't going into the master branch) on its merits s a graph-based API, and I can also make a pull request against master that backports some of the less disruptive features to the current major version. I could probably allow References to join onto each other, for example. However, I think the most useful features, like the allelic model of variation, are going to require breaking changes.

jeromekelleher:
This is entirely reasonable @adamnovak , and you've convinced me that we will have to make some breaking changes. I would also like to get a version of the graph API committed as soon as possible, and to start refining this using subsequent, smaller PRs. However, I am still opposed to forking, for reasons I'll explain. I'd like to propose an alternative strategy that will allow us to start making progress,  won't impact on existing users or implementers, and will allow us to make efficient use of our developer resources on the reference implementation.

The main reason that I am opposed to working on a branch and forking a different version of the protocol is a selfish one. It will make our lives on the reference server a lot more difficult, because we want to (a) support the graph API as soon and completely as possible, and (b) to provide a useful complete implementation of the mainline protocol for users who don't care about graphs  and just want to make their data available (who, I would imagine, will be in the majority). If we fork the protocol, then we'll have to fork the reference implementation. I think it will be difficult keeping the protocol definitions up to date in git between the main and graph branches, but it will be very tricky indeed keeping the actual implementations in sync. It's a lot of git engineering and, fundamentally, wasted effort in my opinion. I'd much prefer to write code than resolve merge conflicts and try to remember which commit should be applied to which branch.

My proposal is that instead of creating a branch and forking the protocol, we add the new graph types into the main branch, and use either namespacing or a prefix (say, Graph*) to distinguish the new graph oriented aspects of the API. For example, we might have a `GraphVariant` object, which implements the new allelic variant model.  These new graph types and API calls would be marked as optional, so that implementers would be free to ignore them. We can also clearly mark these as highly experimental, with no guarantees at all about their API stability. We might even filter them out of the official API documentation entirely, to avoid confusing users. All the existing APIs will remain untouched.

As the graph API matures, we can seek to merge the new types into the old in a compatible way. At any rate, we can think about how to do this properly over the long term---right now, we can just add in the types and APIs that we want, and refine how they work as we get more experience with them. Does this seem like a reasonable approach?

adamnovak:
Adding Graph* objects would work fairly well for some of the changes. I'm already proposing a `GraphAlignment` on the reads side as an alternative to the `LinearAlignment`.

I don'r know if it would make sense to have e.g. `Position` and `GraphPosition`: some objects can be updated to work well in a graph context with just a few changes. While they might not be backwards-compatible, neither do they require refactoring.

I'm worried we might have quite a proliferation of Graph* objects on the variants side though. I think we would need `GraphCall`, `GraphAllele`, `GraphGenotype`, `GraphVariant`, `GraphAlleleScaffold`. and possibly `GraphCallSet` and `GraphVariantSet` if we don't want graph objects in the normal collections.

This actually sounds somewhat like one of my initial drafts, where I had graph-supporting methods under a `/graph/` subdirectory in the API.

Can we get some other opinions on whether this sort of split would be a good idea? @richarddurbin? @lh3? @benedictpaten? @calbach?

haussler:
I feel uncomfortable putting the "graph" prefix on any object names that
won't always absolutely need these prefix. In the end, "segment",
"position", "side", etc. are what these names need to be. On the other
hand, I'm all for integrating the graph work that has been done asap into
the main branch. We have had a toe in the water long enough. Time to dive
in. -D

diekhans:
Avro provides namespaces that can be used instead of prefixes to
keep names unique without making them huge.  I would suggest
that we consider creating another level of namespace below
org.ga4gh.models as a far more modern approach than prefixing to
keeping names unique.

jeromekelleher:
Namespacing works fine for me, if we can get Avro to do it. It does seem like we've gone in a bit of a circle here @adamnovak --- sorry to cause trouble at a late stage, but I hadn't thought through the implications of forking fully until last week.

diekhans:
Avro supports name spaces.  I am planning on putting in PR schemas to
discuss this.

adamnovak:
So is there consensus around me integrating what I can into the base objects, and doing any needed refactoring in `org.ga4gh.whatever.graph` namespaces, for a pull request against master?

jeromekelleher:
I think so --- I would try to leave the base objects alone as much as possible, but adding an extra field or two (marked optional and experimental) should be fine. Everything else should go into the new namespace. I think it's fine duplicating APIs for (e.g.) references, as this will give us the flexibility to experiment without worrying about the impact on existing users.

benedictpaten:
Sorry, I'm behind here - off working on endless proposals. I appreciate Jerome's points. I was also worrying about the graph branch fork of the reference server! 

I think Adam should try to make a pull request against the master. The danger is that we may add a bunch of prototype graph stuff to the main API that we later end up having to substantially revise. I guess namespaces may partially help avoid this - but not for objects where we are adding new fields.

Given a substantial enough team I think the graph branch approach is the correct one, but as our team is not that large I kind of agree this may be the best path.

benedictpaten:
Quick update: Adam and I have come up with a reasonable revision to the current pull request that could work against the master branch API. Good to be given a challenge!

We think we probably don't need namespaces, and can rather unionise the current and new objects to create a fully backwards compatible pull request. This has led to some good thinking about how to further make the graph changes overlap as closely as possible with the current API. The downside will be that some objects, like Call, will be a bit more complex, having essentially two different ways to specify a variation - by reference to an allele object or by reference to a variant object - a variant in the old schema being essentially a site. This should no break backwards compatibility, providing existing clients of v0.5.2 ignore newly added fields.

haussler:
excellent

richarddurbin:
Sounds great.  Looking forwards to seeing this.  Richard

On 28 Jan 2015, at 19:35, haussler <notifications@github.com> wrote:

> excellent 
> 
> On Wed, Jan 28, 2015 at 11:29 AM, Benedict Paten <notifications@github.com> 
> wrote: 
> 
> > Quick update: Adam and I have come up with a reasonable revision to the 
> > current pull request that could work against the master branch API. Good to 
> > be given a challenge! 
> > 
> > We think we probably don't need namespaces, and can rather unionise the 
> > current and new objects to create a fully backwards compatible pull 
> > request. This has led to some good thinking about how to further make the 
> > graph changes overlap as closely as possible with the current API. The 
> > downside will be that some objects, like Call, will be a bit more complex, 
> > having essentially two different ways to specify a variation - by reference 
> > to an allele object or by reference to a variant object - a variant in the 
> > old schema being essentially a site. This should no break backwards 
> > compatibility, providing existing clients of v0.5.2 ignore newly added 
> > fields. 
> > 
> > — 
> > Reply to this email directly or view it on GitHub 
> > <https://github.com/ga4gh/schemas/pull/233#issuecomment-71898963>. 
> >
> —
> 




-- 

adamnovak:
Closing in favor of #238

PR #238, opened 2015-02-03
Graph Support for GA4GH API

adamnovak:
This PR adds graph support to the References, Reads, and Variant APIs in a backwards-compatible way.

`common.avdl` is enhanced with `Segment` and `Path` types, as well as the notion of a sequence and a sequence graph, which is described at length at the top of the file.

`Reference`s can now have associated `Segment`s, placing them in a sequence graph for the ReferenceSet.

Reads in the Reads API can now be aligned using `GraphAlignment`s as well as `LinearAlignment`s.

`Variant`s in the Variants API can have `Allele`s defining their reference and alternate forms as paths in a sequence graph, and `Call`s can call copy numbers for individual `Allele`s as well as genotypes for `Variant`s.

Closes #233


src/main/resources/avro/referencemethods.avdl
```diff
@@ -230,4 +236,68 @@ ListReferenceBasesResponse getReferenceBases(
...
+  /**
+  The continuation token, which is used to page through large result sets.
+  Provide this value in a subsequent request to return the next page of
+  results. This field will be empty if there aren't any additional results.
+  */
+  union { null, string } nextPageToken = null;
+}
+
+/**
+Lists bases by sequence UUID and optional range.
```
jeromekelleher: Why UUID here? I think this should just be an implementation private ID value.

src/main/resources/avro/references.avdl
```diff
@@ -2,21 +2,56 @@
...
+
+  /**
+  The `Segment` describing this `Reference`'s sequence in the `ReferenceSet`'s
+  sequence graph.
+
+  The `Segment` will generally cover the entirety of the sequence, but may be
+  shorter if only a subrange of a sequence is to be included in the
+  `ReferenceSet`.
+
+  The sequence UUID of the `Segment` is equal to ID of the `Reference`. The
```
jeromekelleher: s/UUID/ID ?

src/main/resources/avro/variantmethods.avdl
```diff
@@ -51,6 +51,89 @@ SearchVariantSetsResponse searchVariantSets(
...
+
+/******************  /variantsets/{id}/sequences/search  *********************/
+/**
+This request maps to the body of `POST /variantsets/{id}/sequences/search` as
+JSON.
+*/
+record SearchVariantSetSequencesRequest {
+  /**
+  If nonempty, return `Segment`s describing novel sequences in the `VariantSet`
+  that are joined onto sequences with one of the specified UUIDs.
```
jeromekelleher: s/UUID/ID

src/main/resources/avro/variantmethods.avdl
```diff
@@ -51,6 +51,89 @@ SearchVariantSetsResponse searchVariantSets(
   /** This request maps to the body of `POST /variantsets/search` as JSON. */
   SearchVariantSetsRequest request) throws GAException;
 
+
+/******************  /variantsets/{id}/sequences/search  *********************/
+/**
+This request maps to the body of `POST /variantsets/{id}/sequences/search` as
+JSON.
+*/
+record SearchVariantSetSequencesRequest {
```
jeromekelleher: Would `SearchVariantSetSegmentsRequest` mapping to `/variantsets/{id}/segments/search` be more natural here? We're returning lists of `Segments` so it would seem a bit clearer to me.

src/main/resources/avro/variantmethods.avdl
```diff
@@ -51,6 +51,89 @@ SearchVariantSetsResponse searchVariantSets(
...
+  union { null, string } nextPageToken = null;
+}
+
+/**
+Gets a list of `Segment` matching the search criteria.
+
+`POST /variantsets/{id}/sequences/search` must accept a JSON version of
+`SearchVariantSetSegmentsRequest` as the post body and will return a JSON
+version of `SearchVariantSetSegmentsResponse`.
+*/
```
jeromekelleher: There's a mismatch between the docs and the definition here. Did you previously have `Segment` rather than `Sequence`?

src/main/resources/avro/variants.avdl
```diff
@@ -211,9 +273,59 @@ record Variant {
...
+  string id;
+
+  /** The ID of the variant set this allele belongs to. */
+  string variantSetId;
+
+  /**
+  The time at which this `Allele` was last updated in
+  milliseconds from the epoch.
+  */
+  union { null, long } updated = null;
```
jeromekelleher: Do we need this at the moment? It seems like keeping an update time is to solve a problem we're a long way from encountering, and can be easily added in later if/when we need it.

jeromekelleher:
Thanks very much @adamnovak, this looks great. I can't see any backwards compatibility issues, so I would vote to merge this ASAP and start getting some real experience working with the graph model. A couple of minor, non-blocking issues:

1. Why UUIDs for sequences? This seems like an unnecessary requirement to me; implementation private IDs would be sufficient surely.
2. It would be good to clarify the relationship between a sequence/reference/segment, as they are used somewhat interchangeably in the documentation and in the methods.

src/main/resources/avro/variantmethods.avdl
adamnovak: It should definitely be sequence here, and later where I've accidentally talked about a `SearchVariantSetSegmentsRequest`.

The sequences are the actual components of the sequence graph, which is what you are asking about here. However, since sequences aren't represented by an Avro record type, we describe the sequences in the graph with a set of `Segment`s that cover them, one per sequence.

Segments do not have IDs; they are merely regions defined on sequences, possibly including the sequences' end joins.

adamnovak:
@benedictpaten wanted UUIDs for sequences to facilitate merging across different API endpoints without having to rewrite or prefix IDs. If we are fine with having to manually construct truly global IDs (or just using accessions of some kind), I can get rid of the UUID thing.

For the sequence/reference/segment confusion, here's the simplest way I have found to put it so far:

A sequence is a component in a sequence graph, optionally joined at its ends to other sequences. Sequences are not directly retrievable through the API, but ranges of their bases can be downloaded.
            
A `Segment` is a pointer to a range on a sequence, possibly including the joins at the sequence's ends. It does not include base data. When the server has to refer to some or all of the bases in a sequence, it uses a `Segment` to indicate them. A `Path` through a sequence graph is a list of `Segment`s, while a sequence graph itself can be described using a set of `Segment`s, one for each sequence.
            
A `Reference` is an object that marks a sequence as part of a `ReferenceSet`'s sequence graph. It contains a `Segment` referring to the entirety of the sequence it annotates.

This all kind of depends on the explanation of sequence graphs at the top of common.avdl, but I feel like maybe it should come first to give a higher-level overview. Thoughts?

adamnovak:
OK, as a result of the discussion on the call this morning, I've introduced the concept of supported modes. API servers that support the "graph" mode will always populate the graph fields, and API endpoints that support the "classic" mode will always populate the non-graph-supporting fields that I made nullable (and also won't do confusing things like return `GraphAlignment`s in the reads API).

If a server supports the "classic" mode, a client doesn't have to know anything about graphs in order to use it. If a server supports the "graph" mode, a client can just look at the new graph fields and not have to worry about doing its own up-conversion. A server can support both modes perfectly fine, although it won't be able to return things only expressible in graph terms.

This doesn't require any state on the server side; a server is permanently fixed in the modes it supports and the fields it fills in, regardless of what a client actually wants to use.

jeromekelleher:
Re UUIDs - we've gone with hashes everywhere else for these sorts of requirements. I think we should change to ID here and add in a hash field of some sort later.

mbaudis:
For metadata we make the format for id => UUID (recommended):
```
  /**
  The individual UUID. This is globally unique. 
  The recommendation is to use an anonymous UUID type with extremely low collision probability.
  The typical implementation would be UUIDv4.  
  This has to be provided by the implementation.
  */  
  string id;
```
This is from the working doc https://github.com/ga4gh/metadata-team/blob/master/avro-playground/metadata_redo.avdl

jeromekelleher:
I like the modes distinction @adamnovak --- I'm not sure about the exact mechanism for clients to discover the mode that a server is running in, but I think something like this is what we need. I'd like to clarify what I think is important about having these two modes, and why we need them.

There are a lot of people out there who have VCF data relative to a linear reference who would like to share their data. There are a lot of tools out there that ingest data in VCF. The v0.5 classic API is a direct translation of the concepts that these tools understand, and allow us to serve VCF data in a relatively straightforward manner. Importantly, we can convert from VCF to the classic API easily allowing us to make data available. We can also convert data from the classic API back to VCF in a one-to-one manner, allowing us to use existing tools on data coming from a GA4GH server. There will be VCF data for quite a long time, and it's critical to the success of the protocol that these workflows are supported naturally.

Ultimately, we hope that the field will move from a linear reference to a graph reference, but this will take a long time. We need to provide an evolutionary path for tools to follow: first they adopt the classic APIs, which translate fairly directly into the concepts that they work with today. Then, slowly, they start to adapt to the graph APIs and begin reaping the benefits (which we will demonstrate). Eventually, everyone (or at least a lot of people) will have moved over to the graph API, and we can then start to deprecate the classic API, eventually removing it altogether.

The main reason we don't want two different protocols is to avoid splitting our resources, and wasting time keeping them in sync. The classic and graph mode APIs  _must_ be closely compatible for this strategy to work, and the best way to do this is by implementing both simultaneously. If we fork off two separate protocols, they will never again merge and we'll split our limited developer resources between working on one or the other.

There are unarguably downsides to this dual protocol approach, but I think they can be alleviated fairly easily. Firstly, server implementations are in no way required to implement graph mode, and are free to continue using classic mode for as long as they want. Secondly, clients should be encouraged to use classic mode by default, as this is by far the most likely thing that they would want to do. Also, graph mode is experimental, and will probably change considerably over time and so new users of the protocol should not be exposed to it unless this is explicitly what they want. A graph mode capable server running in classic mode should not return any of the new attributes in the JSON responses (since they will be null, it is the same thing to just leave the fields out).

The approach is a compromise, but it is hopefully one which allows us to both support users who want to work with existing datasets, and also to develop the graph based methods.

ekg:
As long as our data can be converted between VCF and the graph model,
people will continue to be able to use tools structured around the linear
reference. The same goes for BAM and graph-based alignments. Obviously
these models are not completely congruent, so conversion from graphs ->
linear systems will lose data, but in the vast majority of cases it seems
trivial to convert back and forth between a linear and a graph way of
working.

I think that a gradual development to using the graph is unlikely, and
instead big leaps will be made where it is adopted. Most old tools will
remain useful as long as conversion back to the linear model is easy. New
tools will be written against a graph model.

adamnovak:
@jeromekelleher The way I've done the modes now, it's perfectly fine for a server to support both the graph and classic modes, as long as they don't say things that can't be articulated in classic mode. I think that that would be the mode you want to run your servers in, because you have maximum compatibility with all clients.

I guess it makes sense not to send any graph fields if you don't actually advertise support for graph mode; however, I didn't write that in as a requirement, and I think clients should have to handle it. A classic-mode-only client doesn't have to *interpret* graph mode fields, but it does have to correctly ignore them if they are present. Otherwise it wouldn't work on a server that sends both modes.

While a client will probably want to support reading data in classic mode, I agree with @ekg and think new client development should be focused on graph mode. A client designed around graph mode can do its own trivial upconversion of classic mode replies, but the reverse is not true.

@mbaudis I think standardizing all the GA4GH APIs on recommending the use of UUIDs for record IDs is probably a good idea, but I'd like to do it in a seperate pull request, since it would apply to a bunch of API objects I haven't touched in this one.

Does anyone have any recommendations for things I should change? Or other reasons why they can't give this a +1?

mbaudis:
@adamnovak My UUID comment was just a reminder to push this as a global default, not necessarily here; though objects with clear & limited context/scope may probably do better without the generation/storage overhead (?).

Don't feel entitled to +1 in though out of sheer personal incompetence ...

benedictpaten:
I'm +1 on this (obviously), but we need at least a couple of non-UCSC people to +1 it. I think Adam should squash down the commits to make it a bit easier to digest the changes.

adamnovak:
I've squashed everything down to one commit for ease of evaluation. I'd appreciate it if people could take a look at this.

src/main/resources/avro/references.avdl
```diff
@@ -1,22 +1,74 @@
...
-  MD5 of the upper-case sequence excluding all whitespace characters
-  (this is equivalent to SQ:M5 in SAM).
+  The MD5 checksum uniquely representing this `Reference` and its position in
+  the `ReferenceSet`'s sequence graph, as a lower-case hexadecimal string.
+
+  If `segment.startJoin` and `segment.endJoin` are both null, or `segment`
+  itself is null, this is the MD5 of the upper-case sequence excluding all
+  whitespace characters (this is equivalent to SQ:M5 in SAM).
+
+  Otherwise, this is the MD5 of the above MD5 checksum, `segment.startJoin`'s
```
calbach: How is the MD5 of the `{start,end}Join` defined? Is this the `md5checksum` fields of those segments? I believe that given the semantics of {start,end}Join, there cannot be a circular dependency here (please verify), but what's the worst case "cascading" MD5 checksum recalculation I could cause by adding a new reference to the graph?

I'm not sure that this is a real problem that should matter; the question is more just to test my understanding of how checksums are working here.

adamnovak: The next paragraph in the comment describes the MD5 of a `Position` (which is what `startJoin` and `endJoin` are) as the MD5 of the `md5checksum` of the `Reference` it is on, the position it is at, and the strand as "+" or "-".

You are right that there can't be circular dependencies. But it's also impossible to cause a recalculation by adding a new sequence. New sequences are only allowed to join onto existing sequences---existing sequences can't be made children of new sequences. So since a `Reference`'s `md5checksum` depends only on itself and its parents, adding new sequences won't change the `md5checksum`s of any existing `Reference`s.

If you did somehow go back and change the parent of some existing sequence, it and all its children, and all their children, and so on, would all need their `md5checksum`s recalculated.

diekhans: IMHO, these type of questions means the documentation needs enhanced.

adamnovak: I've added a comment to the effect that `Reference`s are designed to be immutable, with `ReferenceSet`s most easily extensible through the addition of new child references. Any other ideas on how the documentation could be enhanced?

lh3:
Several comments.

1. Are "classical" and "graph" the right names? This PR consists of two largely independent components: graph representation and allelic model. By independence, I mean we could in principle split this PR into the two with little interference. The graph mode is actually "graph+allelic", and the "classical" is "linear+genotypic". There could also be other combinations "linear+allelic" and "graph+genotypic" as well. In `variants.avdl` of this PR, we frequently say that some fields are mandatory in the graph mode, but more exactly, most of them are required by the allelic model, not by the graph representation. I think we need to rename the "graph mode" to something that is conceptually correct. I couldn't think of a clean name, though.

2. The `Call` object is overloaded. In the genotypic model, a call is a genotype call. In the allelic model, a call is an allele call with a very different set of new fields (the only overlapping field is apparently `callSetId`). The overloading obscures the meaning of `Call`. I would suggest completely separating allelic and genotypic calls by adding an `AlleleCall` object and moving allele-associated fields to the new object. `AlleleCall` is conceptually different from the old `Call` object.

3. Similarly the `Variant` object is overloaded to a lesser extent - some fields are allelic only and some are genotypic only. We might consider a similar treatment (see also point 4 below). On the other hand, `Variant` is a concept common to both allelic and genotypic models. Adding `AlleleVariant` seems inconsistent. I am not sure if there is a better solution.

4. If we decide to use one `Variant` object in both the genotypic and allelic models, we need to be careful about the `updated` time. In the genotypic model, the allele sequences are members of `Variant`, which would imply that update the allele sequences would change the `updated` time. In the allelic model, the ID of alleles, rather than the alleles themselves, are members of `Variant`. If we change an allele sequence without changing its ID, do we change the `updated` time? In addition, for the `referenceName`, `start` and a few other fields, the API doc only says "If the API server supports the classic mode, this field must not be null", but it does not specify what these fields mean in the "graph" (more exactly allelic) mode. Do they have to be null? If we create an `AlleleVariant` object, this won't be a problem.

5. Could someone write down code snippets to retrieve the genotype of some samples in a small region? I wrote some when the allelic model was on a separate branch. With both genotypic and allelic models mixed together, I am even less sure about what is the right code. This is critically important. If we creators of the APIs have difficulties in writing down the queries, endusers will have more troubles.

jeromekelleher:
@lh3 raises excellent points. I have no issue with creating new types `AlleleCall` and `AlleleVariant`, except it does complicate things further up the stack. For example, `SearchVariantsResponse` contains a list `variants`, which has the type `array<Variant>`. This would then need to be `union( array<Variant>, array<AlleleVariant>)`, so that we can return either type. (Avro's lack of inheritance is getting increasingly annoying.) We're then just pushing the overloading further up the stack. I'm not sure which option is better: have the overloading occurring at the query level, or at the object level.

In general, we seem to have two choices:

1. Overload some types so that we can to implement the graph+allelic model as well as the linear+genotype model using the same queries;
2. Make new types and queries exclusively for the graph+allelic model. This brings us back to the namespacing approach.

What do you think @lh3, which is the lesser of two evils?

haussler:
Heng is right. The big distinction here, the one that will affect genetics
for decades, is "genotypic" versus "allelic". It would be a mistake to try
to get the world to convert to a graph model without explaining the deep
motivation for adding an "allelic" mode and eventually deprecating the
"genotypic" mode.

haussler:
Well, not really "deprecating". Genotypic is here to stay. Maybe making it
more of a derived summary from the allelic information.

diekhans:
Hi David, Heng,

Having been involved in API design before, it is hard to stress
how important it is to get the terminology and naming right now.
If the terminology is confusing, it will confuse people for the
next 20 years.

We also need to have complete, comprehensive documentation for
all of the APIs, including rationale, in of the source tree.
Documentation not in the source tree will be out-of-sync, lost,
or just plain wrong.

Mark

fnothaft:
@diekhans for comprehensive documentation inside the repository, I suggest using [Pandoc](http://johnmacfarlane.net/pandoc/). We use this with [ADAM](https://github.com/bigdatagenomics/adam/tree/master/docs) and it makes the process of attaching docs to a release seamless.

fnothaft:
> Re UUIDs - we've gone with hashes everywhere else for these sorts of requirements. I think we should change to ID here and add in a hash field of some sort later.

+1 @jeromekelleher

fnothaft:
> @lh3 raises excellent points. I have no issue with creating new types AlleleCall and AlleleVariant, except it does complicate things further up the stack. For example, SearchVariantsResponse contains a list variants, which has the type array<Variant>. This would then need to be union( array<Variant>, array<AlleleVariant>), so that we can return either type. (Avro's lack of inheritance is getting increasingly annoying.) We're then just pushing the overloading further up the stack. I'm not sure which option is better: have the overloading occurring at the query level, or at the object level.

Wouldn't it make more sense to have `SearchAllelicVariants` and `SearchGenotypicVariants`? I think earlier we had said that a server that was going to serve both variant types would need to maintain some state (indicating what sort of variants it had). Here, we would make the request type explicit, and the server could just error if it can't serve either of `Allelic` or `Genotypic` variants.

I'd generally prefer to not overload types via unions.

adamnovak:
`Call` shares a few fields between `Call`s that are associated with an `Allele` and `Call`s that aren't: both can use `totalCopies`, `phaseset`, `callSetName`, `callSetId`, and `variantId` (since a call on an `Allele` can also have an associated `Variant`).

If we split it up, we'll have to duplicate a substantial portion of the Variants API, and have some sort of constraint that the two views agree with each other.

fnothaft:
> Well, not really "deprecating". Genotypic is here to stay. Maybe making it
more of a derived summary from the allelic information.

+1; it seems like this is a _necessary_ condition for the API. If people can't convert data stored in the allelic model to the genotypic model, people won't _store_ data in the allelic model.

src/main/resources/avro/common.avdl
```diff
@@ -2,6 +2,325 @@
 /**
 This protocol defines common types used in the other GA4GH protocols. It does
 not have any methods; it is merely a library of types.
+
```
fnothaft: Small nit, but it looks like the indentation on this documentation is off. Should be a 2 space indent.

diekhans: and can we *please* do consistent indentation and not special
case namespace to not be indented.  This makes it harder to read
and breaks automatic indention modes.  This kind of special-case
formatting to save columns when out of style a long time ago.

adamnovak:
For mode names, I definitely don't want to call the modes "genotypic" and "allelic". If you're using linear sequences you can only make genotypic calls, but if you're using graphs you can make both genotypic and allelic calls, and will probably want to have both in the same `CallSet`. I think "classic" and "graph" (or maybe "linear" and "graph") is the right terminology.

So here's how a split API would probably want to work:

I think we need to keep the `Variant` type unified, because as I said you might want to make genotypic and allelic calls on what is really the same variant. You do end up with duplication of information between the backwards-compatible linear-reference fields and the new `Allele` ID references, but everything else is still shared, and we don't want to be calling it `AllelicVariant` or something forever into the future.

We would probably want to keep `Call` as the name of the genotypic call for backwards compatibility, and add a new `AlleleCall`. While a `Call` has to be associated with a `Variant` and gives a genotype, an `AlleleCall` is associated with an `Allele` and optionally a `Variant`, and gives a copy number.

We would then have one method `searchCalls()` to get `Call` objects (by `CallSet`, `Variant`, or both), and another method `searchAlleleCalls()` to get `AlleleCall` objects (by `CallSet`, `Allele`, and/or `Variant`).

These two methods don't quite map to the different modes, though. A server that serves classic/linear mode has to support `searchCalls()`, but a server that serves graph mode has to support both `searchAlleleCalls()` and `searchCalls()`, because we still want to provide people with a way to make genotype calls when they are defining `Variant`s in terms of `Allele`s in the graph.

How would that look to people?

src/main/resources/avro/common.avdl
fnothaft: Also, a general comment about the documentation: words are fundamentally a bad way to describe graphical concepts. I'm glad to see the documentation come into place, but many of these concepts can be more clearly described via annotated illustrations.

src/main/resources/avro/common.avdl
```diff
@@ -2,6 +2,325 @@
...
+```
+
+It looks like this:
+
+```
+      --C--
+     |     |
+=G==G==T==G==G==N==G=
+      |     |
+       -----
```
fnothaft: Also worth noting that illustrations like this don't lay out clearly in ASCII text.

src/main/resources/avro/common.avdl
adamnovak: @fnothaft Are you sure? According to [CONTRIBUTING.md](https://github.com/ga4gh/schemas/blob/master/CONTRIBUTING.md) it looks like the comment is supposed to be at the same level as the delimeters, which are at the same level as the surrounding code, which is what I've done for the top comment in common.avdl.

Or are you talking about something else?

src/main/resources/avro/common.avdl
```diff
@@ -18,33 +337,112 @@ enum Strand {
...
+
+  /**
+  The ID of the sequence on which the `Position` is located. This may be a
+  `Reference` sequence, or a novel piece of sequence associated with a
+  `VariantSet`.
+
+  If `referenceName` is null, this must not be null.
   */
-  string referenceName;
+  union { null, string } sequenceId = null;
```
fnothaft: Perhaps I missed this earlier, but why is `sequenceId` a string?

src/main/resources/avro/common.avdl
```diff
@@ -18,33 +337,112 @@ enum Strand {
...
+  Joins may occur on the outer sides of the terminal bases in a sequence: the
+  left side of the base at index 0, and the right side of the base with maximum
+  index. These are the "terminal sides" of the sequence. `startJoin` is the join
+  on the side indicated by `start`, and may only be set if that side is a
+  terminal side. Similarly, `endJoin` is the join on the opposite side of the
+  piece of sequence selected by the segment, and may only be set if that side is
+  a terminal side. The value of `startJoin` or `endJoin`, if set, is the side to
+  which the corresponding side of this `Sequence` is connected.
+  */
+  union { null, Position } startJoin, endJoin;
```
fnothaft: Nit: if you define these as `union {null, Position} startJoin;` you should initialize the fields to `null`: `union { null, Position } startJoin = null;` to eliminate runtime exceptions.

benedictpaten:
Thanks Heng, Jerome.

We debated splitting out the graph fields from both call and variant. Adam
has a nice UML diagram that he's working on that makes it easier to see the
relationships - @adamnovak, I think you should post this.  Jerome wanted
backwards compatibility - well we've got it, but the cost is some apparent
duplication.

I agree with Heng about creating code snippets. Adam - can you get Maciek
to do this as an exercise?

To David's point, I think it's very important to distinguish the concept of
genotypes (sites) vs. allelic and linear reference vs. graph reference.
These are separate ideas. What we have now allows us to express, as Heng
says, all four combinations of model: genotypes+linear-ref,
 genotypes+graph-ref, allelic+graph-ref, and allelic+linear-ref (because
the linear-reference is a special case of the graph model - so you can use
the new segment object to describe an allele).

I think this will be cleaner in the future, because we can slowly deprecate
the linear reference stuff, then we'll end up with a single variant object,
which allows us to express genotypes on the graph and a single allele
object.

My strong opinion is not to split variant into variant and allele variant,
it does not fit the semantics.

I am ambivalent about splitting the call object into allele and genotype
calls.

src/main/resources/avro/common.avdl
adamnovak: @diekhans Can you do a PR against CONTRIBUTING.md for that?

I think we don't indent for the protocol because then the doc comments on the objects in the protocol would be indented. This was found to confuse Avrodoc, because it doesn't properly strip the leading spaces on the comment lines, and we then get something silly like the entire docstring in a markdown code block when we generate the API docs. Maybe if we switch over to Pandoc or something else for doing the API docs, we can change this.

adamnovak: >Also, a general comment about the documentation: words are fundamentally a bad way to describe graphical concepts. I'm glad to see the documentation come into place, but many of these concepts can be more clearly described via annotated illustrations.

I would like to have a `doc` directory where we can put pictures of things, associated long Markdown writeups that embed them, and maybe even slideshows, but I think we should settle on the structure and layout of that in a separate PR.

src/main/resources/avro/common.avdl
```diff
@@ -18,33 +337,112 @@ enum Strand {
...
+}
+
+/**
+A `Path` is an ordered list of `Segment`s. In general any contiguous path
```
fnothaft: How is ordering defined for `Segment`s? I assume order implies the order in which the segments were observed, but this should be explicitly stated. If not specified, someone will eventually assume reference sort order. E.g., let's say that I've got three sequences `A`, `B`, `C` each of which are 1000bp long and describe the first 3000bp of reference sequence `myRef`. If I see a translocation where `A` is moved to be between `B` and `C`, the list inside of `Path` will be `{ B, A, C }`, correct?

src/main/resources/avro/common.avdl
adamnovak: It's a string because the other IDs in the API are strings. Implementations could use integers, UUIDs, accessions, or whatever else they want.

src/main/resources/avro/common.avdl
adamnovak: OK, I'm fixing this now.

src/main/resources/avro/reads.avdl
```diff
@@ -177,7 +183,41 @@ record LinearAlignment {
...
+  The mapping quality of this alignment. Represents how likely
+  the read maps to this position as opposed to other locations.
+  */
+  union { null, int } mappingQuality = null;
+
+  /**
+  Represents the local alignment of this sequence (alignment matches, indels,
+  etc) versus the `Path`.
+  */
+  array<CigarUnit> cigar = [];
```
fnothaft: Implementation question: can we require that a cigar operator will be restricted to mapping to a single `Path` in the `Segment`? If I've got a 100bp read that maps perfectly to two 50bp segments, I think that it'd be preferable to have `cigar = { 50=, 50= }` instead of `cigar = { 100=  }`.

src/main/resources/avro/common.avdl
adamnovak: The `Segment` order in a `Path` is the order in which you see those `Segment`s when walking from one end of the `Path` to the other. I can clarify that in the comment.

Unfortunately, we don't let you just take `Segment`s of a single `Sequence` and throw them into a `Path` in whatever order you want. See the bit about "no novel adjacencies"?

There needs to be a path in the sequence graph that your `Path` object is following, so if you want to see B, then A, then C, you need to add new (0-length) sequences for the novel adjacencies in that translocation, and have (0-length) `Segment`s in your `Path` saying that you follow those sequences. So it would really be ` { B, B-to-A adjacency, A, A-to-C adjacency, C}`.

src/main/resources/avro/variants.avdl
```diff
@@ -120,30 +168,66 @@ record Call {
...
   value of ["A", "C"], and the genotype was [2, 1], that would mean the call
   represented the heterozygous value "CA" for this variant. If the genotype
   was instead [0, 1] the represented value would be "TA". Ordering of the
   genotype values is important if the phaseset field is present.
+
+  If the API server supports the "classic" mode, and this `Call` is associated
+  with a `Variant`, this field must not be null.
   */
-  array<int> genotype = [];
+  union { null, array<int> } genotype = null;
```
fnothaft: Why are we encoding behavior via a `null` here? This is clunky.

src/main/resources/avro/variants.avdl
```diff
@@ -120,30 +168,66 @@ record Call {
...
+  For example, if one `Call` has `genotype` [0, 1], another is associated with
+  `Allele` 3, and both have a `phaseset` of ["maternal", "paternal"], then one
+  DNA molecule carries 0, 3 and another carries 1, 3.
+
+  If the API server supports the "classic" mode, and this `Call` is associated
+  with a `Variant`, this field must not be an array. If the API server supports
+  the "graph" mode and not the "classic" mode, it must not return strings here.
   */
-  union { null, string } phaseset = null;
+  union { null, string, array<string> } phaseset = null;
```
fnothaft: Why not just make this an `array<string>`? I _really_ feel that encoding the behavior of the variant (e.g., allelic vs. genotypic) via type polymorphism is just going to make things difficult to use for developers.

src/main/resources/avro/variants.avdl
```diff
@@ -180,8 +264,10 @@ record Variant {
   /**
   The reference on which this variant occurs.
   (e.g. `chr20` or `X`)
+
+  If the API server supports the "classic" mode, this field must not be null.
   */
-  string referenceName;
```
fnothaft: Shouldn't all of this `referenceName, start, end` data be encapsulated in a `Position` object that can be nulled, instead of requiring each field to be nulled?

lh3:
I also think we should split `Call`. I actually more like to add `GenotypeCall` and gradually deprecate `Call`, but I wouldn't insist on this. As to `Variant`, are we regarding a variant in the VCF/classical mode exactly the same as a variant in the graph+allelic mode?

src/main/resources/avro/reads.avdl
adamnovak: I think you mean single `Segment` in a `Path`, and I'm not sure what we would gain by that. The `Segment`s in the `Path` are required to be connected by reference adjacencies, and we don't want end join adjacencies between bases to be second-class relative to on-the-same-sequence adjacencies between bases.

If I map perfectly to one end of a known translocation, I should have a CIGAR of one run of matches, and a `Path` explaining that I map to one end of a known translocation. Right?

benedictpaten:
Might I propose @lh3 and @jeromekeller that we explore splitting these
objects as a subsequent pull request? For sake of getting things done, we
really need to get something accepted soon - we have a developer hired now
to build this into the reference server! I suspect that as we implement it
we will be proposing changes to improve it. As I say, I'm currently the
only +1 on this, and it would be great for others to feel that this is at
least a good-enough starting point on which we can iterate.

src/main/resources/avro/variants.avdl
adamnovak: It is a bit clunky. We need to be able to not have a `genotype` for a call that's there to give a copy number for an `Allele`. If we split up `Call`s and `AlleleCall`s, this will go away.

src/main/resources/avro/variants.avdl
adamnovak: While an `array<string>` would be able to do everything, I wanted to keep the old values as allowed options for backwards compatibility. If we change this to only be an `array<string>`, existing `Call` objects will have to be converted. Is that worth requiring do you think?

src/main/resources/avro/variants.avdl
adamnovak: It would have to be a `Segment` to hold the end, and then if each `Variant` has a `Segment`, why not just use a `Path` and support the whole graph model?

I kept the original fields around for backwards compatibility. Anything that was a valid `Variant` before this PR should be a valid `Variant` afterwards.

src/main/resources/avro/variants.avdl
diekhans: +1 on this

benedictpaten:
On Wed, Feb 11, 2015 at 9:57 AM, Heng Li <notifications@github.com> wrote:

> I also think we should split Call. I actually more like to add
> GenotypeCall and gradually deprecate Call, but I wouldn't insist on this.
> As to Variant, are we regarding a variant in the VCF/classical mode
> exactly the same as a variant in the graph+allelic mode?
>

That was the intention. There is some weirdness. In the vcf/classic model a
site is a set of edits to a reference substring. In the graph model a
variant is a collection of alleles - they don't have to be homologous,
because we don't know how to define homologous in the graph world. I think
@richarddurbin has made the point repeatedly, but homology is difficult to
define and not transitive when talking about the relationship between
substrings (rather than individual bases). I'd welcome a proposal to
clarify this.

src/main/resources/avro/variants.avdl
benedictpaten: On Wed, Feb 11, 2015 at 10:07 AM, Mark Diekhans <notifications@github.com>
wrote:

> In src/main/resources/avro/variants.avdl
> <https://github.com/ga4gh/schemas/pull/238#discussion_r24517372>:
>
> >    */
> > -  union { null, string } phaseset = null;
> > +  union { null, string, array<string> } phaseset = null;
>
> +1 on this Frank Austin Nothaft <notifications@github.com> writes:
>

No, because this breaks backwards compatibility. We're all for breaking
backwards compatibility down the road - but it needs to be in separate pull
requests. This is supposed to be a minimal set of changes.

src/main/resources/avro/variants.avdl
```diff
@@ -189,20 +275,45 @@ record Variant {
...
+
+  /**
+  The IDs of the reference and alternate `Allele`s for this `Variant`. The first
+  is the reference allele, and any others are alternates. If this field is set
+  along with `referenceName`, `start`, `end`, `referenceBases`, and/or
+  `alternateBases`, those fields must agree with the `Allele`s given here.
+
+  If the API server supports the "graph" mode, this field must not be null.
+  */
+  union { null, array<string> } alleleIds;
```
fnothaft: I thought that in graph mode, the `alternateBases` and `referenceBases` needed to be `null`? If so, then what does agreement mean?

> If this field is set along with `referenceName`, `start`, `end`, `referenceBases`, and/or `alternateBases`, those fields must agree with the `Allele`s given here.

src/main/resources/avro/variants.avdl
diekhans: Good point on compatibility, need to move forward in small
steps.  I withdraw my +1.  Lets get this merged so we can start
implementing and then evolve it.

Everything should be considered vaporware until it works in both
the reference server and has conformance tests.

lh3: I still think it is important for someone to write down code for a few common queries, which I don't know for now. If we cannot even reach this step, we will not be able to implement the APIs.

benedictpaten: On Wed, Feb 11, 2015 at 10:50 AM, Heng Li <notifications@github.com> wrote:

> In src/main/resources/avro/variants.avdl
> <https://github.com/ga4gh/schemas/pull/238#discussion_r24521388>:
>
> >    */
> > -  union { null, string } phaseset = null;
> > +  union { null, string, array<string> } phaseset = null;
>
> I still think it is important for someone to write down code for a few
> common queries, which I don't know for now. If we cannot even reach this
> step, we will not be able to implement the APIs.
>
Agreed. We'll do this.

src/main/resources/avro/variants.avdl
adamnovak: Nope. They have to be non-null if the server supports "calssic" mode, but they don't have to be null if the server supports "graph" mode, because a server can support "classic" and "graph" mode at the same time. 

You can have both groups of fields set; they just need to agree with each other (which in practice means you can't use some of the features of the graph reference).

adamnovak:
OK, by popular demand I have split `Call` into `Call` and `AlleleCall`, and added a `searchAlleleCalls()` method.

This is simpler in that `Call`s will always have genotypes, and `AlleleCall`s will always have `Allele` copy number assertions (possibly associated with `Variant`s). This lets the objects be simpler and removes a bunch of constraints about what needs to be set in what mode.

It's more complicated in that if you want all the information for a `CallSet` you need to use both `searchCalls()` to download any genotypes and `searchAlleleCalls()` to download `Allele` copy number assertions. I've added the constraint that the two views need to be consistent. I've also required that you serve `Call`s for every `Variant` you have `AlleleCall`s for if you support "classic" mode, and visa versa if you support "graph" mode, so you should be able to just query the kind of call that you want and get a more or less complete view of the `CallSet`.

haussler:
Going back to last night's comments by Heng, I've tried to craft a simple
beginner's introduction to what we are doing for the benefit of those
outside our group. It is pasted below and attached in Word. Heng, Richard,
Adam others, please replace or edit as needed, pointing to diagrams and
examples (I agree with Adam these are essential). I think we need something
that gives a high-level intro and is pointed to from the schemas. -D

The allelic model of genetic variation



The field of genetics was born before we could read genomes, yet through
exceedingly clever experiments and careful observations, the pioneers of
genetics were able to form a very accurate model of the basic structure of
genetic information. We refer to this model as the genotypic model.  In the
genotypic model, (1) each individual in a diploid population has a set of
chromosomes that may be numbered 1, …, *n*,  (2) each individual carries
two copies of each of these *n* chromosomes, the sex chromosomes and
mitochondria being considered a special case, and (3) each chromosome copy
consists of a defined linear (or circular) sequence of sites, where each
site carries one of a finite set of possible alleles. In its most
fine-grained form, each site consists of a single basepair of DNA. While
this model has enormous conceptual and practical advantages, it is an
abstraction that is not completely accurate, and this has been a problem
for the field.


The more accurate model is simple: each individual has a genome, where a
genome is simply a set of molecules of double-stranded DNA. In a real
genome, pieces of DNA that represent alleles of particular sites in the
genotypic model occasionally appear in different numbers, different
locations, and different orientations, and this matters. Also, completely
novel DNA is observed. As we learn to read genomes with increasing
fidelity, the inadequacies of the genotypic model become more apparent. We
are headed toward an era in which complete and highly accurate genomes will
be commonplace in genetics. We might contemplate simply replacing the
genotypic model with the direct observation of a complete genome. However,
it will never be practical to base all genetic analysis directly on the
reading of an entire genome. For one thing, a genome is too big and
complicated to be treated as a unit. Moreover, despite its shortcomings,
the genotypic model remains a very powerful and useful abstraction.



The allelic model of genetic variation is an abstraction that sits at a
level below the genotypic model, but above the model that merely represents
a genome as a set of molecules of double-stranded DNA. Instead of a single
linear coordinate system that is used to demarcate multi-allelic sites
along the chromosomes, in the allelic model the genome variation present in
the population is described by a network- or graph-like structure,
representing segments of DNA that appear in typical genomes in the
population, along with their common alternatives and common variations in
number, order and orientation [diagram].  We refer to this graph as a
(genome) reference structure. A chromosome is a path through this reference
structure, and a genome is a set of chromosomes.


Instead of sites, the elementary units of a reference structure are
positions. Each position consists of an orientation and a single base of
DNA. The orientation tells us whether that base is on the top or bottom
strand, and thus indicates on which side of it we expect to find the next
base when moving in the standard 5’ to 3’ direction. When a component of
the reference structure is linear, with no branching or “bubble”
alternatives, one can think of it in conventional terms as a series of
1-base sites defining the coordinate system plus the typical base content
of a reference chromosome, e.g. as in the “primary path” of the current
human genome reference of the International Nucleotide Sequence Database
Consortium, version GRCh38. Non-linear aspects of the reference structure,
e.g. “bubbles” that come off of the primary path, can be thought of as
alternate haplotypes, like the alternate haplotypes that are defined as an
adjunct to the primary paths of GRCh38. One difference, however, is that a
general reference structure allows cycles, so the path that defines a
particular individual’s chromosome may repeat a part of the reference
structure an arbitrary number of times.


In the allelic model, an allele is defined as a sequence of positions that
form a (sub)path in the graph structure, i.e. such that each position is
allowed by the edges of the graph structure to follow the previous
position. One can think of an allele roughly as the sequence of DNA spelled
out by a particular sequence of positions, but it is more than that. The
same sequence of DNA may be spelled out other places in the graph as well.
When we say that a genome has a particular allele, we are asserting that
the path for one of the DNA molecules in that genome passes through this
sequence of positions, contiguously, at some point along its extent, either
in one direction (strand) or the other.  This means that the individual’s
genome contains whatever biological feature, such as a particular form of
an exon, gene, etc., that is represented by this sequence of positions in
the reference structure. If we say the allele appears twice in the
individual’s genome then this asserts that the set of paths representing
that individual’s chromosomes, considered in aggregate, pass through this
particular sequence of positions twice. It is not asserted that these
allele traversals occur in two separate DNA molecules, or in any particular
orientations. Two alleles, be they the same or different, are not assumed
to be either phased or unphased by default. Phasing information must be
explicitly specified in addition to allelic information.


The allelic model does not have the classical genetics concepts of “site”
or “genotype at a site”. These concepts must be added on top of the allelic
model if needed.  In the allelic model, each position, and hence each
allele, has just one DNA sequence associated with it.   One might define an
“allelotype” as a set of alleles that are asserted to be present in an
individual’s genome, each with a specified copy number, zero being an
allowed copy number. This asserts that the double-stranded DNA molecules of
the individual’s genome follow a set of paths that traverse the sequences
of positions defined by these alleles the required number of times. These
traversals may occur in any combination of order, orientation and phasing.
Other DNA may also be included in the individual’s genome that is not
listed in the allelotype. This contrasts with a “genotype” in which it
asserted that the individual’s double-stranded DNA molecules include a
specific set of (classical) allelic variants in the default order and
orientation, with only the phasing (and possible additional DNA) left
unspecified. In this sense we claim that the classical genotypic model can
be built on top of the allelic model as a special case, by using additional
assumptions or assertions.


The allelic model is a richer foundation for genetics. Using something like
it becomes a requirement when one attempts to analyze and reason about
whole genomes. We hope that the new data schemas and application
programming interfaces we are defining for the allelic model will support
the further development of this model for both research and clinical use.

adamnovak:
Speaking of the importance of images, I've done up a UML class diagram of all the data model records, so we can see my proposed changes (on the right) as well as how the whole API fits together.

http://hgwdev.sdsc.edu/~anovak/images/Global%20UML.png

pgrosu:
@fnothaft , we don't need to use Avro IDL.  We can either change it - which many companies do internally - or use something like WebIDL which provides inheritance capabilities:

http://www.w3.org/TR/WebIDL/

If we keep working around the limitations of Avro, the schema will become cumbersome to support and expand upon.

Paul

jeromekelleher:
I'm in favour of merging this now. It seems clear that the consensus is that we want something like this, but it's also clear that it will take some time to resolve the remaining issues. I would suggest that the most productive approach now would be to merge what we have and start filing issues and PRs against it to resolve the outstanding problems. 

+1.

haussler:
Richard astutely pointed out that there were some obnoxious aspects to the
previous summary of the allelic model. Here is a revised and a bit shorter
version. Further comments, edits very welcome. D

The allelic model of genetic variation



In the textbook genetic model, genetic information in a diploid population
is approximated as follows: (1) each individual has a set of chromosomes
that may be numbered 1, …, *n*,  (2) each individual carries two copies of
each of these *n* chromosomes, the sex chromosomes and mitochondria being
considered a special case, and (3) each chromosome copy consists of a
defined linear (or circular) sequence of sites, where each site carries one
of a finite set of possible alleles. We refer to this model as the
genotypic model.  In its most fine-grained form, each site consists of a
single basepair of DNA. While this model has enormous conceptual and
practical advantages, even prior to genome sequencing, geneticists realized
that it is not always sufficient, and proposed various generalizations. The
allelic model is such a generalization.


Genetic information is carried in a genome consisting of molecules of
double-stranded DNA. Inadequacies in the genotypic model arise because in a
real genome, pieces of DNA that represent alleles of particular sites
occasionally appear in different numbers, different locations, and
different orientations, and this can matter. While sufficient for many
purposes, the genotypic model doesn't handle these cases. As it becomes
more commonplace to read genomes with increasing fidelity, the inadequacies
of the genotypic model become more apparent.


The allelic model of genetic variation is an abstraction that sits at a
level below the genotypic model, but above the trivial model that merely
represents a genome as a set of molecules of double-stranded DNA. Instead
of the linear coordinate system that is used in the genotypic model to
demarcate multi-allelic sites along the chromosomes, in the allelic model
the genome variation present in the population is described by a network-
or graph-like structure, representing segments of DNA that appear in
typical genomes in the population, along with their common alternatives and
common variations in number, order and orientation [diagram].  We refer to
this graph as a (genome) reference structure. A chromosome is a path
through this reference structure, and a genome is a set of chromosomes.


Instead of sites, the elementary units of a reference structure are
positions. Each position consists of an orientation and a single base of
DNA. The orientation tells us whether that base is on the top or bottom
strand, and thus indicates on which side of it we expect to find the next
base when moving in the standard 5’ to 3’ direction. When a component of
the reference structure is linear, with no branching or “bubble”
alternatives, one can think of it in conventional terms as a series of
1-base sites defining the coordinate system plus the typical base content
of a reference chromosome, e.g. as in the “primary path” of the current
human genome reference of the International Nucleotide Sequence Database
Consortium, version GRCh38. Non-linear aspects of the reference structure,
e.g. “bubbles” that come off of the primary path, can be thought of as
alternate haplotypes, like the alternate haplotypes that are defined as an
adjunct to the primary paths of GRCh38. One difference, however, is that a
general reference structure allows cycles, so the path that defines a
particular individual’s chromosome may repeat a part of the reference
structure an arbitrary number of times.


In the allelic model, an allele is defined as a sequence of positions that
form a (sub)path in the graph structure, i.e. such that each position is
allowed by the edges of the graph structure to follow the previous
position. One can think of an allele roughly as the sequence of DNA spelled
out by a particular sequence of positions, but it is more than that. The
same sequence of DNA may be spelled out other places in the graph as well.
When we say that a genome has a particular allele, we are asserting that
the path for one of the DNA molecules of that genome passes through this
sequence of positions, contiguously, at some point along its extent, either
in one direction (strand) or the other.  This means that the individual’s
genome contains whatever biological feature, such as a particular form of
an exon, gene, etc., that is represented by this sequence of positions in
the reference structure. If we say the allele appears twice in the
individual’s genome then this asserts that the set of paths representing
that individual’s chromosomes, considered in aggregate, pass through this
particular sequence of positions twice. It is not asserted that these
allele traversals occur in two separate DNA molecules, or in any particular
orientations. Two alleles, be they the same or different, are not assumed
to be either phased or unphased by default. Phasing information must be
explicitly specified in addition to allelic information.


The allelic model is a generalization because it does not include the
classical genetics concepts of “site” or “genotype at a site”. These
concepts must be added on top of the allelic model in order to derive the
genotypic model.  In the allelic model, each position, and hence each
allele, has just one DNA sequence associated with it.   One might define an
“allelotype” as a set of alleles that are asserted to be present in an
individual’s genome, each with a specified copy number. Zero is an allowed
copy number. An allelotype asserts that the double-stranded DNA molecules
of the individual’s genome follow a set of paths that traverse the
sequences of positions defined by its alleles the required number of times.
These traversals may occur in any combination of order, orientation and
phasing. Other DNA may also be included in the individual’s genome that is
not listed in the allelotype. This contrasts with a “genotype” in which it
asserted that the individual’s double-stranded DNA molecules include a
specific set of (classical) allelic variants in the default order and
orientation, with only the phasing (and possible additional DNA) left
unspecified. Thus, the classical genotypic model can be built on top of the
allelic model as a special case, by using additional assumptions and
assertions.


Being more general, the allelic model is a broader foundation for genetics.
Using something like it often becomes a requirement when one attempts to
analyze and reason about whole genomes. We hope that the new data schemas
and application programming interfaces we are defining for the allelic
model will support the further development of this model for both research
and clinical use.

hershman:
+1, @adamnovak your presentation from today was fantastic

src/main/resources/avro/references.avdl
calbach: I think the segments are explained clearly enough elsewhere. My only suggestion would be that this comment make it clearer that it's referring to the `md5Checksum` field of the associated joined segment. Right now it's slightly ambiguous, and could be interpreted as the MD5 checksum of the `position` record itself.

adamnovak: It's not referring to the `md5checksum` of the associated joined segment; it's referring to the MD5 of the joined segment's `md5checksum` concatenated with the join position as a decimal string and a "+" or "-" character denoting the strand of the join.

I've rewritten this section entirely to specify exactly what you are supposed to stir into your hash. Please let me know if it's still unclear.

calbach: Thanks, much clearer.

jeromekelleher:
Can you squash the commits down please @adamnovak? Looks like we're ready to merge.

adamnovak:
OK I have squashed everything down. I'm going to tag this as ready to merge.

skeenan:
I'm merging this.

src/main/resources/avro/common.avdl
```diff
@@ -2,6 +2,325 @@
...
+called a contig. This sequence is a scaffold.
+
+In its default forward orientation, each basepair has a left or "+" side and a
+right or "-" side. For example, the left side of the T/A basepair in the above
+example lies immediately to the left of this basepair. The left side of the T/A
+basepair is represented by the position (2,+), where the index 2 is the same as
+that of the basepair itself, and the "+" confirms that we mean the default
+forward orientation. The right side is indicated using the "-" orientation. For
+example, the right side of the following G/C base pair is represented by the
+position (3,-).
```
ryan-williams: An slightly confusing coincidence in this example is that, in this sequence of length 7, the index `3` could be interpreted as being measured from either end. 

The asymmetry between the `start` index having the nick come just *before* it while the `end` index has the nick *after* it (when both are read from left-to-right, as the indices are apparently intended to be read) made it hard for me to figure out all that was meant to be implied by the `(3,-)` notation. 

Let me know if that doesn't make sense.

ryan-williams: very good, thorough example overall though, thank you for writing this up!

src/main/resources/avro/common.avdl
```diff
@@ -18,33 +337,116 @@ enum Strand {
...
+`Position` is represented by a sequence name or ID, a base number on that
+sequence (0-based), and a `Strand` to indicate the left or right side.
+
+For example, given the sequence "GTGG", the `Position` on that sequence at
+offset 1 in the forward orientation would be the left side of the T/A base pair.
+The base at this `Position` is "T". Alternately, for offset 1 in the reverse
+orientation, the `Position` would be the right side of the T/A base pair, and
+the base at the `Position` is "A".
+
+Offsets added to a `Position` are interpreted as reading along its strand;
```
ryan-williams: This sentence seems confusing given [line 36](https://github.com/ga4gh/schemas/pull/238/files#diff-f478e5ba9b38667091d5f5a7e1a0fc97R36) above.

The "offset" discussed here, which moves "right to left" (i.e. 5'-to-3') along the negative strand, sounds like it has the opposite behavior from the "positions" discussed previously (also referred to as "indices" on [L50](https://github.com/ga4gh/schemas/pull/238/files#diff-f478e5ba9b38667091d5f5a7e1a0fc97R50)) and the comment on the `position` field below ([L376](https://github.com/ga4gh/schemas/pull/238/files#diff-f478e5ba9b38667091d5f5a7e1a0fc97R376), which says it measures from "the start of the forward strand").

Am I misreading? Are there deliberately different semantics between {"indices", "positions"} vs. "offsets"?

src/main/resources/avro/common.avdl
```diff
@@ -18,33 +337,116 @@ enum Strand {
...
+  `Segment`s covering the sequences onto which this `Segment` is joined are
+  called its "parents", while this segment is a "child".
+
+  Joins may occur on the outer sides of the terminal bases in a sequence: the
+  left side of the base at index 0, and the right side of the base with maximum
+  index. These are the "terminal sides" of the sequence. `startJoin` is the join
+  on the side indicated by `start`, and may only be set if that side is a
+  terminal side. Similarly, `endJoin` is the join on the opposite side of the
+  piece of sequence selected by the segment, and may only be set if that side is
+  a terminal side. The value of `startJoin` or `endJoin`, if set, is the side to
```
ryan-williams:
```
s/value/`strand`/
```
?

src/main/resources/avro/common.avdl
```diff
@@ -18,33 +337,116 @@ enum Strand {
...
+  `length`). If `start` is on the reverse strand, the `Segment` contains the
+  range (`start.position` - `length`, `start.position`]. This is equivalent to
+  starting from the side indicated by `start`, and traversing through that base
+  out to the specified length.
+  */
+  long length;
+
+  /**
+  Start and end `Position`s where this `Segment` attaches to other sequences.
+  Note that the segmentId for start and end might not be the same. The
```
ryan-williams: "segmentId" doesn't appear to be defined anywhere, is it out-of-date terminology?

jeromekelleher:
This PR has been merged @ryan-williams, so the discussion thread has ended. Please open a PR if you'd like to make some updates, or open some issues corresponding to the individual comments.

PR #244, opened 2015-02-25
Indel query support for beacons

mfiume:
It was decided that indel sequences need to be specified explicitly in beacon queries. This pull request captures a language for specifying indel sequences.

### DIFF
--- a/src/main/resources/avro/beacon.avdl
+++ b/src/main/resources/avro/beacon.avdl
@@ -10,7 +10,11 @@ protocol BEACON {
 A request for information about a specific site
 */
 record QueryResource {
-  /** Allele string (might include wildcards). */
+  /** Allele string. Use I<seq> for insertions and Dn for deletions, 
+      where <seq> is the nucleotide sequence inserted after position 
+      and n is a number of nucleotides deleted from the reference 
+      starting at position.  
+   */
   string allele;
 
   /** The chromosome of the request */
@@ -24,18 +28,6 @@ record QueryResource {
 
   /** The name of the targeted population */
   union{ null, string } dataset = null;
-}
-
-
-/**
-Frequencies of alleles
-*/
-record AlleleResource {
-  /** Allele string as per allele property of requests */
-  string allele;
-
-  /** Frequency of the allele. Between 0 and 1 */
-  double frequency;
 }
 
 /**
@@ -153,8 +145,8 @@ record ResponseResource {
   otherwise. */
   string exists;
 
-  /** frequencies */
-  array<AlleleResource> frequencies = [];
+  /** frequency */
+  double frequency;
 
   /** # observed */
   union{ null, int } observed = null;


awz:
-1.    I am relatively opposed to this change.    It seems much simpler to represent indels as questions about sequences.   So you can ask queries about indels now by asking about the presence of a sequence with the indel already incorporated.    Thoughts?  Should this be discussed on the mailing list?

maximilianh:
+1

We can't refer people to the VCF doc on the I/D syntax (can't find anything
similar on
http://www.1000genomes.org/wiki/Analysis/Variant%20Call%20Format/vcf-variant-call-format-version-40),
but given that requests for insertions are relatively rare, I guess it's OK
to overload the allele field.

awz:
@mfiume  - On closer inspection this is fine.   I wonder if we shouldn't allow specification of substitution, indel and deletions for the separate alleles?     So if you have an indel and a substitution or any combination of these three you can still make a Beacon request?    It seems that you should be able to query for apoe3 homozygotes or more complicated situations.

awz:
Changing my "vote" to +1 on closer inspection of the actual commit.

adamnovak:
I'll give this a +1, but it might be good to note that (since I and D are now reserved) you can't use DNA wildcard characters in your queries.

maximilianh:
Hi Marc,
when I just started to implement this, I found that since the beacon
queries deviate from VCF-style descriptions, a few VCF indels can't be
expressed as beacon queries.

quick summary:
- VCF expresses insertions and deletions as "replacement" of sequence, with
a table (chrom, position, ID, reference allele, alternate allele).
- Beacon allows insertions at a position with a given sequence and
deletions at a position with a given length.

Here are some examples from the ICGC VCF which cannot be translated
directly to beacon queries:

1       5121469 MU32560333      TAATGTCCAGACTTGC        TTAATGTCTTACT
1       71228196        MU32559126      TCGGCACTTTG     TGA
1       119929244       MU23604552      GG      CT

How common is this?

I took the first 1M lines in the ICGC file
simple_somatic_mutation.aggregated.vcf.gz.
25k of these are deletions (ref length != 1, alt length == 1), 24k of these
are insertions (ref length==1, alt length != 1) and 86 are "replacements"
like the ones above (ref length!=1 and alt length !=1). So we're not fully
compatible with VCF, but on the other hand, 86 out of a 1M is not that
much. I don't have a clear opinion, just wanted to report this potential
issue.

cheers
Max




zcat projects/bd2k/beacon/simple_somatic_mutation.aggregated.vcf.gz | head
-n 1000000 | grep -v ^# | tawk '(length($4)==1 && length($5)!=1)' | wc -l

ekg:
The beacon system should support alleles that are arbitrary replacements of one sequence with another. That's the most natural way to describe variation. It doesn't involve particular classification of the allele type.

As @maximilianh finds in the ICGC somatic set, examples like these are be very common in practice. But even more essentially, some things are hard to express as simple indels and SNPs. Small haplotypes can have a particular functional effect that is more than the aggregate of their parts, so talking about component point mutations is often problematic.

[I've collected a few examples that demonstrate the importance of reporting and sharing haplotypes rather than point mutations when the goal is related to understanding genomic function](https://docs.google.com/presentation/d/1DiKqqLEvMc24VJ8ocnvlzVQ61X4Eh7iaKrse7vpLF3g/edit?usp=sharing). I hope the significance of this for the beacon interface is clear.

iqbal-lab:
+1 for Erik

mfiume:
Thanks for pointing those cases out @maximilianh. We are left with deciding a syntax for expressing replacements, hopefully in a way that doesn't make SNV and simple indels more complex.

Beacon queries have an allele field, which we were hesitant to split into "ref" and "alt" as VCF does. One possibility is to have a syntax like allele=REF>ALT, where REF> is optional for SNVs. The one drawback is that the reference allele needs to be known for indels, but that seems like a reasonable trade-off for increased expressibility.

Thoughts?

maximilianh:
+1 for the ref allele being optional for SNVs.

I slightly favor two separate fields (ref and alt) with one being usually
not set instead of an overloaded field allele with a special character in
it, but that's really a matter of taste, I'm fine with either of the two.

ekg:
@mfiume At minimum you need a reference genome identifier, a coordinate, a reference length, and the alternate sequence. This allows you to deduce what the allele was unambiguously.

I apologize for jumping in here as I don't understand the reasons why the group didn't want to encode both ref and alt, but it seems extremely natural to describe the genomic context somehow and then say what it was replaced by. It's also completely general to any allele type.

@maximilianh How would you know which SNP allele is present if it's not provided? You could have any non-reference sequence, no? Sorry again for any confusion.

maximilianh:
Hi Erik,
the beacon system was supposed to be as simple as possible, to be used
manually, for only a handful of alleles and allowed - until the latest
change - to query only SNVs. It does not target non-reference genomes. For
this purpose, as the reference ID is provided in the query, the reference
allele was implicitly specified by (chrom, position).

For the case of SNV queries in a reference genome, we do not need the
reference allele and could make it optional. It would be ignored anyways by
the implementation.

If we require the reference allele for SNVs, than the main use case
(SNP-lookups) would require clients to have an indexed reference genome at
hand, even though many users might be completely fine with just the
position, e.g. when they have a list of dbSNP locations.

cheers
Max

ekg:
@maximilianh Sorry, I'm perplexed. What is the objective of the beacon API? I thought it was to share the presence or absence of a particular allele without sharing entire genomes.

In the 2500 individuals in the 1000GP phase3, ~2% of SNP alleles are multiallelic. This number will invariably increase as more genomes are sequenced. And that's just for SNPs...

> If we require the reference allele for SNVs, than the main use case (SNP-lookups) would require clients to have an indexed reference genome at hand, even though many users might be completely fine with just the position, e.g. when they have a list of dbSNP locations.

You won't need an indexed reference genome if you provide coordinates (reference genome name, chromosome, position, reference-relative length) and the alternate allele. The ref allele would further simplify things for people without the indexed reference available, at very little cost.

How are people supposed to use this? What's the vision for the API in the wild? Sorry for my confusion.

maximilianh:
>
> You won't need an indexed reference genome if you provide coordinates
> (reference genome name, chromosome, position, reference-relative length)
> and the alternate allele. The ref allele would further simplify things for
> people without the indexed reference available, at very little cost.
>
Yes, this is pretty much what Marc suggested. We already have (reference
genome, chromosome, position, alternate allele) right now. I'm sorry, I
think we agree on everything, there is just maybe some confusion. What I
meant is that clients should be able to query for single-base pair SNVs
with just (chrom, position, alternateAllele). Does this make sense?

ekg:
I think we agree but there is still a point of confusion. You need to keep the reference-relative length to handle anything other than SNPs. If you do this then the complex and indel alleles that you mentioned before aren't a problem at all. So queries would be of the form `(reference_name, chromosome, position, reference_relative_length, alternate_allele)`. For SNPs `reference_length == 1` and the alternate allele is just one base. This is basically VCF's ref/alt pattern, and so I was also confused why you said you didn't want to use this.

maximilianh:
Yes, sorry, what I meant is that I didn't want this for SNPs. We agree on
everything. We only need to decide now if we're using reference length or
reference allele in the beacon. I don't have a strong opinion.

richarddurbin:
+1 for the Beacon API using the same representation of variant alleles as the Variation API.  The replacement representation 
	is unambiguous and well established.  Furthermore it doesn't say much for us as a standards organisation if we
	can't use our own standards in our own interfaces.

On 5 Mar 2015, at 14:07, Maximilian Haeussler <notifications@github.com> wrote:

> Yes, sorry, what I meant is that I didn't want this for SNPs. We agree on
> everything. We only need to decide now if we're using reference length or
> reference allele in the beacon. I don't have a strong opinion.
> —
> 




-- 

maximilianh:
Sorry if this is slightly out of topic:

As beacon is most likely going the the VCF way, at some point I'll have to
convert the LOVD variants to VCF format.
The LOVD data looks like this: hg19.chr21:45711096_45711098(
AIRE:c.995+3_995+5delinsTAT)

Does anyone happen to know a library that could convert these descriptions
to VCF?

cheers
Max

>

mfiume:
Thanks @richarddurbin. Based on your suggestion, we can model specification of query allele after [GAVariant](http://ga4gh.org/documentation/api/v0.5.1/ga4gh_api.html#/schema/%2FUsers%2Fkeenan%2FDropbox%2Fgit-checkouts%2Fschemas%2Fsrc%2Fmain%2Fresources%2Favro%2Ftarget%2Fall.avpr/org.ga4gh.GAVariant). That is, the Beacon query can have "referenceBases" and "alternateBases" fields.

Would you require the "referenceBases" field for SNVs (since the reference base is inferred by assembly, chr, and position)? Just trying to keep Beacon simple.

Thanks very much for your guidance.

awz:
"The beacon system should support alleles that are arbitrary replacements
of one sequence with another. That's the most natural way to describe
variation. It doesn't involve particular classification of the allele type."

I completely agree with the above.

On Thu, Mar 5, 2015 at 5:19 AM, Erik Garrison <notifications@github.com>
wrote:

> The beacon system should support alleles that are arbitrary replacements
> of one sequence with another. That's the most natural way to describe
> variation. It doesn't involve particular classification of the allele type.
>
> As @maximilianh <https://github.com/maximilianh> finds in the ICGC
> somatic set, examples like these are be very common in practice. But even
> more essentially, some things are hard to express as simple indels and
> SNPs. Small haplotypes can have a particular functional effect that is more
> than the aggregate of their parts, so talking about component point
> mutations is often problematic.
>
> I've collected a few examples that demonstrate the importance of reporting
> and sharing haplotypes rather than point mutations when the goal is related
> to understanding genomic function
> <https://docs.google.com/presentation/d/1DiKqqLEvMc24VJ8ocnvlzVQ61X4Eh7iaKrse7vpLF3g/edit?usp=sharing>.
> I hope the significance of this for the beacon interface is clear.
>
> —
> <https://github.com/ga4gh/schemas/pull/244#issuecomment-77339491>.
>



-- 
Alexander (Sasha) Wait Zaranek, PhD.  <sasha@curoverse.com>
Director Informatics, Personal Genome Project, Harvard Medical School
Chief Scientist, Curoverse Inc.  Founder http://arvados.org

Get paid to write free software: https://curoverse.com/about/jobs/

richarddurbin:
Yes I would require the reference base for single base variants, even though it is inferrable from reference, chromosome and position.  It is reasonable that if someone
is asking about a variant allele they know the reference allele, so this is not much to ask.  From experience this also helps catch errors, particularly strand errors which 
always lead to a mis-specification of the reference base.  So I would have the server check that the specified reference matches.

Richard

On 5 Mar 2015, at 17:32, Marc Fiume <notifications@github.com> wrote:

> Thanks @richarddurbin. Based on your suggestion, we can model specification of query allele after GAVariant. That is, the Beacon query can have "referenceBases" and "alternateBases" fields.
> 
> Would you require the "referenceBases" field for SNVs (since the reference base is inferred by assembly, chr, and position)? Just trying to keep Beacon simple.
> 
> Thanks very much for your guidance.
> 
> —
> 




-- 

maximilianh:
I'd agree with Richard if the end users were bioinformatics programmers.
But it seems to me that many users of beacons are less proficient with
genomic indexes. Also, when querying the beacon manually (which was the
original use case, if I remember well), it is a lot quicker to type a
position, instead of looking up the reference allele for the position with
a genome browser.
I think it would be nice if the API allowed skipping the reference alleles
for SNPs. It could involve a special value for this parameter.

deannachurch:
But the problem is if you don't require the users to provide enough information  you run the risk of returning them incorrect information. I especially worry about this as we move towards a time when people may be using different assemblies (GRCh37 vs. GRCh38). You have to balance ease of use with correctness (the later being more important).

ekg:
@mfiume As a potential user of the API, I think it's simpler if you have the same interface for all types of variants. So I would be comfortable if the referenceBases field were required for SNPs as well as indels.

maximilianh:
Hi Deanna,
thanks, yes, the beacon has a "reference" parameter that specifies the
version of the reference (GRCh37, GRCh38 or others).
cheers
Max

deannachurch:
Sorry I wasn't clearer in my earlier comment. Requiring the reference allele is a good check to ensure you are using the assembly you think you are using. I've already been bitten by accidentally running the wrong version of VEP (which defaults to GRCh38)- but caught the mistake as I had these checks in place.

maximilianh:
Yes, the current beacon reply always shows again the reference that was
used so it is shown together with the result.

ekg:
You'll greatly improve the redundancy by using the Variant model (which also includes the reference allele). The reference version isn't enough to actually catch the errors unless the user remembers the reference version.

nouyang-curoverse:
Hi all -- Just catching up on all this and not quite clear what we decided. Can anyone explain the current thinking with some simple concrete examples of how we plan to support indels?

Let's say that our coordinate system genome is
    `ATTG` = positions `0 1 2 3`

How do you specify a query that you would like to search for any occurrences of the sequence
1)   `ATTTG`
or
2)    `ATG`

in the population that start at position `0`? Furthermore, let's say that actually our query begins with an indel. How do you specify that you would like to search for 
3) `TG`
or
4) `TTTG`
?

Finally, let's say that something more complex happens, like an SNP after an Indel.
5) `ATTTA`
How do you specify that query using our current schema?

Thanks!

nouyang-curoverse:
@ekg, I left some comments on your "use case collection" powerpoint. I hope you can see them -- if not I have copied them below:

> Hey ekg -- I think a way to describe this in a less jargon-heavy way (more friendly to CS ppl vs bio ppl :]) is:
> Proteins are encoded by three bases (codons) between a start and stop codon, if the first base is deleted, all the proteins are screwed up by an off-by-one error and you have a sad gene. This is called "frameshift mutation".
> However, you could easily have another two-base deletion further along which mitigates the initial deletion, resulting in a happy gene. This happens frequently in real-life biology. --> aka "frame-restoring indel"
> I'm not super-sure what implications you have in mind for the beacon interface, if you can clarify that I'd much appreciate it. 

disclaimer: I'm also not super confident about the biological claims I'm making, just trying to understand ekg's slideshow

ekg:
This is great, and exactly what I mean!

Sometimes specific haplotypes have particular significance for genomic function (and thus health). An example would be a common allele seen in the BLM gene which is responsible for bloom syndrome--- a small block substitution of one chunk of bases for another. And it isn't easy to express using particular classification terms. So it's easier to talk about as this as a replacement of one sequence for another. This generalizes just fine for SNPs, although in many cases the extra information required to express it may seem redundant.

PR #246, opened 2015-03-04
proposed sequence annotation API

diekhans:
Pull request for sequence annotation from topic branch.

This impacts several other task teams, so please examine and vote.

### DIFF
--- a/src/main/resources/avro/sequenceAnnotations.avdl
+++ b/src/main/resources/avro/sequenceAnnotations.avdl
@@ -1,0 +1,167 @@
+@namespace("org.ga4gh.models")
+/**
+This protocol defines annotations on GA4GH genomic sequences It includes two
+types of annotations: continuous and discrete hierarchical.
+
+The discrete hierarchical annotations are derived from the Sequence Ontology
+(SO) and GFF3 work 
+
+   http://www.sequenceontology.org/gff3.shtml
+
+The goal is to be able to store annotations using the GFF3 and SO conceptual
+model, although there is not necessarly a one-to-one mapping in Avro records
+to GFF3 records.
+
+The minimum requirement is to be able to accurately represent the current
+state of the art annotation data and the full SO model.  Feature is the
+core generic record which corresponds to the a GFF3 record.
+*/
+protocol SequenceAnnotations {
+
+  import idl "common.avdl";
+  import idl "ontologies.avdl";
+
+  /**
+  Identifier from a public database
+  */
+  record ExternalIdentifier {
+    /**
+    The source of the identifier.
+    (e.g. `Ensembl`)
+    */
+    string database;
+
+    /**
+    The ID defined by the external database.
+    (e.g. `ENST00000000000`)
+    */
+    string identifier;
+
+    /**
+    The version of the object or the database
+    (e.g. `78`)
+    */
+    string version;
+  }
+
+  /**
+  Type defining a collection of attributes associated with various protocol
+  records.  Each attribute is a name that maps to an array of one or more
+  values.  Values can be strings, external identifiers, or ontology terms.
+  Values should be split into the array elements instead of using a separator
+  syntax that needs to parsed.
+  */
+  record Attributes {
+    map<array<union {string, ExternalIdentifier, OntologyTerm}>> vals = {};
+  }
+  
+  /**
+  Continuous numerical annotation along a path.
+  */
+  record Wiggle {
+    /**
+    The path for the region being annotated.
+    Note that we are fusing parts of UCSC BedGraph and Wiggle syntax.
+    The region is being fully annotated, but can be divided into bins.
+    If you have gaps, you need to define a sequence of such Wiggles.
+    */
+    Path path;
+
+    /**
+    The values associated to this region.
+    If this list contains _count_ elements, then the region is divided
+    as cleanly as possible into _count_ bins of equal width.
+    We thus define a numerical function:
+
+    value(position):
+      if position < start or position >= start + length:
+        return None
+      else:
+        return array[floor((position - start)*count/length)]
+    */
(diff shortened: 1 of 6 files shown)


mbaudis:
Courageous effort, and much wished for progress! However, at the moment, I'm not sure about all points. Things immediately coming to mind:

* I support moving the OntologyTerm from metadata to common. However, its structure is being much discussed and probably will change.
* Region resides in common, but doesn't have an associated reference. While I would understand this implementation wise (you may not want to name your reference for each SNP, CNV etc.) - shouldn't it be somehow guaranteed in the schema definition? Or may I have missed a strict inheritance path?

So - enlightenment I hope for!

diekhans:
I believe that common needs to be removed all together and
specific modules create for ontologies, coordinates,
cigar. Modules like `common' and `utils' do nothing to aid
organization and understand.  They tend to grow to make things
even more confusing.

Region contains Position which contains the references.

We *must* get things implemented for them to be real.

mbaudis:
+1

    Region contains Position which contains the references.

O.k., I didn't follow this (would also be a nice place for a comment there, for Avro blind ppl. like me ...).

As for the detangling, at least I'm glad if the ontology format specification gets out of metadata...

buske:
The Matchmaker group is also interested in implementing paging, so we'll likely adopt whatever naming you use here.

kellrott:
+1
The Genotype2Phenotype API is already connecting to elements created in the branch. It will be nice to have these concepts join the master branch.

src/main/resources/avro/sequenceAnnotations.avdl
```diff
@@ -0,0 +1,170 @@
...
+    The version of the object or the database
+    (e.g. `78`)
+    */
+    string version;
+  }
+
+  /**
+  Type defining a collection of attributes associated with various protocol
+  records.  Each attribute is a name that maps to an array of one or more
+  values.  To work around Avro restrictions, three fields are define, one for
```
adamnovak: This should say "defined", not "define".

src/main/resources/avro/sequenceAnnotations.avdl
```diff
@@ -0,0 +1,170 @@
...
+  each type of value.  However attribute names must be unique across all three
+  fields.  If possible, multi-part string values should be split into array
+  elements instead of adding separator syntax that need to parsed.
```
adamnovak: "needs to be parsed"

adamnovak:
I'm +1 on what's here, but we appear to be missing the wiggle methods (as is noted in the code). Those should either be added on to this PR, or come soon after.

nlwashington:
+1.  We intend to use it for G2P.  I would recommend, however, that the record name of Feature change to GenomicFeature.  This is to make it more descriptive, and avoid conflict with the matchmaker API (that also has Feature record for phenotype -- though that should also probably be updated too).

buske:
@nlwashington Happy to update the Matchmaker term to something more explicit.

pgrosu:
General confirmation question: Matchmaker API is still part of [schemas](https://github.com/ga4gh/schemas), or does it have another github location for discussions?

Thanks,
Paul

nlwashington:
@buske that sounds great.  perhaps PhenotypicFeature?

buske:
@pgrosu We have [another github repo](https://github.com/MatchMakerExchange/mme-apis/) for internal discussions, but I've been porting our API to Avro within schemas (see #258) and have occasionally been posting questions to the schemas issue tracker.

@nlwashington Sounds good. I'll change this within our current outstanding PR #258, since I don't think we're set on any of the object names (we don't use them in our JSON API, so they're a necessary result of porting to Avro).

buske:
@nlwashington Realized we also have a `GenomicFeature` class, so that would be another name collision. I'm not sure what the right way to proceed is. I'm happy to prefix all our classes with `MME`, or is the different namespace sufficient?: `@namespace("org.ga4gh.match")`

diekhans:
A one point before the sequence annotation records had a prefix
(SeqAnn*), but that was removed as not-GA4GH like.  While namespaces
can prevent collisions in the software, it can still be confusing.

We need to do a global review of the names and other patterns
in the APIs, but first we need to get all the current work in so
we have something to review!

pgrosu:
Thank you @buske for bringing me up to speed on this and for the link.

Paul

benedictpaten:
I like 99% of what's there. I'm nearly 1+, however, I'm a bit unsure about the QueryRange object. Would it not be better to just use a path? 

I also agree with @adamnovak  that having the wiggle methods in would be nice.

diekhans:
Sounds good.  Edit to you liking..



... Sent from my computer phone

-----Original Message-----

benedictpaten:
Not trying to weasel out of work, but I'd be happy if you could just drop
the QueryRange object and change lines:

/**+ If specified, this query matches only annotations which overlap this
range.+ */+ union { null, QueryRange } range = null;

to

/**+ If specified, this query matches only annotations which overlap this
path.+ */+ union { null, Path } path = null;
I can do this - but I'd need to create a new pull request - much better if
you keep your momentum and +1s!

Note for others reading this - Mark uses path to specify wiggles/features,
so this is consistent.

diekhans:
Updated to use Path (sorry, delayed by having a lame Internet connection).

We need to get all the methods in there, but I think get the
data structures in the master so others can build on it is
really important.

I do expect this to change as the other groups link with it.

benedictpaten:
1+, the other methods can be added later.

Benedict
(from phone)

jeromekelleher:
It looks like there are enough +1s here. Can you rebase off upstream/master please and squash the commits please @diekhans?

jeromekelleher:
Thanks @diekhans --- merging.

Issue #243, opened 2015-02-25
Where are sequences stored in the graph?


ekg:
Reading [common.avdl](https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/common.avdl#L8), I'm curious where the association between actual DNA sequences and segments is made. Why should it not be possible to directly access the sequences associated with a particular segment via the API?

awz:
It's actually a bit more confusing than this.   Because it is possible to
ask questions about actual sequences via the Beacon API.     It would be
nice if we could keep everything consistent without bloating the simplest
possible Beacon implementation.

On Wed, Feb 25, 2015 at 9:57 AM, Erik Garrison <notifications@github.com>
wrote:

> Reading common.avdl
> <https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/common.avdl#L8>,
> I'm curious where the association between actual DNA sequences and segments
> is made. Why should it not be possible to directly access the sequences
> associated with a particular segment via the API?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ga4gh/schemas/issues/243>.
>



-- 
Alexander (Sasha) Wait Zaranek, PhD.  <sasha@curoverse.com>
Director Informatics, Personal Genome Project, Harvard Medical School
Chief Scientist, Curoverse Inc.  Founder http://arvados.org

Get paid to write free software: https://curoverse.com/about/jobs/

ekg:
I meant to cancel a comment rather than close/reopen.

@awz So I'm not wrong in wondering how I could examine the internal structures of the database backing the graph via the API? Does the system provide any concept of nodes and edges with identifiers?

awz:
@ekg - I except it is possible (maybe?), but for the time being the "how" seems to be extremely opaque.    It would be great to have some documentation / code examples for this.  Even if it is possible, this is so important that we might want to change things around so the code is much simpler/obvious.

ekg:
@awz It's expedient to be able to say "there is a node with id=N, get me that node and its context."  In vg this is a very core concept: `vg find -n <id> -c <context> x.vg >node_context.vg`. This matters a lot when working with the reference system and I'm having some trouble getting my head around how tools are supposed to operate, or at very least people are supposed to be able to debug what's going on with the backend of the API, if it isn't available.

pgrosu:
Anything is possible.  Here a few standard algorithms (BFS/DFS) for doing this:

[...]

Issue #131, opened 2014-08-24
Testing the schema with data on Mesa at Google


pgrosu:
Hi Cassie, David, et al,

I just read the paper regarding Google's Mesa from the following links:

http://research.google.com/pubs/pub42851.html

http://research.google.com/pubs/archive/42851.pdf

I noticed it has several advantages such as online schema changes as well as query by function on sets of values, which can perform in near real-time. Another advantage is that it has petascale data-warehousing with ACID properties for transactions.  The function on sets can be especially useful for the many-to-many relationships we have in our schema.

This seems to have some advantages over Megastore, Spanner, and F1 that we can try to leverage.

I was wondering if we can test the schema with data on a development area of Mesa.

Thank you,
Paul

PR #250, opened 2015-03-06
Side graphs: `Sequence`s and `Join`s

adamnovak:
Here's a draft of what our graph model would look like if instead of the "insert graphs" we are using now (where sequences come with associated end joins), we used a "side graph". In a side graph, we split out `Sequence`s (which are just strings with IDs) and `Join`s (which bind two sides of two bases in two sequences together). Both kinds of objects, in this world, can occur in `ReferenceSet`s and `VariantSet`s.

I've made a bunch of API changes to see how that would actually work: creating a `Sequence` object, to start with. @richarddurbin made `Join`s, and then I changed the methods for searching to give us nice clean `searchSequences()` and `searchJoins()` methods to get them.

@richarddurbin: is this what you were going for?

@lh3: have I captured side graphs correctly?

All: do we actually like side graphs better than insert graphs?

I don't expect this to be merged for a while, but I think we need a concrete proposal to help anchor the discussion in #243 and its ensuing massive e-mail thread.

### DIFF
--- a/src/main/resources/avro/common.avdl
+++ b/src/main/resources/avro/common.avdl
@@ -5,24 +5,28 @@ not have any methods; it is merely a library of types.
 
 ##Sequences, Segments, and References
 
-A `Sequence` is a component in a sequence graph, optionally joined at its ends
-to other `Sequence`s.
+A `Sequence` is a component in a sequence graph, optionally joined to other 
+`Sequence`s via `Join`s. It has to have length at least 1, but does not contain 
+base data itself - the base data is obtained with the `getSequenceBases()` method 
+call.
 
-A `Segment` is a range on a `Sequence`, possibly including the joins at the
-`Sequence`'s ends. It does not include base data. When you ask about the novel
+A `Segment` is a range on a `Sequence`. When you ask about the novel
 `Sequence`s that come in a `VariantSet`, you will get back `Segment`s covering
-them. A `Path` through a sequence graph is a list of `Segment`s.
+them. A `Path` through a sequence graph is a list of `Segment`s, with the
+requirement that adjacent segments are joined by `Join`s.
 
-A `Reference` (defined in references.avdl) is an object that marks a `Sequence`
-as part of a `ReferenceSet`'s sequence graph. It contains a `Segment` covering
-the `Sequence` it belongs to.
+A `Reference` (defined in references.avdl) is part of a `ReferenceSet` which
+represents the reference sequence graph. It contains a `Segment` identifying
+the underlying `Sequence`.  The reason we do this rather than directly containing
+the `Sequence` is to allow generation of local `Reference`s and `ReferenceSet`s 
+from part of a larger one, e.g. the MHC region of chromosome 6.
 
 ##Sequence Graph Concepts
 
 Many of the types defined here are related to defining sequence graphs. Sequence
 graph concepts can be illustrated on a piece of double-stranded DNA with two
 nicks in it, one on the top strand and one on the bottom strand. Here is an
-example where the top strand is GGTGGNG.
+example where the top strand is GGTGGAG.
 
 ```
             | <- top strand nick at position (2,+)
@@ -63,18 +67,15 @@ position (2,+) is the exposed phosphate group of the top strand nick in the
 diagram above, and the position (3,-) is the same thing in  the bottom strand
 nick.
 
-Each nonempty `Sequence` has a "start", which is the left side of the first
-basepair, and an "end", which is the right side of the last basepair. Each
-`Sequence` has associated, possibly null, "start join" and "end join"
-adjacencies, which indicate sides from other `Sequence`s that have been observed
-participating in phosphodiester bonds with the start and end, respectively, of
-this `Sequence`. The `Sequence` carrying the join is the "child", while the
-`Sequence` on which the join is positioned is the "parent"; parent-child
-relationships are constrained to be acyclic. If a `Sequence` with a start join
-and end join contains no basepairs, it represents a phosphodiester bond directly
-between the start join and end join sides.
+To make the sequence graph we allow `Join`s between `Sequence`s.  A `Join` can
+be thought of as an unordered pair of (oriented) positions.  The orientation 
+determines which side of a base to join to, so which direction in each sequence
+a path will continue along when traversing the `Join`.  A `Join` can connect
+two locations on the same sequence, resulting in a deletion, duplication or
+hairpin.
 
 We use ordered lists of `Segment`s to describe paths in sequence graphs.
+Adjacent `Segment`s in the list must be connected by `Join`s.
 
 A `Segment` consists of a start position (including orientation) on a
 `Sequence`, and a distance along that `Sequence` (implying an end position). The
@@ -111,10 +112,10 @@ inserted at the first break, so the result would look like this:
 We would like to take the original sequence graph given above, and extend it
 with additional `Sequence`s to also describe this new observed variation.
 
-Since the added C/G basepair is novel DNA, an additional reference `Sequence`
+Since the added C/G basepair is novel DNA, an additional `Sequence`
 must be defined to carry it. Let us call the existing GGTGGNG reference
-`Sequence` ref" and the new reference `Sequence` "nov1". Thus, the new
-reference `Sequence` "nov1" looks like
+`Sequence` ref" and the new `Sequence` "nov1". Thus, the new
+`Sequence` "nov1" looks like
 
 ```
     "ref": "nov1":  "ref":
@@ -135,13 +136,17 @@ The subgraph of our new `Sequence` graph involving "nov1" can be described as:
 
 ```
 Sequence with top strand C and ID "nov1":
-startJoin = (1,-), "ref"
-start = (0,+)
 length = 1
-end = (0,-)
-endJoin = (3, -), "ref"
+
+Join at start of nov1 
+ends = [{(1,-), "ref"}, {(0,+), "nov1"}]
+
+Join at end of nov1
+ends = [{(0,-), "nov1"}, {(3,+), "ref"}]
 
 ```
+
+TODO EXERCISE FOR ADAM
 
 To describe the other new phosphodiester bonds that need to be added to the
 graph, between (2,+) and (4,+) in "ref", we need to create a new reference
@@ -301,39 +306,27 @@ protocol Common {
 Indicates the DNA strand associate for some data item.
 * `POS_STRAND`:  The postive (+) strand.
 * `NEG_STRAND`: The negative (-) strand.
-* `NO_STRAND`: Strand-independent data or data where the strand can not be determined.
+Richard and Adam removed NO_STRAND.  Instead have a union with null if you want to support unstranded annotations.
 */
 enum Strand {
   POS_STRAND,
-  NEG_STRAND,
-  NO_STRAND
+  NEG_STRAND
 }
-
 
 /**
 Represents a join of one end of a "child" `Sequence` onto a certain side of a
 certain base in a "parent" `Sequence`.
 */
-record EndJoin {
+record Join {
   /**
-  The ID of the parent `Sequence` onto which the `EndJoin` joins. This may be a
-  `Reference` `Sequence`, or a novel `Sequence` associated with a `VariantSet`.
+  A join is simply a pair of Position objects, which we call the sides.
+  The are logically unordered (i.e. swapping makes no difference), but
+  we require a rank on the Sequences, and so implicitly on the positions,
+  so to avoid ambiguity we require that the position for side1 is less
+  than that for side2.
   */
-  string sequenceId;
-
-  /**
-  The 0-based offset from the start of the forward strand for that `Sequence` to
-  the base to which the join is made. Genomic positions are non-negative
-  integers less than the `Sequence` length.
-  */
-  long position;
-
-  /**
-  Strand onto which the join is made. `POS_STRAND` represents the forward
-  strand, or equivalently the left side of a base, and `NEG_STRAND` represents
-  the reverse strand, or equivalently the right side of a base.
-  */
-  Strand strand;
+  Position side1;
+  Position side2;
 }
 
 /**
@@ -345,36 +338,21 @@ Does not include any base data. The bases for a `Sequence` are available through
 the `getSequenceBases()` API call.
 */
 record Sequence {
-
   /**
   The ID of the sequence.
   */
   string id;
-
-
   /**
-  Join at the left end of this `Sequence`. The left (5') side of the leftmost
-  base in this "child" sequenced can be attached to the specified side of the
-  specified base in a "parent" `Sequence`.
-
-  Note that a `Sequence` can have between 0 and 2 parents.
+  The length of the sequence.  Must be greater than 0.
   */
-  union { null, EndJoin } startJoin = null;
-
-  /**
-  Join at the right end of this `Sequence`. The right (3') side of the rightmost
-  base in this "child" sequence can be attached to the specified side of the
-  specified base in a "parent" `Sequence`.
-
-  Note that a `Sequence` can have between 0 and 2 parents.
-  */
-  union { null, EndJoin } endJoin = null;
+  long length;
 }
 
 /**
-A `Position` is a side of a base pair in some already known sequence. A
+A `Position` is an oriented base in some already known sequence. A
 `Position` is represented by a sequence name or ID, a base number on that
-sequence (0-based), and a `Strand` to indicate the left or right side.
+sequence (0-based), and a `Strand` to indicate the forward or reverse-complement
+orientation.
 
 For example, given the sequence "GTGG", the `Position` on that sequence at
 offset 1 in the forward orientation would be the left side of the T/A base pair.
@@ -392,20 +370,24 @@ record Position {
   Does not generally include a "chr" prefix, so for example "X" would be used
   for the X chromosome.
 
-  If `sequenceId` is null, this must not be null.
-  */
-  union { null, string } referenceName = null;
-
   /**
   The ID of the sequence on which the `Position` is located. This may be a
   `Reference` sequence, or a novel piece of sequence associated with a
   `VariantSet`.
 
-  If `referenceName` is null, this must not be null.
+  We allow a null value for sequenceId to support the original (pre-graph) model.
 
   If the server supports the "graph" mode, this must not be null.
   */
   union { null, string } sequenceId = null;
+
+  /**
+  The name of the reference sequence on which the `Position` is located.
+  This is not used in the "graph" mode.
+
+  If `sequenceId` is null, this must not be null.
+  */
+  union { null, string } referenceName = null;
 
   /**
   The 0-based offset from the start of the forward strand for that sequence.
@@ -422,9 +404,8 @@ record Position {
 }
 
 /**
-A `Segment` is a range on a `Sequence`. It does not include the `Sequence`'s end
-joins, or any base data. (The bases for a `Sequence` are available through the
-`getSequenceBases()` API call.)
+A `Segment` is a range on a `Sequence`. It does not include any base data. 
+(The bases for a `Sequence` are available through the `getSequenceBases()` API call.)
 
 In the sequence "GTGG", the `Segment` starting at index 1 on the forward strand
 with length 2 is the "TG" on the forward strand. The length-2 `Segment` starting
@@ -467,20 +448,6 @@ record Path {
   the other.
   */
   array<Segment> segments = [];
-
-  /**
-  If the first `Segment` in the `Path` is joined onto another `Sequence` at its
-  left end, that join can be stored here. That join is then
-  considered to be included in the `Path`.
-  */
-  union { null, EndJoin } startJoin = null;
-
-  /**
-  If the last `Segment` in the `Path` is joined onto another `Sequence` at its
-  right end, that join can be stored here. That join is then
-  considered to be included in the `Path`.
-  */
-  union { null, EndJoin } endJoin = null;
 }
 
 /**
--- a/src/main/resources/avro/common.avdl
+++ b/src/main/resources/avro/common.avdl
@@ -5,10 +5,10 @@ not have any methods; it is merely a library of types.
 
 ##Sequences, Segments, and References
 
-A `Sequence` is a component in a sequence graph, optionally joined to other 
-`Sequence`s via `Join`s. It has to have length at least 1, but does not contain 
-base data itself - the base data is obtained with the `getSequenceBases()` method 
-call.
+A `Sequence` is a component in a sequence graph, optionally joined to other
+`Sequence`s via `Join`s. It has to have length at least 1, but does not contain
+base data itself - the base data is obtained with the `getSequenceBases()`
+method call.
 
 A `Segment` is a range on a `Sequence`. When you ask about the novel
 `Sequence`s that come in a `VariantSet`, you will get back `Segment`s covering
@@ -17,8 +17,8 @@ requirement that adjacent segments are joined by `Join`s.
 
 A `Reference` (defined in references.avdl) is part of a `ReferenceSet` which
 represents the reference sequence graph. It contains a `Segment` identifying
-the underlying `Sequence`.  The reason we do this rather than directly containing
-the `Sequence` is to allow generation of local `Reference`s and `ReferenceSet`s 
+the underlying `Sequence`. The reason we do this rather than directly containing
+the `Sequence` is to allow generation of local `Reference`s and `ReferenceSet`s
 from part of a larger one, e.g. the MHC region of chromosome 6.
 
 ##Sequence Graph Concepts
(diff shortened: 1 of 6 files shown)


src/main/resources/avro/common.avdl
```diff
@@ -5,44 +5,45 @@ not have any methods; it is merely a library of types.
...
-                  | <- bottom strand nick at position (3,-)
-
-    +0- +1- +2- +3- +4- +5- +6- <- coordinates
-```
-
-A sequence is a piece of double-stranded DNA composed of a series of DNA
-basepairs. In the default forward orientation a sequence is specified by the DNA
-letters of its top strand, e.g. in this case GGTGGNG, where N indicates an
+example where the top strand is GGTGGAG.
+
```
macieksmuga: Example below still uses GGTGGNG. Do you want to change N->A throughout the rest of the example?

src/main/resources/avro/common.avdl
```diff
@@ -387,19 +300,59 @@ record Position {
 }
 
 /**
-A `Segment` is a range on a sequence, possibly including the joins at the
-sequence's ends. It does not include base data. (The bases for a sequence are
-available through the `getSequenceBases()` API call.)
+A `Join` is simply a pair of Position objects, which we call the sides. The are
+logically unordered (i.e. swapping makes no difference), but we require a rank
+on the Sequences, and so implicitly on the positions, so to avoid ambiguity we
+require that the position for side1 is less than that for side2.
```
macieksmuga: possibly include this in documentation on `Position`, but from what it says here, I'm guessing that it's something like:
```
{(p1, o1), seq1} < {(p2, o2), seq2}  <=>  (seq 1 < seq2) or (seq1 = seq2. and p1 < p2)
```

src/main/resources/avro/common.avdl
```diff
@@ -5,44 +5,45 @@ not have any methods; it is merely a library of types.
...
-A `Segment` is a range on a sequence, possibly including the joins at the
-sequence's ends. It does not include base data. When you ask about the novel
-sequences that come in a `VariantSet`, you will get back `Segment`s covering
-them. A `Path` through a sequence graph is a list of `Segment`s.
-
-A `Reference` (defined in references.avdl) is an object that marks a sequence as
-part of a `ReferenceSet`'s sequence graph. It contains a `Segment` covering the
-sequence it belongs to.
+A `Sequence` is a component in a sequence graph, optionally joined to other
+`Sequence`s via `Join`s. It has to have length at least 1, but does not contain
```
macieksmuga: Previously `Segment`s were used to represent deletions as well. In this version of the graph model, a deletion, transposition, or other base-less change is effected through use of pure `Join`s, with no `Segment`s involved. Thus the requirement for length > 0.

src/main/resources/avro/common.avdl
```diff
@@ -412,36 +365,21 @@ record Segment {
...
-  union { null, Position } startJoin = null;
-  union { null, Position } endJoin = null;
 }
 
 /**
 A `Path` is an ordered list of `Segment`s. In general any contiguous path
 through a sequence graph, with no novel adjacencies, can be represented by a
 `Path`.
+
+Can optionally include the joins at either end of the path.
```
macieksmuga: Wait, how can a path include joins (now `Join`s) if it's just an array of `Segment`s?

src/main/resources/avro/references.avdl
```diff
@@ -31,65 +31,43 @@ field. New client and server implementations should prefer the "graph" mode.
 record Reference {
 
   /**
-  The `Segment` describing this `Reference`'s sequence in the `ReferenceSet`'s
+  The `Segment` describing this `Reference`'s `Sequence` in the `ReferenceSet`'s
   sequence graph.
 
-  The `Segment` will generally cover the entirety of the sequence, but may be
-  shorter if only a subrange of a sequence is to be included in the
+  The `Segment` will generally cover the entirety of the `Sequence`, but may be
```
macieksmuga: This is probably just me being dense but, is there a good reason besides this for continuing to hold on to the `Segment` as an entity, now that `Sequence` is one? And even in this case, couldn't a `Reference` simply include a `Sequence` ID together with `start` and `length` fields to denote the relevant subrange?

macieksmuga: Amended after discussing with Adam: I'm suggesting removing `Segment` from `Reference` altogether, replacing it with an obligatory non-null `Sequence` ID combined with position and length information.

src/main/resources/avro/common.avdl
adamnovak: Yeah, it should probably be N. I will fix that.

src/main/resources/avro/common.avdl
adamnovak: OK, I'll specify `Position` ordering in the `Position` doc comment.

src/main/resources/avro/common.avdl
adamnovak: Yep!

src/main/resources/avro/common.avdl
adamnovak: It's an array of `Segment`s together with these optional end joins. It's to handle weird cases where you want to talk about a `Path` that includes only a `Join`, or a `Path` where it matters whether a join at one end is present or not.

This is sort of a fuzzy bit of the proposed changes; are there better ways to handle those cases, or good justifications for why we don't need them?

src/main/resources/avro/references.avdl
adamnovak: OK, I'm going to pull `Segment` out of `Reference` so it only lives in `Path`. You might now have to define your `Reference`s in the forward orientation on your `Sequence`s.

src/main/resources/avro/common.avdl
```diff
@@ -410,26 +365,11 @@ record Segment {
...
-  Joins may occur on the outer sides of the terminal bases in a sequence: the
-  left side of the base at index 0, and the right side of the base with maximum
-  index. These are the "terminal sides" of the sequence. `startJoin` is the join
-  on the side indicated by `start`, and may only be set if that side is a
-  terminal side. Similarly, `endJoin` is the join on the opposite side of the
-  piece of sequence selected by the segment, and may only be set if that side is
-  a terminal side. The value of `startJoin` or `endJoin`, if set, is the side to
-  which the corresponding side of this `Sequence` is connected.
+  A `Segment` may have zero length (for example, when it is being used to
+  specify a `Path` consisting onyl of a `Join`.
```
jeromekelleher: typo: onyl

src/main/resources/avro/common.avdl
```diff
@@ -387,19 +302,59 @@ record Position {
...
 /**
-A `Segment` is a range on a sequence, possibly including the joins at the
-sequence's ends. It does not include base data. (The bases for a sequence are
-available through the `getSequenceBases()` API call.)
+A `Join` is simply a pair of Position objects, which we call the sides. The are
+logically unordered (i.e. swapping makes no difference), but we require a rank
+on the Sequences, and so implicitly on the positions, so to avoid ambiguity we
+require that the position for side1 is less than that for side2.
+*/
+record Join {
```
jeromekelleher: Do we need an explicit `Join` type? If it's just an ordered pair of positions, then wouldn't an `array<Position>` would be more natural (and give a more elegant programming model)? We don't actually seem to use the `Join` type anywhere except for the `SearchJoinsRequest` which seems odd.

src/main/resources/avro/variants.avdl
```diff
@@ -360,24 +360,23 @@ single objects that are present in some copy number.
...
 reference and alternate alleles of `Variant`s.
 
-Note that `Path`s cannot follow adjacencies not represented in the augmented
+Note that `Path`s cannot follow `Join`s not represented in the augmented
 sequence graph. For example, if  an `Allele` that spans a novel deletion is
-required, a new empty-string sequence should exist to describe that deletion
-adjacency, with a `Segment` describing that sequence available through
-`searchVariantSetSequences()` or `getVariantSetSequence()`.
+required, a new `Join` should exist to describe that deletion, available through
+`searchVariantSetJoins()`.
```
jeromekelleher: Does `searchVariantSetJoins` exist?

src/main/resources/avro/referencemethods.avdl
```diff
@@ -325,4 +396,99 @@ GetSequenceBasesResponse getSequenceBases(
     /** Additional request parameters to restrict the query. */
     GetSequenceBasesRequest request) throws GAException;
 
+/****************  /joins/search  *******************/
+/**
+This request maps to the body of `POST /joins/search`
+as JSON.
+*/
+record SearchJoinsRequest {
```
jeromekelleher: This method seems pretty heavily overloaded and complicated. I think it would be better to have `SearchReferenceJoins` and `SearchVariantSetJoins`, so the semantics aren't so complicated.

src/main/resources/avro/referencemethods.avdl
```diff
@@ -325,4 +396,99 @@ GetSequenceBasesResponse getSequenceBases(
...
+  // range.
+
+  /**
+  If nonempty, return only `Join`s with at least one `Position` on one of these
+  sequences.
+  */
+  array<string> sequenceIds = [];
+  
+  /**
+  Must not contain more entries than `sequenceIds`.
```
jeromekelleher: This query model seems very complicated and vague. We can query on any or all of `sequenceIds`, `starts`, `lengths` or `strands`? Are we returning only the results which match all these filters? What difference does it make if the length of the `starts` for example is greater than `sequenceIds`?

What does client code want to do with this?

pgrosu: Jerome, yes it is intended to be flexible, since it's just a definition as Cassie (@cassiedoll) mentioned a while ago in the following post https://github.com/ga4gh/schemas/issues/131#issuecomment-53291001.  So it's up to the implementation to decide the most reasonable, elegant and efficient algorithms to provide the functionality, which means you have almost complete freedom :)

~p

src/main/resources/avro/common.avdl
```diff
@@ -53,293 +54,201 @@ forward orientation. The right side is indicated using the "-" orientation. For
...
 * `NEG_STRAND`: The negative (-) strand.
-* `NO_STRAND`: Strand-independent data or data where the strand can not be determined.
+* `POS_STRAND`:  The postive (+) strand.
+Richard and Adam removed NO_STRAND. Instead have a union with null if you want
+to support unstranded annotations.
 */
 enum Strand {
-  POS_STRAND,
   NEG_STRAND,
-  NO_STRAND
```
jeromekelleher: Shouldn't we remove this comment and carry the change through on to `Position.strand` (i.e., make it a union with null)?

src/main/resources/avro/referencemethods.avdl
```diff
@@ -261,17 +260,89 @@ ListReferenceBasesResponse getReferenceBases(
...
 
-/****************  /sequence/{id}  *******************/
+/******************  /sequences/search  *********************/
+/**
+This request maps to the body of `POST /sequences/search` as JSON.
+*/
+record SearchSequencesRequest {
+  /**
+  If not null, return only `Sequence`s that appear in the indicated
+  `ReferenceSet`, or any included `ReferenceSet`s.
```
jeromekelleher: I don't understand the last clause: "... or any included `ReferenceSet`s.". What do you mean here?

src/main/resources/avro/referencemethods.avdl
```diff
@@ -261,17 +260,89 @@ ListReferenceBasesResponse getReferenceBases(
...
+  */
+  union { null, string } referenceSetId = null;
+  
+  /**
+  If not null, return only `Sequence`s that are novel in the indicated
+  `VariantSet`.
+  */
```
jeromekelleher: Novel between the VariantSet and what? What's the use case for this argument?

src/main/resources/avro/references.avdl
```diff
@@ -146,10 +129,23 @@ record ReferenceSet {
...
+  array<string> includedReferenceSets = null;
+
+  /**
+  Order-independent MD5 checksum which identifies this `ReferenceSet`.
+
+  To compute this checksum, make a list of `Reference.md5checksum` for all
+  `Reference`s in this set, including any in any included set. Add to that list
+  a lower-case MD5 checksum for each `Join` in the `ReferenceSet` or any
+  included set, which is calculated by hashing the `md5checksum`, position as a
+  decomal string, and strand as "+" or "-" for the two `Position`s of the `Join`
```
jeromekelleher: s/decomal/decimal

jeromekelleher:
Thanks for the updates @adamnovak. The model itself seems sensible enough to me, but I'm really struggling to understand how the API would work. How would a client do something useful with this? In particular, I don't understand what the client is supposed to do with `Join` objects. I'm obviously missing something important here! It would be really useful I think to try and write down some client code to do some very straightforward calculations. Perhaps we could flesh out some of the operations that we are going to do as part of the graph comparison, and use those as test-cases for the API?

src/main/resources/avro/referencemethods.avdl
adamnovak: I think it's supposed to be an AND of ORs. You specify several acceptable values for each filter, and the results have to match all of them.

I can try to make that clearer in the documentation here, but I think that's the style we've been using throughout the search functions.

src/main/resources/avro/common.avdl
adamnovak: I think we do; on balance I think not having a Sequence was a bad idea, even when it was possible to make Segment do all the work, so I think if it's a first-class object in the model it needs to have a type, if only to avoid confusing people.

src/main/resources/avro/common.avdl
fnothaft: +1 @jeromekelleher

src/main/resources/avro/common.avdl
fnothaft: I don't know if we _strictly_ need a `Join` type, but I believe that having the `Join` type instead of overloading an array makes the data model cleaner.

benedictpaten: On Wed, Mar 18, 2015 at 9:37 AM, Frank Austin Nothaft <
notifications@github.com> wrote:

> In src/main/resources/avro/common.avdl
> <https://github.com/ga4gh/schemas/pull/250#discussion_r26680976>:
>
> > @@ -387,19 +302,59 @@ record Position {
> >  }
> >
> >  /**
> > -A `Segment` is a range on a sequence, possibly including the joins at the
> > -sequence's ends. It does not include base data. (The bases for a sequence are
> > -available through the `getSequenceBases()` API call.)
> > +A `Join` is simply a pair of Position objects, which we call the sides. The are
> > +logically unordered (i.e. swapping makes no difference), but we require a rank
> > +on the Sequences, and so implicitly on the positions, so to avoid ambiguity we
> > +require that the position for side1 is less than that for side2.
> > +*/
> > +record Join {
>
> I don't know if we *strictly* need a Join type, but I believe that having
> the Join type instead of overloading an array makes the data model
> cleaner.
>

+1 @franknothaft

src/main/resources/avro/referencemethods.avdl
fnothaft: Agree with @jeromekelleher

src/main/resources/avro/referencemethods.avdl
fnothaft: I think this would mean "novel sequences that have been added to the graph when creating variant set X"?

src/main/resources/avro/referencemethods.avdl
fnothaft: I agree that this model is vague; I think @adamnovak's comment confirms this ;). However, as it is consistent with _other_ GA4GH APIs, I don't think we should hold up the PR on this. As an aside, there's no reason that the query definition needs to be unstructured in order to allow implementations to implement a flexible execution model. E.g., databases rewrite user provided queries in order to improve query performance. I think it would make sense to refactor the way that we express queries, but that's a longer debate and not really germane to this discussion.

fnothaft:
+1

As an aside, I agree with @jeromekelleher:

> How would a client do something useful with this?

But, I think we're in a bit of a chicken/egg situation (e.g., we can't reason about how to perform common query X, unless we've got a data model). I think this is a data model that we can work with, and we'll figure out what the query patterns are by using the data model.

benedictpaten:
+1


Critical to "how do we do useful things" question is adding in variant
search queries, like "give me all the variants in the graph related to chr1
positions 1 to 100?". The rank on positions/joins makes it possible to
define this concretely.  A related problem is agreeing on exactly what we
refer to as an allele in the graph. There has been a good deal of
conversation on that subject, but we need some pull requests (building on
this one), to make it concrete.

src/main/resources/avro/common.avdl
diekhans: Frank Austin Nothaft <notifications@github.com> writes:
> In src/main/resources/avro/common.avdl:
> 
> >    NEG_STRAND,
> > -  NO_STRAND
> 
> +1 @jeromekelleher

-1, NO_STRAND  is required for genome annotations.

Although it would be perfectly reasonable to have two different
stand representation, I would +1 that if it's *really* need and we
create two enums where POS and NEG have the same value.

But it makes it a lot easier if there isn't.

src/main/resources/avro/common.avdl
adamnovak: OK, I am fixing this.

jeromekelleher:
I'm basically +1 on the data model here, but I've still got issues with the API methods. Perhaps we could break these into separate PRs, so we're not holding up implementation of the data model?

src/main/resources/avro/common.avdl
macieksmuga: @diekhans I don't understand why a `NO_STRAND` enum value would be preferable to making the type of the `strand` field in `Position` to be `union { null, Strand }`. Wouldn't the latter be a more idiomatic way of expressing the same thing?

Related question: does a `NO_STRAND` value as used for genome annotations mean that we don't know the strand in some particular case, or that we don't care (i.e. care only about the position, not side)?

src/main/resources/avro/variants.avdl
adamnovak: Nope, this is supposed to be `searchJoins()`.

macieksmuga:
@jeromekelleher Interesting suggestion on splitting the PR. I know it would help me a lot to have the basic graph model stabilized, even as the discussion on API methods and Alleles carries on. @adamnovak what do you think?

fnothaft:
I think @jeromekelleher's suggestion is a great suggestion.

src/main/resources/avro/referencemethods.avdl
adamnovak: OK, I'm adding a comment to the `SearchJoinsRequest` object explicitly stating that results have to match all of the filters. In general I think (as noted in the TODO) that we need a more general way to do queries, but I agree with @fnothaft that it ought to be its own task (and presumably its own pull request).

I've made an issue (#260) to deal with the general case.

src/main/resources/avro/referencemethods.avdl
adamnovak: See https://github.com/ga4gh/schemas/pull/250/files#diff-ffb7a8b69c62fd7062a6483e1e95d905R132 which @richarddurbin added.

Basically `ReferenceSet`s are also now supposed to be composeable: you can declare a new `ReferenceSet` as an old `ReferenceSet` plus some new `Sequence`s and `Join`s.

src/main/resources/avro/referencemethods.avdl
adamnovak: Yes, @fnothaft has the right idea. Clearly this is unclear; I will change it to "`Sequence`s that are part of the indicated `VariantSet`".

src/main/resources/avro/references.avdl
adamnovak: OK, I am fixing this.

src/main/resources/avro/referencemethods.avdl
macieksmuga: Shouldn't this composability be made explicit in the definition of `ReferenceSet`? This was news to me.

src/main/resources/avro/references.avdl
```diff
@@ -125,13 +100,21 @@ record Reference {
...
 reference assembly, such as `GRCh38`. A `ReferenceSet` defines a common
 coordinate space for comparing reference-aligned experimental data.
+
+In classic mode, this is just a set of sequences.  In graph mode, there
+are both sequences and joins, but we only access the joins via method
+getReferenceJoins().
 */
+
 record ReferenceSet {
   /** The reference set ID. Unique in the repository. */
```
macieksmuga: If `ReferenceSet`s are now meant to be composeable, should this be explicitly added somewhere in the definition of the object here?

adamnovak:
I do see how it would be good to hash out things like `searchJoins` versus `searchVariantSetJoins` and so on independently from the data model change, but if we accept the data model change without either of those methods, there would be no way to retrieve connectivity information until that discussion was finished, which is definitely a regression.

Would people be OK with that?

fnothaft:
> Would people be OK with that?

I am OK with that.

src/main/resources/avro/references.avdl
adamnovak: It is in the definition (see `includedReferenceSets`), but it should be in the doc comment.

macieksmuga: @adamnovak oops... was accidentally searching on master, not your PR version of the file. Sorry!

src/main/resources/avro/referencemethods.avdl
pgrosu: Adam, the way we approached the definitions so far are more like proofs in mathematics - which provide more flexibility - rather than an API implementation which creates bounds and stringency.  Implementation-wise we can also make a specific program's implemented-API more stringent - which might be due to resource-constraints - but the schema definitions should be as flexible as possible to match the intended biological questions (especially the large ones).  I feel the current flow from schema_definition -> implemented_program is easier to go with in the long run, and will allow for extensibility when we want to perform large queries across many repositories such as with Beacon or Matchmaker.  There are always distributed algorithms that can adapt to the definition, so I would lean more to keep the definitions as pure as possible.

~p

src/main/resources/avro/common.avdl
adamnovak: OK, I'm dropping the comment and making `Position.strand` nullable, but that means that some places where we have a `Position` (`Segment`s and `Join`s) need to require explicitly that their `Position`s have non-null `Strand`s.

Did I miss any of those instances?

adamnovak:
OK, in response to @diekhans's comments about needing a good way to specify a Position without a strand, I now have a `Position` with no strand field, and a `Side` which is a position and a strand (which is required to be + or -, and cannot be null). If you don't know if you will have a strand or not, you would use a `uinon { Position, Side }`.

@diekhans, is that good for your use case? I didn't find anything in the metadata API to switch over to the new type.

src/main/resources/avro/common.avdl
richarddurbin: Sorry, when I was editing I half edited N to A, then stopped.  I think it would be good to remove N from
the examples, although to say somewhere that N is a legitimate character in a sequence.  There are a few
Ns, but they are not typical.

Richard

> On 17 Mar 2015, at 04:40, Maciek Smuga-Otto <notifications@github.com> wrote:
> 
> In src/main/resources/avro/common.avdl <https://github.com/ga4gh/schemas/pull/250#discussion_r26549104>:
> 
> > -```
> > -            | <- top strand nick at position (2,+)
> > -5' -------  -------------------  3'
> > -     G   G   T   G   G   N   G
> > -     C   C   A   C   C   N   C
> > -3' ----------------  ----------  5'
> > -                  | <- bottom strand nick at position (3,-)
> > -
> > -    +0- +1- +2- +3- +4- +5- +6- <- coordinates
> > -```
> > -
> > -A sequence is a piece of double-stranded DNA composed of a series of DNA
> > -basepairs. In the default forward orientation a sequence is specified by the DNA
> > -letters of its top strand, e.g. in this case GGTGGNG, where N indicates an
> > +example where the top strand is GGTGGAG.
> > +
> Example below still uses GGTGGNG. Do you want to change N->A throughout the rest of the example?
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/ga4gh/schemas/pull/250/files#r26549104>.
> 




-- 

jeromekelleher:
Re. splitting the PR, I think it's totally fine if we're temporarily missing some methods from our git repo.

adamnovak:
OK, here's a version with no changes to the methods files.

jeromekelleher:
Thanks @adamnovak, +1.

We should squash the commits - I guess one Adam commit, one Richard commit, one Adam commit is the most appropriate?

adamnovak:
OK I've squashed everything down (even though I have heard it is rude to rewrite other people's commits). What do you think?

jeromekelleher:
Looks good Adam - we could probably tidy up the commit comments a bit though. Stuff like "Rolling methods back" doesn't really tell future-us very much about what this commit actually does.

benedictpaten:
I'll merge this in the next 24 hours (unless Jerome feels strongly about tidying up the commit comments).

adamnovak:
OK, I've cleaned up the concatenated commit messages slightly.

jeromekelleher:
Looks good, thanks @adamnovak. Merging.

PR #252, opened 2015-03-07
renaming 'allele' to alternateBases and adding a field referenceBases

maximilianh:
Based on the discussion after mfiume's last pull request, adding a field referenceBases and renaming the allele field to alternateBases to make this API more similar to the variants API. Indels can now be specified just as in VCF and the variants API.


mfiume:
+1 for consistency with other GA4GH APIs, and based on discussions with @richarddurbin, @ekg, and @deannachurch on PR #244

Issue #248, opened 2015-03-06
Dataset role not clearly defined


jeromekelleher:
Currently, the `Dataset` type is poorly specified, and included in `reads.avdl`. We have some comments like 
```
TODO: Reads and variants both want to have datasets. Are they the same object?
```
This needs to be clarified. 

I propose:
1. We change the name to `DataSet`, so that we are consistent with `VariantSet`, `ReferenceSet`, `ReadGroupSet` and others.
2. We make the role of the `DataSet` much more explicit, and make a formal definition of the data model as a hierarchy. (This relates to #247 also, since my proposal there assumes a hierarchy.) In this hierachy, the`DataSet` is the root, and it has child `VariantSet`s and `ReadGroupSets`. Each `VariantSet` has child `Variant`s;  each `ReadGroupSet` has child `ReadGroup`s, and `ReadGroup`s have child `Read`s. We don't need to make any changes to the API to do this, we just need to be more explicit about what the model means. 

This formalisation should make reasoning about authorisation much more straightforward.

Thoughts?

jeromekelleher:
This hierarchy idea is broken somewhat by the `Call` and `CallSet` classes:

1. `CallSet` has a field `variantSetIds`, so it can belong to more than one `VariantSet`. Do we really want `CallSets` to belong to more than one `VariantSet`? It would be a lot simpler if we had one call belonging to exactly one `Variant` and one `CallSet`.
2. `SearchCallsRequest` is badly overloaded: we can search for a calls contained in a list of `callSetIds`, `variantSetIds` and `variantIds`. We should replace all of these with a single `callSetId`, since (a) a `CallSet` belongs to a `VariantSet`, so specifying the  `vaiantSetId` is redundant; and (b) obtaining the `Calls` associated with a given `Variant` should be done directly using a `SearchVariantsRequest`. We should ask what the point of a `SearchCallsRequest` is at all in this case, as the idea of getting all the `Calls` in a `CallSet` without reference to their parent `Variants` seems somewhat obscure...
3. `SearchAlleleCalls` suffers from similar problems, having four different lists of IDs that we can provide as parameters.

dcolligan:
I would also like to know what function the Dataset was supposed to serve.

richarddurbin:
It encapsulates a set of data with shared access control (most important) and typically provenance.  For example from a single deposited VCF.
There may also be namespace implications, e.g. sample names are unique in a dataset, but not necessarily between.
[...]

Issue #247, opened 2015-03-05
Query semantics for parent IDs inconsistent and problematic


jeromekelleher:
The structure of the queries for the different aspects of the data hierarchy are inconsistent, with variants, reads and references all having slightly different semantics when it comes to specifying the parent IDs that we are allowed to search over.This is unfortunate and makes implementation difficult. 

We have:

```
SearchReadGroupSetsRequest {
  /** The list of datasets to search. */
  array<string> datasetIds = [];
  // removing rest for clarity
}
SearchReadsRequest {
  /** If specified, restrict this query to reads within the given readgroups. */
  array<string> readGroupIds = [];
  // removing rest for clarity
}
SearchVariantSetsRequest {
  /**
  If specified, will restrict the query to variant sets within the
  given datasets.
  */
  array<string> datasetIds = [];
  // removing rest for clarity
}
SearchVariantSetSequencesRequest {
  /**
  If nonempty, return `Segment`s describing novel sequences in the `VariantSet`
  that are joined onto sequences with one of the specified IDs.
  */
  array<string> parentIds = [];
  // removing rest for clarity
}
SearchVariantsRequest {
  /** Required. The IDs of the variant sets to search over. */
  array<string> variantSetIds = [];
  // removing rest for clarity
}
SearchAllelesRequest {
  /** Required. The IDs of the variant sets to search over. */
  array<string> variantSetIds = [];
  // removing rest for clarity
}
SearchCallSetsRequest {
  /**
  If specified, will restrict the query to call sets within the
  given variant sets.
  */
  array<string> variantSetIds = [];
 // removing rest for clarity
}
SearchAlleleCallsRequest {
  /**
  If nonempty, only return the `AlleleCall`s in these `CallSet`s.
  */
  array<string> callSetIds = [];
  /**
  If nonempty, return only `AlleleCall`s in these `VariantSet`s.
  */
  array<string> variantSetIds = [];
  // removing rest for clarity
}
SearchReferenceSetsRequest {
  // No parent IDs required
}
SearchReferencesRequest {
  /**
  If present, return only references which belong to this reference set.
  */
  union { null, string } referenceSetId = null;
   // removing rest for clarity
}
```

I've lost count here, but I think there are at least four different conventions with different semantics. In short, it's a mess.

[...]

Issue #148, opened 2014-09-19
Auth


skeenan:
**Proposed topic for ASHG** 
So far we have been ignoring Auth. But we all know that in the real world a large fraction of. Are there parts of an auth model that should be standardised - or not - and how would we do that.

delagoya:
Closing in favor of Oauth discussions and implementation from  ga4gh/server#31
Authentication will be resource specific and not a part of the API at this point. 
If someone disagrees, please reopen the issue with a specific example as a new issue/PR

Issue #125, opened 2014-08-20
The Reference : ReferenceSet relationship


calbach:
Primarily this is directed towards @richarddurbin.

While not stated explicitly, the schema implies that a `GAReference` belongs to exactly one `GAReferenceSet`. The implication is made by `GAReference` comments referring to its parent `GAReferenceSet` to inherit values for various fields.

Would there be value in supporting a many:many relationship here? I can see it being useful if two `GAReferenceSet`s have a moderate probability of being similar; for instance if they differed only in one contig or one were a subset of another. Furthermore, a many:many relationship would enable a repository to hold a 1 `GAReference` per unique MD5 invariant, if it desired to do so. This is not feasible given the current 1:many semantics.

If #120 were accepted, the inheritance semantics of the current schema could be safely removed as getting a `GAReference` would be a lightweight metadata-only operation. The last piece I'm not quite grasping is where `isDerived` fits in.

Note that I'm not necessarily advising that the schema dictate a many:many relationship (this is a separate question on which I haven't formed an opionion), my primary desire is to shift the semantics so that it would not be disallowed.

PR #253, opened 2015-03-10
Rationalises the SearchObjectRequest semantics.

jeromekelleher:
This PR changes the semantics of all `SearchObjectRequest`s so they require the ID of a single container object, if applicable. Currently, a number of different conventions exist, making implementation and authorisation difficult. This solves issue #247 and is related to the issues in #248.

Most of the changes are straightforward, and just require the change of a field like (for example) `array<string> dataSetIds = [];` to `string dataSetId;`. There are some problematic cases:

1. `SearchCallsRequest` and `SearchAlleleCallsRequest` both had a large list of options for the containers that they could search over. As an initial strawman I've reduced this to just the parent `CallSet` to be consistent with all the other objects. Reasoning for these:
   1. Since a `CallSet` is part of a `VariantSet`, the variant set is implied by specifying `callSetId`.
   2. Finding  `Calls` for a particular `variantId` is redundant, since `Call`s are embedded in `Variant` objects. If we wish to find the calls for a given variant, we can either find them directly using `SearchVariantsRequest` or make a call to GET `/variants/<id>/`
   3. We are then left with iterating over all calls in a CallSet, which seems like a peculiar thing to do without reference to the containing `Variant` object. However, this might be updated with more meaningful semantics later.
2. `SeachAlleles` is problematic; what is the interplay between the `variantSetId` and the `sequenceId`? If `sequenceId` is required, then why do we need the `variantSetId`? Conceptually, it seems natural that an Allele belongs to a single `VariantSet`.

In terms of backwards compatibility, obviously there are some breaking changes here. These are minor, however, and easily worked around. Server-side, this is a considerable simplification, and so implementation can simply be a special case of the existing semantics. Client-side, users that are using the multi-container search semantics will need to now explicitly iterate over the container IDs that they are interested in.

### DIFF
--- a/src/main/resources/avro/readmethods.avdl
+++ b/src/main/resources/avro/readmethods.avdl
@@ -14,8 +14,10 @@ to `ReferenceSet`s containing that same `Reference`. If no reference is
 specified, all `ReadGroup`s must be aligned to the same `ReferenceSet`.
 */
 record SearchReadsRequest {
-  /** If nonempty, restrict this query to reads within the given readgroups. */
-  array<string> readGroupIds = [];
+  /**
+  The ReadGroups to search. At least one readGroupId must be specified.
+  */
+  array<string> readGroupIds;
 
   /**
   The reference to query. Leaving blank returns results from all
@@ -82,8 +84,10 @@ SearchReadsResponse searchReads(
 /******************  /readgroupsets/search  *********************/
 /** This request maps to the body of `POST /readgroupsets/search` as JSON. */
 record SearchReadGroupSetsRequest {
-  /** The list of datasets to search. */
-  array<string> datasetIds = [];
+  /**
+  The dataset to search.
+  */
+  string datasetId;
 
   /**
   Only return read group sets with this name (case-sensitive, exact match).
--- a/src/main/resources/avro/sequenceAnnotationmethods.avdl
+++ b/src/main/resources/avro/sequenceAnnotationmethods.avdl
@@ -7,7 +7,7 @@ protocol SequenceAnnotationMethods {
   import idl "sequenceAnnotations.avdl";
 
   // FIXME: wiggle methods not defined.
-  
+
   /******************  /features/search  *********************/
 
   /**
@@ -15,16 +15,16 @@ protocol SequenceAnnotationMethods {
   */
   record SearchFeaturesRequest {
     /**
-    The annotation sets to search within. Either `featureSetIds` or
-    `parentIds` must be non-empty.
+    The annotation set to search within. Either `featureSetId` or
+    `parentId` must be non-empty.
     */
-    array<string> featureSetIds = [];
+    union { null, string } featureSetId;
 
     /**
     Restricts the search to direct children of the given parent `feature`
-    IDs. Either `featureSetIds` or `parentIds` must be non-empty.
+    ID. Either `featureSetId` or `parentId` must be non-empty.
     */
-    array<string> parentIds = [];
+    union { null, string } parentId;
 
     /**
     If specified, this query matches only annotations which overlap this path.
--- a/src/main/resources/avro/variantmethods.avdl
+++ b/src/main/resources/avro/variantmethods.avdl
@@ -8,10 +8,9 @@ import idl "variants.avdl";
 /** This request maps to the body of `POST /variantsets/search` as JSON. */
 record SearchVariantSetsRequest {
   /**
-  If nonempty, will restrict the query to variant sets within the
-  given datasets.
+  The Dataset to search.
   */
-  array<string> datasetIds = [];
+  string datasetId;
 
   /**
   Specifies the maximum number of results to return in a single page.
@@ -62,11 +61,95 @@ org.ga4gh.models.VariantSet getVariantSet(
   */
   string id) throws GAException;
 
+/******************  /variantsets/{id}/sequences/search  *********************/
+/**
+This request maps to the body of `POST /variantsets/{id}/sequences/search` as
+JSON.
+*/
+record SearchVariantSetSequencesRequest {
+  /**
+  Return `Segment`s describing novel sequences in the `VariantSet`
+  that are joined onto the sequence with the specified ID.
+  */
+  string parentId;
+
+  /**
+  Specifies the maximum number of results to return in a single page.
+  If unspecified, a system default will be used.
+  */
+  union { null, int } pageSize = null;
+
+  /**
+  The continuation token, which is used to page through large result sets.
+  To get the next page of results, set this parameter to the value of
+  `nextPageToken` from the previous response.
+  */
+  union { null, string } pageToken = null;
+}
+
+/**
+This is the response from `POST /variantsets/{id}/sequences/search` expressed as
+JSON.
+*/
+record SearchVariantSetSequencesResponse {
+  /**
+  The list of `Segment`s for matching sequences from the novel sequence graph
+  for the `VariantSet`.
+  */
+  array<org.ga4gh.models.Segment> segments = [];
+
+  /**
+  The continuation token, which is used to page through large result sets.
+  Provide this value in a subsequent request to return the next page of
+  results. This field will be empty if there aren't any additional results.
+  */
+  union { null, string } nextPageToken = null;
+}
+
+/**
+Gets `Segment`s covering the sequences that match the search criteria.
+
+`POST /variantsets/{id}/sequences/search` must accept a JSON version of
+`SearchVariantSetSequencesRequest` as the post body and will return a JSON
+version of `SearchVariantSetSequencesResponse`.
+*/
+SearchVariantSetSequencesResponse searchVariantSetSequences(
+  /**
+  The ID of the `VariantSet` which provides the novel sequences.
+  */
+  string variantSetId,
+  /**
+  This request maps to the body of `POST variantsets/{id}/sequences/search` as
+  JSON.
+  */
+  SearchVariantSetSequencesRequest request) throws GAException;
+
+/*****************  /variantsets/{id}/sequences/{id}  *******************/
+/**
+Gets a `Segment` describing a novel sequence in a `VariantSet`, by sequence
+ID.
+
+`GET /variantsets/{id}/sequences/{id}` will return a JSON version of a `Segment`
+covering the novel sequence with the given ID, and describing where it is
+joined onto its parent sequences.
+*/
+org.ga4gh.models.Segment getVariantSetSequence(
+  /**
+  The ID of the `VariantSet` which provides the novel sequence.
+  */
+  string variantSetId,
+  /**
+  The ID of the novel sequence to get a covering `Segment` for.
+  */
+  string sequenceId) throws GAException;
+
 /******************  /variants/search  *********************/
 /** This request maps to the body of `POST /variants/search` as JSON. */
 record SearchVariantsRequest {
-  /** Required. The IDs of the variant sets to search over. */
-  array<string> variantSetIds = [];
+  /**
+  The VariantSet to search.
+  */
+  string variantSetId;
 
   /** Only return variants which have exactly this name (case-sensitive, exact match). */
   union { null, string } name = null;
@@ -163,8 +246,10 @@ org.ga4gh.models.Variant getVariant(
 /******************  /alleles/search  *********************/
 /** This request maps to the body of `POST /alleles/search` as JSON. */
 record SearchAllelesRequest {
-  /** Required. The IDs of the variant sets to search over. */
-  array<string> variantSetIds = [];
+  /**
+  The VariantSet to search.
+  */
+  string variantSetId;
 
   /**
   Required. Only return `Allele`s on the sequence with this ID.
@@ -243,10 +328,9 @@ org.ga4gh.models.Allele getAllele(
 /** This request maps to the body of `POST /callsets/search` as JSON. */
 record SearchCallSetsRequest {
   /**
-  If nonempty, will restrict the query to call sets within the
-  given variant sets.
+  The VariantSet to search.
   */
-  array<string> variantSetIds = [];
+  string variantSetId;
 
   /**
   Only return call sets with this name (case-sensitive, exact match).
@@ -300,127 +384,4 @@ org.ga4gh.models.CallSet getCallSet(
   The ID of the `CallSet`.
   */
   string id) throws GAException;
-
-/****************** /calls/search *********************/
-/** This request maps to the body of `POST /calls/search` as JSON. */
-record SearchCallsRequest {
-
-  /**
-  If nonempty, only return the `Call`s in these `CallSet`s.
-  */
-  array<string> callSetIds = [];
-
-  /**
-  If nonempty, return only `Call`s in these `VariantSet`s.
-  */
-  array<string> variantSetIds = [];
-
-  /**
-  If nonempty, return only `Call`s associated with one of these `Variant`s.
-  */
-  array<string> variantIds = [];
-
-  /**
-  Specifies the maximum number of results to return in a single page.
-  If unspecified, a system default will be used.
-  */
-  union { null, int } pageSize = null;
-
-  /**
-  The continuation token, which is used to page through large result sets.
-  To get the next page of results, set this parameter to the value of
-  `nextPageToken` from the previous response.
-  */
-  union { null, string } pageToken = null;
-}
-
-/**
-This is the response from `POST /calls/search` expressed as JSON.
-*/
-record SearchCallsResponse {
-  /** The list of matching calls. */
-  array<org.ga4gh.models.Call> calls = [];
-
-  /**
-  The continuation token, which is used to page through large result sets.
-  Provide this value in a subsequent request to return the next page of
-  results. This field will be empty if there aren't any additional results.
-  */
-  union { null, string } nextPageToken = null;
-}
-
-/**
-Gets a list of `Call` matching the search criteria.
-`POST /calls/search` must accept a JSON version of
-`SearchCallsRequest` as the post body and will return a JSON version of
-`SearchCallsResponse`.
-*/
-SearchCallsResponse searchCalls(
-  /** This request maps to the body of `POST /calls/search` as JSON. */
-  SearchCallsRequest request) throws GAException;
-
-/****************** /allelecalls/search *********************/
-/** This request maps to the body of `POST /allelecalls/search` as JSON. */
-record SearchAlleleCallsRequest {
-
-  /**
-  If nonempty, only return the `AlleleCall`s in these `CallSet`s.
-  */
-  array<string> callSetIds = [];
-
-  /**
-  If nonempty, return only `AlleleCall`s in these `VariantSet`s.
-  */
-  array<string> variantSetIds = [];
-
-  /**
-  If nonempty, return only `AlleleCall`s associated with one of these `Allele`s.
-  */
-  array<string> alleleIds = [];
-
-  /**
-  If nonempty, return only `AlleleCall`s associated with one of these
-  `Variant`s.
-  */
-  array<string> variantIds = [];
-
-  /**
-  Specifies the maximum number of results to return in a single page.
-  If unspecified, a system default will be used.
-  */
-  union { null, int } pageSize = null;
-
-  /**
-  The continuation token, which is used to page through large result sets.
-  To get the next page of results, set this parameter to the value of
-  `nextPageToken` from the previous response.
-  */
-  union { null, string } pageToken = null;
-}
-
-/**
-This is the response from `POST /allelecalls/search` expressed as JSON.
-*/
-record SearchAlleleCallsResponse {
-  /** The list of matching calls. */
-  array<org.ga4gh.models.AlleleCall> alleleCalls = [];
-
-  /**
-  The continuation token, which is used to page through large result sets.
-  Provide this value in a subsequent request to return the next page of
-  results. This field will be empty if there aren't any additional results.
-  */
-  union { null, string } nextPageToken = null;
-}
-
-/**
-Gets a list of `AlleleCall` matching the search criteria.
-`POST /allelecalls/search` must accept a JSON version of
-`SearchAlleleCallsRequest` as the post body and will return a JSON version of
-`SearchAlleleCallsResponse`.
-*/
-SearchAlleleCallsResponse searchAlleleCalls(
-  /** This request maps to the body of `POST /allelecalls/search` as JSON. */
-  SearchAlleleCallsRequest request) throws GAException;
-
 }
--- a/src/main/resources/avro/metadatamethods.avdl
+++ b/src/main/resources/avro/metadatamethods.avdl
@@ -8,10 +8,9 @@ import idl "metadata.avdl";
 /** This request maps to the body of `POST /individuals/search` as JSON. */
 record SearchIndividualsRequest {
   /**
-  If specified, restrict this query to individuals within the given
-  individual groups.
+  The Group to search.
   */
-  array<string> groupIds = [];
+  string groupId;
 
   /**
   Only return individuals with this name (case-sensitive, exact match).
@@ -52,8 +51,8 @@ record SearchIndividualsResponse {
 /**
 Gets a list of `Individual` matching the search criteria.
 
-`POST /individuals/search` must accept a JSON version of 
-`SearchIndividualsRequest` as the post body and will return a JSON version of 
+`POST /individuals/search` must accept a JSON version of
+`SearchIndividualsRequest` as the post body and will return a JSON version of
 `SearchIndividualsResponse`.
 */
 SearchIndividualsResponse searchIndividuals(
@@ -65,8 +64,10 @@ SearchIndividualsResponse searchIndividuals(
 /** This request maps to the body of `POST /samples/search` as JSON. */
 
 record SearchSamplesRequest {
-  /** If specified, restrict this query to samples for the given individuals. */
-  array<string> individualIds = [];
+  /**
+  The Individual to search.
+  */
+  string individualId;
 
   /**
   Only return samples with this name (case-sensitive, exact match).
@@ -107,8 +108,8 @@ record SearchSamplesResponse {
 /**
 Gets a list of `Sample` matching the search criteria.
 
-`POST /samples/search` must accept a JSON version of 
-`SearchSamplesRequest` as the post body and will return a JSON version of 
+`POST /samples/search` must accept a JSON version of
+`SearchSamplesRequest` as the post body and will return a JSON version of
 `SearchSamplesResponse`.
 */
 SearchSamplesResponse searchSamples(
@@ -158,8 +159,8 @@ record SearchExperimentsResponse {
 /**
 Gets a list of `Experiment` matching the search criteria.
 
-`POST /experiments/search` must accept a JSON version of 
-`SearchExperimentsRequest` as the post body and will return a JSON version of 
+`POST /experiments/search` must accept a JSON version of
+`SearchExperimentsRequest` as the post body and will return a JSON version of
 `SearchExperimentsResponse`.
 */
 SearchExperimentsResponse searchExperiments(


jeromekelleher:
Any thoughts on this? The current situation with numerous different conventions for object search is a major impediment to implementation, so it would be good to resolve this soon.

macieksmuga:
@adamnovak could you comment on Jerome's `SearchAlleles` question? I'd very much like to see the proposed simplifications as a server developer, but can't comment on the impact these will have on intended usage.

lh3:
I am fine with the treatment of dataset, but I would recommend to delay the variant part until we are fully confident. For now, I am not. The ideal model in my head is very different from the current schema. I have not sent a pull request because we have some fundamental questions unanswered, such as "what is an allele?" "what is a Variant made of?". We have not reached a consensus yet.

More specifically, you said "Conceptually, it seems natural that an Allele belongs to a single VariantSet". This is also suggested by the current schema. However, I think the central role of the unary model is to break the boundaries between VariantSets. In that view, an Allele is global, not specific to any VariantSet. I have asked this question but have not got a definite answer from others.

On your point 1, my preferred data model is to embed the `AlleleCall` equivalent to another object and get rid of both `SearchCallsRequest` and `SearchAlleleCallsRequest`.

pgrosu:
+10 Heng!  I definitely prefer the model you mentioned.  We want to share as many unique Alleles across multiple VariantSets, which will simplify searches and also is a more natural way to think.  We had a similar discussion about reads way-back here https://github.com/ga4gh/schemas/issues/24, which simplified our current implementation.  I would definitely prefer to see a PR on this, and would be great to hash this out via an in-depth discussion where more folks jump in.

src/main/resources/avro/readmethods.avdl
```diff
@@ -14,8 +14,10 @@ to `ReferenceSet`s containing that same `Reference`. If no reference is
 specified, all `ReadGroup`s must be aligned to the same `ReferenceSet`.
 */
 record SearchReadsRequest {
-  /** If nonempty, restrict this query to reads within the given readgroups. */
-  array<string> readGroupIds = [];
+  /**
+  The ReadGroup to search.
+  */
+  string readGroupId;
```
calbach: This change seems to remove some significant functionality, illustrated by two use cases in particular:

1. This regresses from BAM for the use case of reading all reads in coordinate sorted order for a `ReadGroupSet`, or a single coordinate sorted BAM file. My concerns would be satisfied if there were an additional option to query by `readGroupSetId`.

2. Loss of the ability to do "multi-BAM slicing" on the server side. Of course there is a cost of implementation, but I do believe this cost was considered when defining the SearchReads semantics. I would be curious to know if opinions have changed on the subject.

In Google's implementation of v0.5.1, we support the readGroupIds field but we make the additional requirement that all provided read group IDs belong to the same read group set. We recommend querying by readGroupSetId instead (and we support searching by multiple IDs).

richarddurbin:
I still need to catch up with the exchanges on definition of Allele etc., but it also makes sense to me that Alleles are global top level objects, not specific to a VariantSet. 
I guess they are in the context of a particular ReferenceSet, though in fact they are defined on sequences, and so may be identifiable across reference sets if the same
underlying reference sequences are present.

Richard

> On 18 Mar 2015, at 17:31, Heng Li <notifications@github.com> wrote:
> 
> I am fine with the treatment of dataset, but I would recommend to delay the variant part until we are fully confident. For now, I am not. The ideal model in my head is very different from the current schema. I have not sent a pull request because we have some fundamental questions unanswered, such as "what is an allele?" "what is a Variant made of?". We have not reached a consensus yet.
> 
> More specifically, you said "Conceptually, it seems natural that an Allele belongs to a single VariantSet". This is also suggested by the current schema. However, I think the central role of the unary model is to break the boundaries between VariantSets. In that view, an Allele is global, not specific to any VariantSet. I have asked this question but have not got a definite answer from others.
> 
> On your point 1, my preferred data model is to embed the AlleleCall equivalent to another object and get rid of both SearchCallsRequest and SearchAlleleCallsRequest.
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/ga4gh/schemas/pull/253#issuecomment-83071383>.
> 




-- 

nlwashington:
+10 again for "Allele is global, not specific to any VariantSet."  I have been wondering why this design decision was made in the first place because it seems topsy-turvy.  For the initial G2P schema, I have chosen to use the proposed <code>Feature</code> in the [sequence-annotation](https://github.com/ga4gh/schemas/tree/sequence-annotation) branch instead of <code>Variant</code> for this specific reason.

src/main/resources/avro/readmethods.avdl
jeromekelleher: Thanks @calbach, this makes sense. I'm perfectly happy to add a `readGroupSetId` field; this is logically consistent, and easily implementable. Would this resolve both (1) and (2)?

lh3:
David also said allele is global. In principle, an allele could be species specific, but due to the way we find/name alleles, a GA4GH allele is specific to a ReferenceSet. Different ReferenceSets get different alleles.

src/main/resources/avro/readmethods.avdl
lh3: Do you want to add `string readGroupSetId` or `array<string> readGroupSedIds`? We need the latter to achieve "multi-BAM splicing". I know it has a cost as @calbach said. I would recommend to keep this part of schema as it is and adopts the google approach in implementation for now.

jeromekelleher: I'd be very much in favour of `string readGroupSetId`, for all of the reasons given above. However, I don't really understand what "multi-BAM splicing" is: is it an important application?

jeromekelleher:
Thanks for all the feedback. The consensus seems to be, "this change is OK, but there are some details to be worked out". Is this fair?

I'm perfectly happy for an Allele to be global, but we will need to think about what the implications for search and authentication are. Will we want to search for all the Alleles in a given `VariantSet` or `ReferenceSet`? (That is, an Allele can exist in multiple VariantSets and ReferenceSets) This is fine by me once we limit the semantics to exactly one of either.

In terms of authentication, how do we decide whether a user has permissions to see a given allele? I'd imagine the model of allowing users to search over a given VariantSet or ReferenceSet or not will give us the restrictions we require.

@lh3 - I'm happy to delete both SearchCall and SearchAlleleCall: I don't think they are particularly useful.

pgrosu:
Jerome, I think authentication was deemed to be resource-specific and not part of this API based on the latest post:

https://github.com/ga4gh/schemas/issues/148#issuecomment-76011682

src/main/resources/avro/readmethods.avdl
calbach: Adding `readGroupsetId` addresses (1) only. Reverting this change or adding `array<string> readGroupSetIds` addresses both (1) and (2).

We've had this feature requested in the past, but others can probably provide better commentary on the use cases and how applicable it is today.

jeromekelleher: Thanks @calbach . Just so I'm clear, the way I understand it is we want to retrieve all the reads in an interval (start, end) across ReadGroupSets [s1, s2, ..., sn]. Can the client not achieve this by making individual requests for (start, end) in s1,s2, ..., sn? Unless the number of reads we hit in each set is tiny, the extra overhead of this approach would be negligible.

lh3: The interval could be so large that paging kicks in. In this case, it will be very difficult to make multiple individual requests and then merge them on the client side. It is fine that the server doesn't implement this feature, but we should not limit API. An appropriate implementation may not have this problem. Samtools can merge multiple BAMs in a small region and generate a merged BAM stream, but it is not exposed in htslib.

I think we should not change this part of the schema.

calbach: @jeromekelleher your understanding is correct. I agree with @lh3 that using the proposed workaround will in many cases require pushing complexity down to the client. I also think that the statement about negligible overhead is predicated on assumptions about the implementation, e.g. how the data is stored and ordered on disk. Probably for a BAM-backed implementation you would expect similar performance with either approach.

pgrosu: I am very hesitant to push complexity down to the client.  If that is the case, why even have the server?  I am more in favor of `array<string> readGroupSetIds`, since in the future if we want to drill back to the reads after performing  Matchmaker or Beacon, this is exactly the case that would be needed where it would be across all repositories of stored reads.

jeromekelleher: Ah, I see, now I understand; thanks for your patience @calbach and @lh3. The reads we return are ordered by position, so it's useful for the client to be able to get all of the reads across multiple variant sets back sorted by genomic position.

My point about overhead was more from the perspective of the client/server exchanges @calback --- would there be substantially more protocol chat if we ask the client to make these requests sequentially? I would say the answer is no, assuming the number of records  returned *r* is larger than the page size *p*.

On page size, I think we have to assume that for anything but trivial queries, that *r* >> *p*, and that the client must always assume that there will be paging. Any client library worthy of the name will abstract away the details of paging. For example, in the reference implementation we use iterators for all protocol queries, and paging happens transparently behind the scenes. I think we should assume that there will be a client library, and that people writing real applications will not be hand-crafting JSON documents and dealing with pagination.

Which brings us back to multi-BAM retrieval. Why should we complicate the server and the protocol, when this can _easily_ be achieved by the client? For a *k*-way merge, all I have to do is make *k* iterators (one for each ReadGroupSet) and merge the reads on the fly. The end-user will simply see the required stream of records, in sorted order. 

There is a fundamental tactical choice here: do we (1) build a protocol that is a set of well-defined simple operations that are composable on the client side to give arbitrarily complex behaviour; or (2) do we try to build every conceivable operation that a client might require into the protocol? I think (1) is  preferable, for a multitude of reasons. 

My argument here is of course very extreme: multi-merge is an important piece of functionality. However, I do think it belongs on the client side. I also think that anything that _can_ go on the client side should be left there, until there is a compelling performance case for moving it into the server.

Apologies for such a long reply, but I do think that this is an important point, and feeds into several of the discussions that are happening at the moment.

lh3: We need to find a balance between (1) and (2). Either extreme seems bad to me. On multi-BAM slicing, a right backend could support it efficiently. Doing that on the client side in python can be slow and may impact the server performance.

jeromekelleher: Sorry to labour the point here @lh3, but I just don't understand the arguments for doing this on the server side. It's not about whether a server _can_ do it efficiently, it's whether it _should_ be doing it at all. 

I'm not suggesting we do it client side using Python necessarily, I just used an example of a client library I'm familiar with. There's no reason this couldn't be done in C, if a C client library is written. I don't understand how this would effect server performance either --- servers are written to deal with many concurrent requests. Having a bunch of concurrent requests coming from a single client is no problem. Indeed, one could easily argue that it is better for the server to serve several simple queries concurrently than to need to serially process the equivalent complex query.

I guess the basic question I have is, why not entirely (1)? Since we must have client libraries in any case, why not push complex features that can be composed out of the atomic units the server provides into these libraries? This way, client libraries can rapidly iterate to add interesting and convenient features and the protocol can stay relatively static. When some of these features prove to be very useful and there is a compelling performance argument for moving them on to the server, we update the protocol.

dglazer: Two high-level thoughts:
a) when designing an API from scratch, I agree that we should make thoughtful tradeoffs between extra work for client developers and extra work for server developers, since the overall goal of wide adoption needs both groups to be happy. I'm open to different answers in different circumstances. My general bias would be towards (1) -- when in doubt, keep the server simple -- since it's always possible to add more server options later, but it's hard to remove them. I'd tilt towards (2) -- tell the server enough to let it optimize for you -- when, as @jeromekelleher says, there's a "compelling performance argument".

b) we're not starting from scratch here -- we have a published API with existing server implementations and existing client apps. Making backwards-incompatible changes is a big deal -- even if they only break a few clients, those are by definition our most eager early adopters. If we set the expectation that coding to the GA4GH API is risky, we'll be discouraging people from ever getting on board.

If I read the PR right, there are about a dozen independent changes proposed here. I defer to CH and others on which of them are likely to break existing clients, vs. still being in the "from scratch" state.

pgrosu: Hi David,

Welcome back :)  I like the points your brig up and I agree that we need to find a good balance, so maybe we should try to have that discussion now.  Let's take the Google Maps API, how much is performed on the server vs the client?  The data throughput for today's analysis would be of a similar scale.  Would one perform the [Google Distance Matrix API](https://developers.google.com/maps/documentation/distancematrix/) on their client?  Most likely not.  So we want to keep petabyte sub-second response times in mind as this API and it's implementations with distributed repositories scales.

Regarding adapting the API by taking into account currently implemented versions, has pros and cons.  On the pro-side it would tell us what is possible and what to tweak.  On the con-side, we do not want it to limit us.  I have dealt in the past with similar situations on very-large implementations, and one definitely does not want to migrate the data from one schema to another.  It is a huge pain, which I could image it to be at Google if the schema changes significantly here.  But let me ask you this.  What if in the future we brainstorm a major re-write that would improve the overall implementation?  Should we go forward, or should we stay away if the currently implemented versions of the API would break?  I think the goal is to work through these versions (since we are still sub-1.0 on GA4GH API) until we find the optimal setup.

What do you think?

Thanks,
Paul

lh3: Multi-BAM slicing is a very useful operation that server can optimize but client cannot. In principle, a right backend could merge 1,000 related samples efficiently on the server side. Do we want a client to open 1,000 connections in parallel just to get data in a tiny region? Or are we thinking that we don't need to support this operation because our BAM-based backend cannot achieve it?

pgrosu: Precisely Heng - great example!

richarddurbin: I agree with Heng that the ability of servers to integrate over readgroups seems activity from the user point of view.  The ability to look 
at a (normally small) region of the genome in thousands of samples is a real issue at the moment, and will be more so going forwards, 
and if we had a server that dealt with it that would be a true win.  Where this is in the server/client world I don't know - I can imagine
wanting to integrate across multiple servers each with different repositories I have access to, so I can see some client side integration
as valuable.  But that leaves the question of what the real API is - is it the remote one across the network, or some user code accessible
API?  I think at the user coding level one should have integration across read groups as a basic operation.  As Jerome says, at the
user code level one also wants all the paging to be hidden and just to use iterators across sets of objects.

How is this handled in other APIs?  Is there a distinction between a user API and a low-level network API that can satisfy it, but which
is optimised for simplicity and performance?

Richard

> On 24 Mar 2015, at 20:41, Paul Grosu <notifications@github.com> wrote:
> 
> In src/main/resources/avro/readmethods.avdl <https://github.com/ga4gh/schemas/pull/253#discussion_r27070656>:
> 
> > @@ -14,8 +14,10 @@ to `ReferenceSet`s containing that same `Reference`. If no reference is
> >  specified, all `ReadGroup`s must be aligned to the same `ReferenceSet`.
> >  */
> >  record SearchReadsRequest {
> > -  /** If nonempty, restrict this query to reads within the given readgroups. */
> > -  array<string> readGroupIds = [];
> > +  /**
> > +  The ReadGroup to search.
> > +  */
> > +  string readGroupId;
> Hi David,
> 
> Welcome back :) I like the points your brig up and I agree that we need to find a good balance, so maybe we should try to have that discussion now. Let's take the Google Maps API, how much is performed on the server vs the client? The data throughput for today's analysis would be of a similar scale. Would one perform the Google Distance Matrix API <https://developers.google.com/maps/documentation/distancematrix/> on their client? Most likely not. So we want to keep petabyte sub-second response times in mind as this API and it's implementations with distributed repositories scales.
> 
> Regarding adapting the API by taking into account currently implemented versions, has pros and cons. On the pro-side it would tell us what is possible and what to tweak. On the con-side, we do not want it to limit us. I have dealt in the past with similar situations on very-large implementations, and one definitely does not want to migrate the data from one schema to another. It is a huge pain, which I could image it to be at Google if the schema changes significantly here. But let me ask you this. What if in the future we brainstorm a major re-write that would improve the overall implementation? Should we go forward, or should we stay away if the currently implemented versions of the API would break? I think the goal is to work through these versions (since we are still sub-1.0 on GA4GH API) until we find the optimal setup.
> 
> What do you think?
> 
> Thanks,
> Paul
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/ga4gh/schemas/pull/253/files#r27070656>.
> 




-- 

jeromekelleher: Thanks for all the comments, I think this is a very useful conversation. Some responses:

- @dglazer I appreciate your point about this not being a green-field site, I really do. Breaking compatibility _is_ a big deal, and not something we should take lightly. On the other hand though, this is a pre-1.0 API, and we must have the freedom to make some changes if we are to develop it. The changes here are simplifications, and as you say, if we don't make these now it'll be very difficult to make them later. Users can be shielded from these changes by the client-side libraries quite easily, so I don't think any serious existing applications need be affected at all. For example, `api-client-java` can detect which version of the API a server is running, and send sequential requests when existing code  uses the multi-container search semantics.

- @lh3 This is a good example. If we have the need to merge reads from 1000 `ReadGroupSets`, and the number of reads per `ReadGroupSet` is often zero or much less than the page size, then this would be a compelling performance argument in favour of doing the merge on the server side. I don't want to be dogmatic about this, and if this is a real requirement that can't be worked around then I'm happy to keep searches across `ReadGroupSets`. However, why would we want to search across 1000 ReadGroupSets? If we have reads for 1000 related samples, why not keep them all in a single ReadGroupSet?  I'm happy to add the ability to search over a single `ReadGroupSet` by adding a `readGroupSetId` field as @calbach suggested at the start of this thread. Would this satisfy the most common requirements? As @richarddurbin pointed out, we can't solve this problem on the server side in general in any case, since we might need to merge reads from different servers.

- @richarddurbin This is a very important question I think. What is the real API: the low-level transport layer of HTTP and JSON, or the API that developers actually interact with in their language of choice when writing applications? As a client-side developer (in say C, Java, Python, R or JavaScript), I don't want to know about HTTP, JSON, or paging any more than I want to know  what my TCP packet size is. What I want is a well-defined client-side API in my language with documentation written specifically for that language. A client-side API will look very different in C and R, and they must have native documentation if developers are going to write applications. Unless we provide excellent client libraries that make writing applications substantially easier than using existing tools,  there will be no significant reason for developers to adopt the GA4GH API.

lh3: If a ReadGroupSet contains multiple samples, we will sometimes want to retrieve a subset of samples from a ReadGroupSet. It is not the right unit. ReadGroup is.

diekhans: We have to feel free to break compatibility when necessary.

The API should be versioned, but 0.5 is way too immature to be
anything set in stone.  We have a weak conformance suite and
virtually nothing real (at least that we have seen) has been
built with it.  If we don't validate by building real
applications while the API is being design, we will end up with
a mess.

The world will have to live with this for probably 20 years.  We
can't do a design-by-committee, through over the wall
specification and expect it to be right. This is an absolute
recipe for failure.  We have to build real applications and
iterate.

This is why the reference service is so important to GA4GH's
success.

Mark

dglazer: * I think most of the dozen independent changes in this PR are non-controversial, since they don't break _any_ published code, and they do simplify things. @jeromekelleher, if you want to divide and conquer, I think you could get consensus on over half of the PR quickly, while we continue to discuss the remainder. @calbach --  other than this `readGroupIds[]` proposal, do you know which other edits are potentially breaking changes?  

* @diekhans -- I agree the API shouldn't be set in stone; I also think that we shouldn't casually break compatibility -- that's why this is a good discussion. And I strongly agree that we need to build real applications as part of the design process -- I'm delighted to see the work on the new-and-improved reference server, in addition to the earlier work by EBI, NCBI, https://github.com/googlegenomics/api-provider-local-java, and Google's product.

* we discussed `ReadGroup` and `ReadGroupSet` at great length the first time around. There was passionate support (including by @lh3) to allow retrieving from a single `ReadGroup`, and there was passionate support (including by me) to allow retrieving all reads from a single `ReadGroupSet` -- that's what led to the current method accepting an array of `readGroupIds[]`. There's more than one way to satisfy those two requirements, but I believe that any proposal needs to continue to do so.

pgrosu: Hi David (@dglazer),

Thank you for the thoughtful and insightful comments, but with the highest esteem and most respect, why are we creating limits based on a design decision at Google?  I read most of the Google papers - and many others - and I know what is possible both from them and the different subject areas in computer science.

I posted the following for a reason:

https://github.com/ga4gh/schemas/issues/131

For instance from the Mesa paper - which one can read from [the following link](http://research.google.com/pubs/archive/42851.pdf) - there is a clear line that says:

_We show how schema changes for a large number of tables can be performed dynamically and efficiently without affecting correctness or performance of existing applications._

Data and schemas are two separate things.  I'm not saying let's use Mesa - which would be nice - but let's implement concepts that utilize some of that functionality in a public setting.  If you're up for it, let's re-engineer Google's backend for the GA4GH implementation to make it flexible enough with future implementations.  It is just a service, and that means the code does not have to be public - which currently is not, and thus satisfies that requirement as well.

I have worked at too many places to see unnecessary limits set because of an earlier design decision.  We have an opportunity here to do things right.  Let's put our minds together and make it right.

Let me know what you think.

Thanks,
Paul

jeromekelleher: @dglazer --- fair points! I've updated the PR, and will comment more below, as this discussion thread will be hidden by default by GitHub.

pgrosu:
I think the client/server and the reads/Multi-BAM slicing discussion will prove fruitful for future aspects of the API, so I put  a link to it here in case folks want to refer to it later:

https://github.com/ga4gh/schemas/pull/253#discussion-diff-26700450

Paul

jeromekelleher:
Thanks for all of the comments on this PR so far, it has been a very interesting discussion. Multi-BAM merge is important, so I accept that it should stay.

I've updated the PR to reintroduce the `readGroupIds` field in the `SearchReadsRequest` object, so that it will support merging reads across many ReadGroups. I have changed the semantics slightly, in that we no longer accept the empty list (interpreted as `search all readGroups`) as an argument. I think this is very important.

Any comments on this or any of the other changes would be much appreciated.

pgrosu:
Hi Richard (@richarddurbin),

Regarding a reference server - which would be a big win - we should just start simple just the way search engines are built via an inverted index.  Below are the general steps:

1. The genome is split into equally sized regions.

2. Each region is contains a pointer to a list of sample ids.  A hash function (i.e. md5sum, sha256sum, etc) is generated between the tuple *(region, sample_id)* for optimized lookup of the reads for that sample in the specified region.

3. The hash is then accessed in O(1) for the collection of reads it contains for that region.

This would also enable efficiently parallelized across multiple machines.  The storage can also be optimized via the Log-Structured-Merge (LSM) Tree data structure with Differentiated Indexing (Diff-Index), which I put a reference for below.

##### References

A nice graphical description of an inverted index is shown at the following link: 

https://developer.apple.com/library/mac/documentation/UserExperience/Conceptual/SearchKitConcepts/searchKit_basics/searchKit_basics.html

Regarding a deeper description on inverted indices, below is a nice paper from Google, that contains a description of it:

http://static.googleusercontent.com/media/research.google.com/en/us/pubs/archive/37367.pdf

Below is another nice paper from Google/IBM/GeorgiaTech regarding the Log-Structured-Merge (LSM) Tree data structure and with Differentiated Indexing (Diff-Index):

http://researcher.watson.ibm.com/researcher/files/us-wtan/DiffIndex-EDBT14-CR.pdf

Let me know what you think.

Thank you,
Paul

dglazer:
@jeromekelleher, that multi-BAM change addresses the biggest concern I saw raised -- thanks.

There are 12 separate changes in this PR, and I was finding the diff's hard to read, so I made the summary below. Assuming I got it right, I:
- am +0 on the 6 "not yet published" changes (they look reasonable, but I haven't been part of those design discussions)
- am +1 on the 2 "published but not implemented" changes
- would like to be sure there aren't any bad consequences to the 4 "already implemented" changes.

A few questions:
- @jeromekelleher -- does it look like I correctly caught all twelve changes?
- @calbach -- does it look like I binned them correctly into "not yet published", "published but not implemented", and "already implemented"? (I think I got them right, but please double-check.)
- @calbach -- of the 4 potentially breaking changes, do you have any thoughts on which if any could cause real problems in existing code?

#### 6 not yet published APIs
*Backwards compatibility isn't an issue here.*

**src/main/resources/avro/metadatamethods.avdl**
```
record SearchIndividualsRequest {
-  array<string> groupIds = [];
+  string groupId;

record SearchSamplesRequest {
-  array<string> individualIds = [];
+  string individualId;
```
**src/main/resources/avro/variantmethods.avdl**
```
record SearchVariantSetSequencesRequest {
-  array<string> parentIds = [];
+  string parentId;

record SearchAllelesRequest {
-  array<string> variantSetIds = [];
+  string variantSetId;

record SearchCallsRequest { 
-  array<string> callSetIds = [];
-  array<string> variantSetIds = [];
-  array<string> variantIds = [];
+  string callSetId;

record SearchAlleleCallsRequest { 
-  array<string> callSetIds = [];
-  array<string> variantSetIds = [];
-  array<string> alleleIds = [];
-  array<string> variantIds = [];
+  string callSetId;
```

#### 2 published APIs, with unimplemented behavior
*As far as I know these changes don't break any existing implementations, so again backwards compatibility isn't an issue.*
**src/main/resources/avro/variantmethods.avdl**
```
record SearchVariantSetsRequest {
-  array<string> datasetIds = [];
+  string datasetId;

record SearchVariantsRequest {
-  array<string> variantSetIds = [];
+  string variantSetId;
```

#### 4 published APIs, with implemented behavior
*These changes are small, but might cause a backwards compatibility issue.*
**src/main/resources/avro/variantmethods.avdl**
```
-  array<string> variantSetIds = [];
+  string variantSetId;
```
**src/main/resources/avro/readmethods.avdl**
```
record SearchReadsRequest {
-  array<string> readGroupIds = [];
+  array<string> readGroupIds;

record SearchReadGroupSetsRequest {
-  array<string> datasetIds = [];
+  string datasetId;
```
**src/main/resources/avro/referencemethods.avdl**
```
record SearchReferencesRequest {
-  union { null, string } referenceSetId = null;
+  string referenceSetId;
```

jeromekelleher:
Thanks for the summary @dglazer, this is very helpful. Your summary looks accurate to me. Since the original PR, another set of APIs has been merged in #246, so I've updated the methods here to follow the single-parent-container-unless-absolutely-necessary paradigm. These go into the **not yet published APIs** bucket:

**src/main/resources/avro/sequenceAnnotationmethods.avdl**
```
record SearchFeaturesRequest {
-    array<string> featureSetIds = [];
+    array<string> featureSetId;
   
-    array<string> parentIds = [];
+    array<string> parentId;
``` 

(Apologies for the whitespace noise in the diff --- I can get rid of this if you prefer).

pgrosu:
@jeromekelleher, @dglazer - One thing to keep in mind, it is not uncommon to merge variants or to search across many variant sets at the same time.  So removing this `array<string> variantSetIds = [];` most likely would be very temporary.

Just out of curiosity, if it is removed, does that mean that before querying across multiple VariantSets, I would merge some or all VariantSets into one gigantic one in order to search across multiple ones.  These would be so many that it would defeat the purpose of removing `array<string> variantSetIds = [];`.  My recommendation is to always design the system to match the most natural and flexible means of asking the biological questions if such a search is possible,  which in this case it is.  This is the foundation of MapReduce and relational algebra.

Paul

calbach:
@dglazer yes, that summary looks correct to me.

Aside from the `referenceSetId` change (I'll comment inline), I'm +1 on the other breaking changes and unimplemented changes. I think they don't remove any important functionality, so the tradeoff for implementation simplicity is worth it.

I'm neutral on the unpublished methods.

src/main/resources/avro/referencemethods.avdl
```diff
@@ -113,9 +113,9 @@ as JSON.
 */
 record SearchReferencesRequest {
   /**
-  If present, return only references which belong to this reference set.
+  The ReferenceSet to search.
   */
-  union { null, string } referenceSetId = null;
```
calbach: Seems this is breaking a bit from the many to many relationship which has been discussed previously around references : referenceSets. Have the graph changes altered that discussion?

I'm thinking of the CRAM use-case, for example. Given an MD5, find the relevant reference.

pgrosu: @calbach, see https://github.com/ga4gh/schemas/issues/247 regarding where this is stemming from.  Though I agree that many-to-many is something that I think is still important, it probably requires more clarity with examples in the structure of the definitions.  Just for reference for others, the original many-to-many was discussed initially in the following issue:

https://github.com/ga4gh/schemas/issues/125#issuecomment-54029648

You can still search with MD5 via this in the same method:

```
  /**
  If nonempty, return references which match any of the given `md5checksums`.
  See `Reference::md5checksum` for details.
  */
  array<string> md5checksums = [];

```

~p

diekhans:
+1 [with great enthusiasm]

It's very important to try to make the GA4GH APIs more consistent and rational.  The current state of the queries are a result of different committees as opposed to a single design effort.  So it's done a good job of defining requirements but not one of defining an overall consistent API.

This PR is an excellent step in right direction.  It can and will evolve once we get more experience with using it.  However, we need to make incremental changes like this to move forward.

src/main/resources/avro/referencemethods.avdl
calbach: Pre-graph changes: it wasn't required but it wasn't disallowed. If a repository serves reference sets hg19 and GRCh37, it would be up to the implementation whether chr4 (MD5 is identical in both assemblies) is represented as one or two references. The former is easier to implement and might be required in some cases, e.g. if ACLs are involved, but the latter may be conceptually cleaner. But yes, it should probably be made explicit in the docs that this is allowed.

Things have also changed with the graph PR. Now you can nest reference sets within a reference set (I missed that when I was reviewing the changes). I guess that makes the semantics of this field slightly more complicated and I'm not sure how it plays with many : many.

If reference sets are still many:many with references after the graph changes, then I'd definitely prefer this line be reverted.

jeromekelleher: OK, fair enough. @adamnovak or @richarddurbin could you clarify the status of the many-to-many relationships between references and reference sets in the graph world please? 

(My opinion here would be to try and keep things simple and avoid many-to-many relationships if we can. Implementations are free to reuse the underlying objects representing the same references within their code, but can present them to the outside world as-if they were different objects.)

richarddurbin: We need to allow the same reference to be in many reference sets.  For example the same human mitochondrial sequence will be in 
many different versions of the human genome reference set.  Even the same GRCh38 primary chromosome 1 will be in different reference
sets, with and without EBV, alts, IMGD MHC versions etc.  This is essential for practical use.  This has nothing to do with the graph world, it
is core to the way that people practically use genome reference sequences.

Slightly more radically, I think in the graph world we need the  ability to create extended reference sets that include one previous reference
set and add variants to it.  This amounts to a single inheritance tree.  In the side graph level hierarchy I would require the included reference 
to be first, so that other stuff adds to it.  I believe this cleanly captures how we want to build and extend graphs.

Richard

> On 3 Apr 2015, at 09:39, Jerome Kelleher <notifications@github.com> wrote:
> 
> In src/main/resources/avro/referencemethods.avdl <https://github.com/ga4gh/schemas/pull/253#discussion_r27720795>:
> 
> >    */
> > -  union { null, string } referenceSetId = null;
> OK, fair enough. @adamnovak <https://github.com/adamnovak> or @richarddurbin <https://github.com/richarddurbin> could you clarify the status of the many-to-many relationships between references and reference sets in the graph world please?
> 
> (My opinion here would be to try and keep things simple and avoid many-to-many relationships if we can. Implementations are free to reuse the underlying objects representing the same references within their code, but can present them to the outside world as-if they were different objects.)
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/ga4gh/schemas/pull/253/files#r27720795>.
> 




-- 

jeromekelleher: Thanks @richarddurbin, that's helpful.

@calbach --- I've reverted the change in question.

jeromekelleher:
By my count, we now have three +1s and all of the contentious issues have been resolved. Unless there are any objections, I think this is mergable.

adamnovak:
I'm a bit concerned about losing the ability to search `Call`s and `AlleleCall`s by `VariantSet`. Isn't a `CallSet` allowed to occur in multiple `VariantSet`s? Or did we drop that?

src/main/resources/avro/variantmethods.avdl
```diff
@@ -446,25 +438,9 @@ SearchCallsResponse searchCalls(
...
-
-  /**
-  If nonempty, return only `AlleleCall`s in these `VariantSet`s.
-  */
-  array<string> variantSetIds = [];
-
-  /**
-  If nonempty, return only `AlleleCall`s associated with one of these `Allele`s.
-  */
-  array<string> alleleIds = [];
```
adamnovak: How would you go about looking up how a particular `CallSet` calls a particular `Allele` without this field? Do you have to look at all the `AlleleCall`s?

benedictpaten:
Once Adam's query is resolved, I'm 1+.

jeromekelleher:
I don't really understand what we need the `SeachCallRequest` and `SeachAlleleCallRequest` for  @adamnovak. What's the use case for getting Calls outside of the context of the encapsulating `Variant`?

pgrosu:
@jeromekelleher, this has applications to the Beacon project:

*The Beacon project is a project to test the willingness of international sites to share genetic data in the simplest of all technical contexts. It is defined as a simple public web service that any institution can implement as a service. The service is designed merely to accept a query of the form “Do you have any genomes with an “A” at position 100,735 on chromosome 3?” (or similar data) and responds with one of “Yes” or “No.” A site offering this service is called a “beacon”.*

~p

adamnovak:
@jeromekelleher I agree that `SeachCallsRequest` and `SeachAlleleCallsRequest` are maybe not useful to have, if all you can do is iterate through the `Call`s or `AlleleCall`s in a `CallSet`. The primary purpose of `SearchCallsRequest` was, I think, to allow paging through the `Call`s of a `Variant` that has too many of them. If you're OK with forcing people to get all the `Call`s at once, there's not much point in the method.

On the `AlleleCall`s side, if we drop all those parameters from `searchAlleleCalls()`, or drop the method altogether, then we also need to make `Allele` hold `AlleleCall`s like `Variant` holds `Call`s, or to come up with another way to retrieve `AlleleCall`s. As far as I can tell, the current proposal offers no way to retrieve `AlleleCall`s other than to iterate through all of them for a `CallSet` until you find the one you want.

So in order for me to like this PR, it needs to:

1. Provide/document a way to take a `CallSet` ID and a `Variant` ID and retrieve the `Call` object (even if that is just "download the whole `Variant` and scan it").

2. Provide/document a way to take a `CallSet` ID and an `Allele` ID and retrieve the `AlleleCall` object(s). I don't think "page through all the `AlleleCall`s in the `CallSet` until you find the one for your `Allele`" is going to be good enough, but that's the only method that looks possible here.

lh3:
While I much prefer to drop `SeachCallsRequest` and `SeachAlleleCallsRequest`, I agree with @adamnovak that simply removing them without heavily modifying other part of the schema will lead to a non-workable schema. I think we should not touch `SeachCallsRequest` and `SeachAlleleCallsRequest` in this PR.

jeromekelleher:
Excellent, thanks for this @adamnovak and @lh3. I think one of these is easy to resolve, but I'm not sure about the other (mostly because I don't understand the subtleties of Allele and AlleleCalls). I have a couple of questions/suggestions, and hopefully we can arrive at a solution quickly. 

To address @adamnovak's points above:

1. This is simple I think: just provide the callSetId in the `callSetIds` field in searchVariants. There's no other way of getting the variant's ID, so you might as well get the Call at the same time as you get the variants. I see the point about having large numbers of calls, but I think we should probably fix that problem when we hit it. So, I would be in favour of just deleting `SeachCallsRequest` as it doesn't really serve much purpose.

2. Sorry for the stupid question, but would it be possible to embed allele calls into the variant object in the same way that they are for classic mode (ie., just make the calls field a union of `{`Call``, ``AlleleCall``}), and we can then use the same API for both? If not, I'm happy enough to revert this change with a more focused `SearchAlleleCallsRequest`, with say, `variantId` and `callSetIds` as arguments.

jeromekelleher:
Any thoughts here @adamnovak? I'm happy to reinstate a version of `SearchAlleleCalls`, I'd just like to get your views on the questions above first.

jeromekelleher:
Bump... @adamnovak, any thoughts on the above?

pgrosu:
Forcing searches only to one `Dataset` id is extremely limiting.  This goes against the whole idea of querying all datasets for determining similar variants.  This is the foundation of Matchmaker and Beacon.

lh3:
> Forcing searches only to one Dataset id is extremely limiting. This goes against the whole idea of querying all datasets for determining similar variants.

More generally, we need to associate `Dataset` with meta data and search for them. When there are thousands of datasets (think about SRA), not being able to search across datasets is a deal breaker. I used to propose something a year ago but at that time others thought that was not the task of read/refvar groups. It is still an unsolved problem.

jeromekelleher:
I thought we'd agreed at this point that a single dataset was a good idea and were resolving the finer issues of how we search for calls and alleles and so on.  Are you disagreeing with this @lh3, or making a different point about what `Datasets` are and how we should use them? I agree that the Dataset concept is far too vague right now (see #248), and it needs a lot of work to make it into a useable concept.

lh3:
I guess I was just generally talking about long-term goals and what we are missing most in this course. It is fine to use a single-dataset for now, but proper support of multiple datasets, the server is likely to remain as a prototype.

pgrosu:
@jeromekelleher, if we limit the schema now, then people will start implementing clients with such a limitation in mind and it will be impossible to turn back with large datasets already loaded using a specific design.  I think this is already a reality for some.  We should not merge this until we solve this issue as @lh3 clearly elucidated.  One will not be able to perform the searches easily that we critically need if we limit ourselves too much in the schema design.  Here is a scenario that would throw us but we would need to perform, using my classic example:

So I open my Genome Browser and select a specific region(s) of a chromosome(s).  I then want to search across all datasets for `VariantSet`s where I filter by `VariantSet.metadata.info` for a specific cancer subtype, setting chromosome via `Variant.referenceName`, in the region [`Variant.start` - `Variant.end`) with  `Call`s filtered by `Call.genotypeLikelihood` to find those rare mutations.  Then collecting these as an enriched set, I want to find similar ones to mine that are in other datasets that are not cancer and tell me the samples that they come from.

Thus we would need to provide a set (array) of `Dataset` ids, and not just a specific value.

By looking at what #248 provides, it has some benefits but would be limiting in other aspects. For instance, having *one call belonging to exactly one `Variant` and one `CallSet`* might make indexing costly, though there are other approaches that would cut down those costs.  We are still below version 1.0, and adding too many limitations would limit the potential of the possiblities of what we can do with the API.

We need to make the `schemas` at least the most flexible because the technology for performing these types of mega-searches is already here for several years, and we should not limit something that would be difficult or impossible to revert back to.

Paul

jeromekelleher:
@lh3 Thanks for the clarification. I agree we absolutely need to support datasets properly and define what they mean. I don't think this means that we need to make all of our methods take multiple parent containers  as arguments though: there's no reason a client can't send off a bunch of requests, one for each dataset. It's not limiting *at all* in terms of what can be done with the API to restrict the attention of each individual API call to a single parent container.

lh3:
But can we search for a sample across multiple datasets? And this is similar to retrieving across read groups. Say I want to get the allele frequency across thousands of datasets. Ideally, this should be done with one query. We wouldn't want to query fire 1000s queries just to get this number on the client side and sometimes querying individual datasets might be prohibited due to security concerns.

Anyway, this is for future. For now, I am fine with the current design.

jeromekelleher:
This is a good point @lh3, but I think we'd want a specialised query for this case (it sounds very much like a Beacon-style query). Anyway, we agree that the approach in this PR is good for the short/medium term and will provide us with something workable that we can make progress on. Excellent.

Now we just need to resolve the issues relating to callsets/alleles. @adamnovak, can you respond to my questions above from ~3 weeks ago please, or at least give some sort of indication of where you are with this? I would be in favour of simply removing the `SearchCalls` and `SearchAlleleCalls` requests, as `SearchCalls` is of very limited utility, and the model for dealing with Alleles is still under discussion. We re-can add API methods for searching for Allele calls when the model allele has been agreed on.

pgrosu:
@lh3 For what we're are asking, we would need to implement distributed, replicated balanced inverted indices like this:

samples_datasets_inverted_indices

Paul

src/main/resources/avro/sequenceAnnotationmethods.avdl
```diff
@@ -15,16 +15,16 @@ protocol SequenceAnnotationMethods {
   */
   record SearchFeaturesRequest {
     /**
-    The annotation sets to search within. Either `featureSetIds` or
-    `parentIds` must be non-empty.
+    The annotation set to search within. Either `featureSetId` or
+    `parentId` must be non-empty.
     */
-    array<string> featureSetIds = [];
+    array<string> featureSetId;
```
adamnovak: If this is just one item maximum, shouldn't it be a `union {null, string}`?

src/main/resources/avro/sequenceAnnotationmethods.avdl
```diff
@@ -15,16 +15,16 @@ protocol SequenceAnnotationMethods {
...
 
     /**
     Restricts the search to direct children of the given parent `feature`
-    IDs. Either `featureSetIds` or `parentIds` must be non-empty.
+    ID. Either `featureSetId` or `parentId` must be non-empty.
     */
-    array<string> parentIds = [];
+    array<string> parentId;
```
adamnovak: Same here; a single ID probably shouldn't be an array.

src/main/resources/avro/variantmethods.avdl
```diff
@@ -69,10 +68,10 @@ JSON.
 */
 record SearchVariantSetSequencesRequest {
   /**
-  If nonempty, return `Segment`s describing novel sequences in the `VariantSet`
-  that are joined onto sequences with one of the specified IDs.
+  Return `Segment`s describing novel sequences in the `VariantSet`
+  that are joined onto sequences with the specified ID.
```
adamnovak: Maybe "the sequence"?

adamnovak:
Sorry I wandered off.

I don't think that access to `AlleleCall`s can be effectively provided by embedding them in the `Variant` object, because in the allelic model you will commonly have `AlleleCall`s on `Allele`s that do not participate in `Variant`s at all, and so there is no `Variant` object that could hold them.

A more focused `SearchAlleleCallsRequest` that takes `alleleId` and `callSetId` might be the way to go with this. I originally wanted to embed the `AlleleCall`s in the `Allele`s like the `Call`s in `Variant`s, but since we are moving towards `Allele`s as global across `VariantSet`s I don't think that would actually work.

I am definitely +1 on the single dataset search for now. We might need to add back search across multiple data sets in a single query later, for efficiency, but it's not good to optimize that step before we know it's slow. For now we can send several queries when we need to use several data sets. Remember that we do have a `/datasets/search` (in readmethods.avdl, which might not be appropriate anymore) to loop over all of them.

src/main/resources/avro/sequenceAnnotationmethods.avdl
jeromekelleher: Yes, well spotted. Fixed.

src/main/resources/avro/sequenceAnnotationmethods.avdl
jeromekelleher: Also fixed.

src/main/resources/avro/variantmethods.avdl
jeromekelleher: Fixed; thanks.

jeromekelleher:
Thanks for the feedback @adamnovak. I've fixed the bugs you pointed out. I've also removed the `SearchCalls` and `SearchAlleleCalls` endpoints, in an attempt to get this PR through. I agree that it's not good to have no way of searching for allele calls in the API, but I propose that we open an issue on this and work out the right way to resolve it there rather than holding this PR up any longer. It has been open for over two months now.

Since it has been open for so long, and it classifies as a major enhancement, I think we need to update  the votes. @delagoya, can you ping the relevant people please so we can get a final vote on this?

delagoya:
Since this touches so many files I added a few labels for this issue. If your group does not have a comment on it, please let me know and I will remove the label

delagoya:
## Major Enhancement

The commit touches `readMethods.avdl` which was a major artifact from the Reads Task Team, which is no longer holding regular meetings. As such the, [major enhancement review process](https://github.com/ga4gh/schemas/blob/master/CONTRIBUTING.md#retired-task-teams) will be applied to this pull request. 

-1

To clear the vote, the following people need to respond with **+1** votes after this message for explicit approval of the changes:

@jeromekelleher (done)
@fnothaft 
@dglazer 
@vadimzalunin 
@saupchurch 
@lh3 

(P.S. if any of the people above do not want to be part of future explicit approval list, let me know).

fnothaft:
I am +1.

saupchurch:
+1 for me.

I agree with @adamnovak concerning early optimization.   Single search (thus multiple request) is fine for now with the caveat that usage metrics once we have them may require us to re-visit this in future.

benedictpaten:
Hi Angel, appreciate you watching out for reads people, but why the need to
get all the people from the retired reads group to 1+ the pull request?
Surely just a couple of 1+s from ex reads members would be sufficient? - it
would have been when the reads group still existed!

delagoya:
Precisely because the changes effect the implementations made by a group that no longer meets regularly. 

I'm really just acting as a box checker, so if there is a process change that needs to be made please put it on the agenda for the DWG meetings or for Leiden.

benedictpaten:
Fair enough. I guess I'll make a pull request against the contributions doc!

dglazer:
Thanks @delagoya -- I don't see any concerns; I'll take a closer look in the morning and vote officially.
(I wish GitHub made it easy to see all changes made between my earlier +1 at the end of March and the latest version.)

pgrosu:
@dglazer I think you are in luck, and it's a piece of cake :)  Just follow link below and adjust accordingly the link parameters. In this case I am comparing `master` from 63 days ago against @jeromekelleher's present pull request :)

https://github.com/ga4gh/schemas/compare/master@%7B63days%7D...jeromekelleher:rationalise-search-requests

To view the changes just go to the split view, by clicking first on the _**File Changes**_ tab and then selecting the _**Split**_ button, as highlighted by the red arrow in the screenshot below:

compare63days

Let me know if this is what you were looking for, or if it was something different.

Thanks,
Paul

dglazer:
+1 on the changes to already-published methods (which afaict are the same as the last time I looked closely in March), and +0 on the changes to not-yet-published methods. Thanks @jeromekelleher for pushing this forward.

Small process suggestion for next time -- break things up into smaller pieces. For example, I think the big changes to brand-new methods, like `{search,get}VariantSetSequences` and `search[Allele]Calls`, could easily  have been split out, as could the changes to comment formatting. I expect that would make the discussion and approval process much quicker and easier.

P.S. In case it's useful to others, here are the 5 substantive edits to already-published methods that I found: 
**src/main/resources/avro/variantmethods.avdl (not yet implemented)**
```
record SearchVariantSetsRequest {
-  array<string> datasetIds = [];
+  string datasetId;

record SearchVariantsRequest {
-  array<string> variantSetIds = [];
+  string variantSetId;
```

**src/main/resources/avro/variantmethods.avdl**
```
-  array<string> variantSetIds = [];
+  string variantSetId;
```

**src/main/resources/avro/readmethods.avdl**
```
record SearchReadsRequest {
-  array<string> readGroupIds = [];
+  array<string> readGroupIds;

record SearchReadGroupSetsRequest {
-  array<string> datasetIds = [];
+  string datasetId;
```

jeromekelleher:
Thanks @dglazer, I appreciate the feedback. I agree about making PRs smaller when possible. Apologies for making it difficult to see the additional changes: this is the drawback of squashing commits before a PR is fully ready for merging.

delagoya:
@vadimzalunin and @lh3 there were some changes since you last approved. Since there is an active working group and two out three major implementations weighing in as +1, I'll give another day to respond before I remove the -1 process vote. Thanks!

jeromekelleher:
bump... I think we need to @lh3 and @vadimzalunin to weigh in here?

delagoya:
@jeromekelleher let me reach out via email. If they have not responded by tomorrow, we should take a look at the process to see if there is a time limit and best effort for further review.

vadimzalunin:
+1

jeromekelleher:
@delagoya, I think we're still waiting on @lh3's +1 here?

delagoya:
Ample time a multiple contact tries with now response, so I'm going to call it and merge.

jeromekelleher:
Thanks @delagoya!

PR #258, opened 2015-03-13
Update to the current version of the Matchmaker API

buske:
This PR brings the Avro doc up to our current version. I moved the file out of `wip` since our API has been rather stable lately and we have a few live implementations now, but I'm not sure what the criteria are.

Since our last update, we've simplified the overall structure of the API. Everything is synchronous, so there is no need for query storage on the server. The request and response use the same `Patient` record, and we've tried to clarify a lot of the confusion around the way candidate genes and variants work. We also try to capture varying levels of granularity of genomic information (e.g. candidate gene; gene and variant type; gene, variant type, and zygosity; gene and specific variant).

I put a brief overview in the header doc, and have copied it below for convenience. I've tried to distill the areas of overlap with the GA4GH API, and what I see as the primary obstacles to merging some of our representations. Comments and feedback are very welcome. 

> The MME API currently adopts a "query by example" approach to matchmaking. The request is a (real) patient description, and the response is a list of the most similar patients on the remote server (based on the similarity of the phenotypic and genotypic features).
> 
> Considerations for further integration with the main GA4GH API:
> * Have the following use (or be represented as) an OntologyTerm:
>   `Feature`, `Disorder`, `GenomicFeatureType`, `ageOfOnset`, `inheritanceMode`, `species`, `sex`.
>   - Obstacle: the `ontologySource`/`id` format is not currently compatible with the CURIE-style ontology IDs we use
> * Utilize the GA4GH Variant record.
>   - Obstacle: we have an allele-based (instead of position-based) representation
>   - Obstacle: our variants seldom have IDs, which are required
>   - Obstacle: our variants are usually curated, not called, so having a `VariantSet` and the notion of a `call` seem excessive

We did merge on `geneticSex`, but the GA4GH API recently changed it's representation, so we're now a bit behind. :)


jeromekelleher:
This looks great, thanks @buske. One thing I really like about this is the documentation of the HTTP status codes, which I think we should also have on the `GAException` object. However, this definition does differ slightly to the `GAException` in two ways:

1. `errorCode` is mapped explicitly to the HTTP status. Why do we need to send HTTP status twice? In the reference server we set the HTTP status to an appropriate value depending on the context, following the same sort of rules used here. We then fill `errorCode` with an integer that uniquely identifies the exception that occured.
2. `message` contains embedded JSON. Why do we need embedded JSON? Isn't the combination of HTTP status, the `errorCode` and a human readable error message sufficient?

In my view, it would be great to see us merge the explicit documentation of the HTTP status values used here into the `GAException` object, and hopefully converge on a single exception type that meets everyone's needs.

buske:
@jeromekelleher Thanks for your feedback! I think we're completely in agreement. A few clarifications:
1. Regarding `errorCode`, the only reason to have it twice was to align with GAException and to provide a place in the file to describe the HTTP status codes, rather than only describing it in comments. I will move this HTTP code specification to the comments for the MatchmakerException.
1. It was not my intent to say that `message` included embedded JSON, but rather that the message body of the HTTP error response should still be JSON-formatted, rather than contain a raw message. Indeed, `message` should simply be a human-readable string. I will clarify this in the comments.

jeromekelleher:
Ah, that makes sense, thank @buske. I'd like to see us being a _lot_ more precise about error handling. I agree it's unfortunate that the HTTP status codes live outside the context of the protocol. The best approach I think is to be as prescriptive as possible in the comments, and treat this as a specification. We'd have something like

- HTTP status must be set to an appropriate error value (which we elaborate);
- `content-type` must be "application/json", and the response must be a well-formed JSON document.
- The JSON must conform to the GAException type: `errorCode` is an integer that can be used to  identify the error that occured on the server, and `message` is human-readable information.

What do you think? Could we merge `GAException` and `MatchmakerException` under a definition like this?

buske:
As long as `errorCode` has a default value (e.g. -1) so that it needn't always be specified, we should be able to merge these representations. Though, I disagree with the `application/json` requirement for the following reason.

Regarding content type, we've adopted a custom content type as part of the `Content-Type` and `Accept` headers (`application/vnd.ga4gh.matchmaker.<version>+json`), which we use for specifying the API version. I should probably add this description to our Avro doc as well. Here are the details: https://github.com/MatchMakerExchange/mme-apis/blob/v1.0/search-api.md#versioning

Relequestual:
@jeromekelleher I'm curious; why have you defined `errorCode` in `GAException` if it only mirrors HTTP status codes? As you say, we don't need to send the status code twice.

jeromekelleher:
Sure, `errorCode` can have a default. I like the `Content-Type` versioning, I think we should adopt something similar for all the APIs. That's a job for another PR though! (In general, I think the MatchMaker API has done a _much_ better job of nailing down the HTTP details, and we should try to port as much of this as possible into the other APIs.)

@Relequestual, in my interpretation `errorCode` isn't for HTTP status, it's used as a code to allow internal server diagnostics. For example, in the reference server, we have a bunch of exceptions that can occur, and each of these has a different code. We can then use this to precisely identify the exception that occured when working with error messages. However, the docs aren't exactly crystal clear on this, so it's open to interpretation.

Relequestual:
@jacmarjorie ah right. In which case, the two should probably stay seperated. As in, we probably do need to have and use both (where appropriate).

jeromekelleher:
@Relequestual, I think so, yeah. HTTP status _must_ be set to an appropriate value; `errorCode` may be set to a value that enables more specific server diagnostics (if the server supports this), and `message` is a human-readable message.

buske:
@mbaudis, would love to hear your (and anyone else from Metadata's) comments re: ontology term structure, or overlap with `Individual`?

Can anyone from RefVar can speak to `Variant` and candidate gene coordination?

Thanks, all.

adamnovak:
I'm going to +1 this based on the presentation on the call today, and a bit of looking it over. I think there's definitely some integration work to be done, especially with respect to `Variant` (since assemblies can now contain references with only IDs and no names) and the individual metadata stuff. However, that's probably best done as new PRs on top of this.

adamnovak:
Has anyone else looked this over? @dcolligan? @diekhans?

dcolligan:
@adamnovak I'm not sure if I'm the best person to review this, but I don't see any obvious syntax red flags.

buske:
Thanks for following up @adamnovak. Would love if at least one more person could take a look and approve the PR.

@jeromekelleher Did we sufficiently respond to your thoughts on status/error codes, or is there anything you think it would be good to still clarify in Avro?

jeromekelleher:
@buske Yes, it looks good to me; sorry for lack of response on this.

skeenan:
This seems to have has stalled. Is this good to merge?

buske:
Closing as this PR is stalled and does not satisfy the subsequent reference server and compliance test requirements for merging into the main API.

Issue #212, opened 2014-12-15
look up all reads in fragment


saupchurch:
For fragment reconstruction when dealing with RNA data we need to have the full CIGAR as well as position of the mate pair.  A method to query all the reads for a fragment would allow this as well as support cases (i.e. middle reads) containing more than a single paired read for a fragment.

saupchurch:
Paul,  thanks for taking a stab at this.  Delagoya has the essence of it - the plan is to implement the Fragment portion of the API hierarchy.  The driver behind this as discussed on the call is the need to have the mate pair CIGAR in addition to position for fragment reconstruction when on the RNA side of the things due to splicing and other potential modifications.  It is also desirable to have the structure in place to handle multiple reads in a fragment for future proofing against things like middle reads.

pgrosu:
Hi Sean,

Thank you for the reply and the helpful information.  I will work through several implementations until one seems most clear/intuitive, and then post something back.  I might have a small question here and there, but this now gives me a good foundation from which to work from.  

If there are any additional meeting notes/diagrams/docs/presentations from the call regarding this, that would be nice to sync with.

Thanks,
Paul

saupchurch:
Paul, how is this going?  Do you have any further questions?

pgrosu:
Sorry for the delay.  I got a little vacation this week so I should have time to complete this now.

pgrosu:
Hi Sean and Angel,

Sorry for the delay and I wanted to post this here before performing a pull-request, in order to open it up first for discussion.

So without further ado, below is the most succinct implementation I am thinking of:

```
record Fragment {

  /* The fragment ID. */
  string id;
  
  /* The ID of the read group this fragment belongs to. */
  string readGroupId;
 
  /* The collection of ReadAlignment which belong to the same fragment. */
  array<ReadAlignment> readAlignments = [];
  
}

record ReadAlignment {

  ...
  
  /* The Fragment that this ReadAlignment belongs to. */
  string fragmentId;
  
  ...

}
```

The array of `ReadAlignment` might not be necessary under `Fragment`, if `ReadAlignment` references a specific `Fragment` via `fragmentId`.

Please let me know what you think. 

Thank you,
Paul

saupchurch:
[...]

PR #259, opened 2015-03-14
Added Fragment to Reads

pgrosu:
This pull-request implements the Fragment representation for Reads, which is mentioned in more detail in https://github.com/ga4gh/schemas/issues/212.

### DIFF
--- a/src/main/resources/avro/reads.avdl
+++ b/src/main/resources/avro/reads.avdl
@@ -216,8 +216,8 @@ record GraphAlignment {
 }
 
 /**
-Each fragment represents a single stretch of a DNA molecule. This enables
-fragment reconstruction of reads which belong to the same fragment.
+A fragment represents a contiguous stretch of a DNA or RNA molecule. Reads can
+be associated with a fragment to specify they derive from the same molecule.
 */
 record Fragment {
 
--- a/src/main/resources/avro/reads.avdl
+++ b/src/main/resources/avro/reads.avdl
@@ -215,6 +215,16 @@ record GraphAlignment {
   array<CigarUnit> cigar = [];
 }
 
+/**
+Each fragment represents a single stretch of a DNA molecule. This enables
+fragment reconstruction of reads which belong to the same fragment.
+*/
+record Fragment {
+
+  /** The fragment ID. */
+  string id;
+
+}
 
 /**
 Each read alignment describes an alignment with additional information
@@ -238,6 +248,9 @@ record ReadAlignment {
   string readGroupId;
 
   // fragment attributes
+  
+  /** The fragment ID that this ReadAlignment belongs to. */
+  string fragmentId;
 
   /** The fragment name. Equivalent to QNAME (query template name) in SAM.*/
   string fragmentName;


benedictpaten:
This seems okay to me, being pedantic, maybe change the comment:

"Each fragment represents a single stretch of a DNA molecule. This enables
+fragment reconstruction of reads which belong to the same fragment."

to

"A fragment represents a contiguous stretch of a DNA or RNA molecule. Reads can
be associated with a fragment to specify they derive from the same molecule."

saupchurch:
Good point.  While we are likely to do a documentation pass on the schemas as a whole at some future time it certainly doesn't help to get this in shape from the start.

pgrosu:
I like precision as well.  Sean, after updating locally do I just push it again?  I haven't done an update to a PR after pushing pull-request before.

Thanks,
Paul

saupchurch:
Once you commit locally and push to GitHub it should update the pull request.

pgrosu:
Amazing!  

Many thanks Sean and Benedict,
Paul

benedictpaten:
+1, @saupchurch I assume you're 1+ too?

fnothaft:
I was expecting the more hierarchical approach described in #212:

```
record Fragment {

  /* The fragment ID. */
  string id;

  /* The ID of the read group this fragment belongs to. */
  string readGroupId;

  /* The collection of ReadAlignment which belong to the same fragment. */
  array<ReadAlignment> readAlignments = [];

}
```

Just wondering what drove the change.

saupchurch:
It seemed a lighter touch to me to pull the reads in a fragment by searching on the fragmentId in reads rather than have an object array in Fragment.  Both approaches are acceptable to me; I guess it depends on how we want to query interface to look.

saupchurch:
@fnothaft Do you have strong feelings towards the hierarchical approach or are we good to merge?

fnothaft:
I am OK with merging; we can always modify later.

delagoya:
+1 looks good to me.

pgrosu:
Thank you :)

dglazer:
Hi all -- apologies for chiming in a bit late, but I have both process and substance concerns on this PR.

Re process:
* I don't know if all the stakeholders have paid attention. When we stopped the regular Reads Task Team meetings, we discussed the risk of PRs not getting noticed, and I think we agreed on some sort of louder "call for closure", with appropriate waiting period, to compensate for the lack of meetings. In particular, I'd want to be sure that server implementors have had a chance to weigh in, since changes to an existing spec make work for them. I don't remember if we agreed on specifics, though -- @skeenan, do you remember more details?

Re substance:
* I don't understand how this change would be used as is. It looks like it defines a new `Fragment` datatype consisting of nothing but an id, and then adds a `fragmentId` to the `ReadAlignment`, next to the existing `fragmentName` field. Without any methods changes, how does that help application developers?
* If we were to add a new method that uses `fragmentId` (e.g. *"a method to query all the reads for a fragment"*, as discussed in #212), how would that be any better than just adding a method that searches by `fragmentName` instead?
* Adding such a method is potentially expensive, since it likely requires server builders to maintain an index of all `ReadAlignment`s by `fragmentName`. That might be worth it, but I'd want to hear from application developers to understand how much work it saves them, and hear from server builders on how much work it costs them, and make sure the tradeoff makes sense.

skeenan:
@dglazer I can't comment on substance save to note that the PR has received more than 4 weeks of discussion and 3 +1s

>-- @skeenan, do you remember more details?

the [minutes](https://groups.google.com/d/msg/ga4gh-dwg/ux5O9KUbcD0/rJmZoQrBd1cJ)  of 28th January DWG call record, amongst other things, that we agreed discussions on Reads API modifications would be conducted via GitHub Issues/PRs and voting. We also trailed and discussed this PR on our [25th March DWG call](https://groups.google.com/d/msg/ga4gh-dwg/qOXljRyRl5k/BtuSq4qyGtEJ)

delagoya:
Hi @dglazer some comments below

``` 
Re process:

- I don't know if all the stakeholders have paid attention. When we stopped the regular Reads Task Team meetings, we discussed the risk of PRs not getting noticed, and I think we agreed on some sort of louder "call for closure", with appropriate waiting period, to compensate for the lack of meetings. In particular, I'd want to be sure that server implementors have had a chance to weigh in, since changes to an existing spec make work for them. I don't remember if we agreed on specifics, though -- @skeenan, do you remember more details?
```

I've been tasked with looking through the issues on a biweekly basis to see if there are issues or PRs that would have effected the `ReadTaskTeam`, if this were still an active working group. Those issues are to be tagged with the `ReadTaskTeam` label and discussion is solicited via the issue, or bringing to the attention of the larger DWG committee during those meetings. 


```Re substance:

- I don't understand how this change would be used as is. It looks like it defines a new Fragment datatype consisting of nothing but an id, and then adds a fragmentId to the ReadAlignment, next to the existing fragmentName field. Without any methods changes, how does that help application developers?
```


The motivation comes from the way the RNA-Seq community works with read data, and how that differs from DNA. `fragmentName` is problematics, since a lot of tools change `QNAME`, and hence `fragmentName`, to add information at the end to signify which paired end this read represents. In FASTQ, the QNAME are equal, but the reads are in separate files, with filename representing the read direction. A naive FQ -> SAM translation would lose that information. 

```
- If we were to add a new method that uses fragmentId (e.g. "a method to query all the reads for a fragment", as discussed in #212), how would that be any better than just adding a method that searches by fragmentName instead?
```

See above, but essentially it would have to be a prefix-matching index. Possible, but not ideal, and who knows what other changes to QNAME happen out there. fragmentID makes the semantics much clearer when translating from other formats. Up for discussion is whether we also want orientation information preserved.

dglazer:
@skeenan , @delagoya -- thanks for responding. (And @delagoya, thanks for your ongoing issue curation.)

Re process -- sorry I missed the March 25 DWG meeting (I was talking at a conference); I'll read meeting minutes more closely in the future. I was following the PR itself, but since it only had a single +1 up until two days ago, I didn't treat it as urgently as I should have.

Re substance:
* so you're saying that `fragmentName` (aka QNAME) is not unique in the real world? I had missed that point in the discussion about this issue, and comments in the schema. Unfortunate, and feels broken to me (why even bother having, and storing, a fragment name if one fragment can have two names?), but if we have to live with it, let's at least document it.
* I'm curious -- what tools mess with QNAME, and is there a good reason for doing so? (Given that SAM already has flags that indicate which paired end the read comes from, that we've captured using the `readNumber` field.)
* even if we can't use `fragmentName`, I don't think adding a `fragmentId` adds any value unless we also define and implement new methods. If I'm missing something (certainly possible), please explain. If not, I suggest we revert the 'new noun' until we also agree on the 'new verb' that makes it useful.

lh3:
When a read pair has multiple mappings, it is insufficient to keep track of all the fragment alignments just by QNAME. There are several ways to add the information. How are star/tophat/etc dealing with this issue? It is important to make sure the feature in the schema used in practice.

pgrosu:
Tophat seems to have an interesting approach as I quickly looked through the code.  It might be good if someone else double-checks this too, since I did it rather quickly.  It uses a `BowtieHit` for a specific hit:

```
/*  Stores the information from a single record of the bowtie map. A given read
    may have many of these.  Reads up to 255bp are supported. 
*/
struct BowtieHit
{ ...
```

The whole structure can be seen here:

https://github.com/infphilo/tophat/blob/fb406c837accbf60e8b328f919ce8fe5e6a7d0c4/src/bwt_map.h#L75-143


Then for a read one has a vector of these:

```
struct HitsForRead
{
	HitsForRead() : insert_id(0) {}
	uint64_t insert_id;
	vector<BowtieHit> hits;
};
```

This is listed here:

https://github.com/infphilo/tophat/blob/fb406c837accbf60e8b328f919ce8fe5e6a7d0c4/src/bwt_map.h#L1033-1038

I haven't had much time to look carefully at STAR, though it's coded approach is understandably very different.

Paul

saupchurch:
One of the drivers for the change is the need to be able to reconstruct the fragment information for reads mapping to smaller constructs like lncRNAs where having only the position of the matched read(s) will not be definitive enough.  Association with a fragmentId would allow the downstream user to link up the specific reads.  As currently defined, there are no 'verbs' for Fragment.  I'm an uncertain if the API should support a query for Fragment or leave it to the downstream user.  If a Fragment query is not to be supported then perhaps we don't need to have the definition.  Regardless, there is a need to match a read with its one or more mates by something more definitive than position alone.

pgrosu:
I'd be happy to do another PR to add the Request and Response methods.

dglazer:
@saupchurch, I agree that without Fragment queries there's no point in adding a Fragment definition; what I'm not clear on is whether adding Fragment queries helps app developers enough to be worth the cost to server developers. Do you have any pointers to sample code that does the reconstruction you're describing (either an existing app that does it from BAM files, or a new app that does it by calling the GA4GH API)? If so, that would help understand if/how an API extension could help.

pgrosu:
So continuing from before, now with STAR the story is more tricky.  Even though the programming is intriguing from one perspective - STAR seems to love global variables - and for many reasons let's try to make ours a little different.  For any implementation we will do, let's try to make our code as elegant as possible while also being efficient.  I'm sure among ourselves we've seen all possible variations of coding styles, so you know what I mean.  So without further ado, let's jump into the code:

The first thing to do is start here:

https://github.com/genome-vendor/star/blob/e1a11f890b3c72cb2bf380c74f652b3213acc1cd/Stats.cpp#L21-32

https://github.com/genome-vendor/star/blob/e1a11f890b3c72cb2bf380c74f652b3213acc1cd/Stats.cpp#L125-127

```
void Stats::addStats(Stats &S) {//add S to Stats
    readN += S.readN; readBases += S.readBases;
    ...
    mappedReadsU += S.mappedReadsU; mappedReadsM += S.mappedReadsM;
	...
}

void Stats::reportFinal(ofstream &streamOut, Parameters *P) {    	
 ...
    <<setw(w1)<< "MULTI-MAPPING READS:\n" \
               <<setw(w1)<< "Number of reads mapped to multiple loci |\t"      << mappedReadsM <<"\n" \
               <<setw(w1)<< "% of reads mapped to multiple loci |\t"           << (readN>0 ? double(mappedReadsM)/double(readN)*100 : 0)<<'%' <<"\n" \
 ...			   
}
```

Thus based on this we need to locate where the `Stats` are stored for reads to lead to incrementing `mappedReadsM`:

https://github.com/genome-vendor/star/blob/e1a11f890b3c72cb2bf380c74f652b3213acc1cd/ReadAlign_outputAlignments.cpp#L63-66

```
	    if (outFilterPassed) {
            if (nTr>1) {//multimappers
                statsRA.mappedReadsM++;
                unmapType=-1;
            } else if (nTr==1) {//unique mappers
```

Next we need to find where `nTr` get's increased and this can be found here:

https://github.com/genome-vendor/star/blob/e1a11f890b3c72cb2bf380c74f652b3213acc1cd/ReadAlign_multMapSelect.cpp#L24-48

```
...
nTr=0;
    for (uint iW=0; iW<nW; iW++) {//scan windows
        for (uint iTr=0; iTr<nWinTr[iW]; iTr++) {//scan transcripts
            if ( ( (trAll[iW][iTr]->maxScore + P->outFilterMultimapScoreRange) >= maxScore ) || \
                 ( chimRecord && trAll[iW][iTr]->iFrag>=0 && (trAll[iW][iTr]->maxScore + P->outFilterMultimapScoreRange) >=maxScoreMate[trAll[iW][iTr]->iFrag] )  ) {//record this alignment
                // if paired-end, record alignments from ALL windows
                if (nTr==MAX_N_MULTMAP) {//too many alignments for this read, do not record it
                    ostringstream errOut;
                    errOut  << "EXITING: Fatal ERROR: number of alignments exceeds MAX_N_MULTMAP, increase it and re-compile STAR";
                    exitWithError(errOut.str(), std::cerr, P->inOut->logMain, EXIT_CODE_PARAMETER, *P);                    
                };
                    
                trMultScores[nTr]=trAll[iW][iTr]->maxScore;
                trMult[nTr]=trAll[iW][iTr];
                trMult[nTr]->Chr = trAll[iW][0]->Chr;
                trMult[nTr]->Str = trAll[iW][0]->Str;                  
                trMult[nTr]->roStr = trAll[iW][0]->roStr;                                 
                if ( P->outSAMprimaryFlag=="AllBestScore" && trMultScores[nTr] == maxScore ) trMult[nTr]->primaryFlag=true; 
                
                if ( (trAll[iW][iTr]->maxScore + P->outFilterMultimapScoreRange) >= maxScore) nTrMate++;
                
                nTr++;           
            };
        };
    };
...
```

Then noticing `trMult` is probably what must be key, it then can be found here:

https://github.com/genome-vendor/star/blob/e1a11f890b3c72cb2bf380c74f652b3213acc1cd/ReadAlign.h#L95-96

```
...
       intScore trMultScores[MAX_N_MULTMAP];//scores for the multiple mappers
       Transcript* trMult[MAX_N_MULTMAP];//multimapping transcripts
...	   
```

So yes, this is another approach but ours is will be targetting a much bigger scale, which would require architecting it a little differently.

Hope it helps,
Paul

pgrosu:
@dglazer, if space is an issue for indexing why not just optimize using coding schemes like Rice coding, Golomb coding, Gamma, Delta or Varint encoding as necessary for the task at hand (disk and/or wire optimizations)?  This is just a short list of coding schemes.  If we want to group multiple values, then we can use Group Varint encoding.  If speed is an issue then we have replicated index shards with balancers.  Some highly-accessed parts of the index can stay in memory and some on disk, so this can even be dynamically adjusted - which has parallels to the Memtable/SSTable approach.  Again many possibilities.

PR #267, opened 2015-03-26
Comments on this datamodel

keesvanbochove:
Hi folks, this is my first commit to this repository, and I'm not even (anymore) a developer. I do have an open source development team, though.

I'm very interested to get involved in the Metadata Task Team and work with my team on whatever useful contributions we can make. We currently have data models in 2 open source software projects that we are involved in that we would like to try to harmonize to the Metadata model:
- **Phenotype Database** (see for base model https://github.com/PhenotypeFoundation/dbxpBase/tree/master/grails-app/domain/dbnp/studycapturing, it's surprisingly similar to the Metadata Avro schema)
- **tranSMART / i2b2** (base model is a bit too much to describe here, but see https://github.com/transmart)

The website suggests that I should get in touch via GitHub, but is there also a call I could join?

Warm regards,

Kees van Bochove

### DIFF
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -65,6 +65,7 @@ record Individual {
   [NCBITaxon Ontology](http://www.obofoundry.org/wiki/index.php/NCBITaxon:Main_Page).
   For example, 'Homo sapiens' has the ID 9606. The NCBITaxon ontology ID for this is
   NCBITaxon:9606, which has the URI http://purl.obolibrary.org/obo/NCBITaxon_9606
+  Use `null` for unknown, undetermined or not applicable.
   */
   union { null, OntologyTerm } species = null;
 
@@ -147,6 +148,8 @@ record Sample {
   /**
   The time at which this sample was taken from the individual, in milliseconds
   from the epoch.
+  TODO: should we consider using relative time (e.g. to the start of a study) rather than absolute time?
+  Something like https://github.com/PhenotypeFoundation/GDT/blob/master/src/groovy/org/dbnp/gdt/RelTime.groovy
   */
   union { null, long } samplingDate = null;
 


delagoya:
-1 on the relative time. 

Also, we should probably be specific about which epoch we are referring to http://en.wikipedia.org/wiki/Epoch_(reference_date)#Notable_epoch_dates_in_computing

src/main/resources/avro/metadata.avdl
```diff
@@ -147,6 +148,8 @@ record Sample {
   /**
   The time at which this sample was taken from the individual, in milliseconds
   from the epoch.
+  TODO: should we consider using relative time (e.g. to the start of a study) rather than absolute time?
```
delagoya: -1 

I am against relative timestamps. This would complicate the gathering of metadata, and force a relationship between fields that don't really need to know about each other. In this case in particular, you would need to know the date of a whole other record, `Experiment`, to be able to interpret the date correctly.

diekhans: Relative timestamps would be a source of error.  Additionally,
GMT should be required.  Timezone should only be a display issue.

Additionally, Avro 1.8 will include a data/time type:
              https://issues.apache.org/jira/browse/AVRO-739

mbaudis:
+1 to @diekhans. There is an off-branch metadata developer version at

https://github.com/ga4gh/metadata-team/blob/master/avro-playground/metadata_redo.avdl

... which uses timestamp-millis

for most of the time attributes. Will see to move this into a PR.

src/main/resources/avro/metadata.avdl
keesvanbochove: The reason I proposed to use a relative time is that in practice, you almost never have exact dates/times available in a research context. So this field will be left null, or people will try to encode relative times in absolute times, so that you get workarounds such as dates around Jan 1, 1900, which is not desirable. Only in a medical context (hospital) you can expect actual sample dates to be available.
But it's always difficult to balance theoretical correctness and tidiness of the data model vs applicability in practice, since I'm new to the group I'm not sure what reasoning should be applied.

skeenan:
where are we with this?

mbaudis:
@skeenan -1 to relative timestamps. But +1 to develop a general way to represent relative times/intervals, e.g. for time series, clinical followup (that is length + start).

delagoya:
Agree with @mbaudis -1 to relative time stamps. 

With respect to a general method for representing time stamps, I think the best way is to close this issue, and create a new issue about a general way to represent time stamps across the entire schema. That issue should identify all of the places in the schema that need adjustment.

skeenan:
Closed.

PR #268, opened 2015-03-27
Genotype to phenotype

jnguyenx:
This protocol defines the associations between genotype and phenotype (G2P).  Associations can be made as a result of literature curation, computational modeling, inference, etc., and modeled and shared using this schema.

Here, we follow the dogma of:

      Genotype + Environment = Phenotype

where a G2P association is between the G(enotype) in the context of some E(environment), which gives rise to a P(henotype). These associations have further evidence, provenance, and attribution.

We leverage the GenomicFeature in the sequenceAnnotation schema here as it can accomodate any genomic feature from a single nucleotide variation (SNV), up through a gene, and/or complex rearrangements.  Each can be modeled as genomic features, and generally linked to a phenotype.
Collections of these features can represent a genotype at different levels of completeness.  Therefore, we can represent single allelic variation, allelic complement, and multiple variants in a genotype that can each or collectively be associated with a phenotype.

To enable standardized integration, this schema relies heavily on OntologyTerms, both for typing phenotype (for example, using the HPO or DO) but also genomic features (such as using SO) and levels of evidence (ECO).


src/main/resources/avro/common.avdl
```diff
@@ -353,6 +353,70 @@ record Sequence {
 }
 
 /**
+<<<<<<< HEAD
```
jeromekelleher: This should be deleted (left over from a merge conflict?).

jnguyenx: Thanks for catching that, it's fixed now.

src/main/resources/avro/common.avdl
```diff
@@ -353,6 +353,69 @@ record Sequence {
...
+  The source of the identifier.
+  (e.g. `Ensembl`)
+  */
+  string database;
+
+  /**
+  The ID defined by the external database.
+  (e.g. `ENST00000000000`)
+  */
+  string identifier;
```
malachig: Minor comment,

'OntologyTerm' above has an identifier entity called `id`, here in 'ExternalIdentifier' we have an identifier entity called `identifier`.  Should we make little things like this consistent?

src/main/resources/avro/genotypephenotype.avdl
```diff
@@ -0,0 +1,138 @@
...
+We leverage the GenomicFeature in the sequenceAnnotation schema here
+as it can accomodate any genomic feature from a single nucleotide variation
+(SNV), up through a gene, and/or complex rearrangements.  Each can
+be modeled as genomic features, and generally linked to a phenotype.
+Collections of these features can represent a genotype at different levels
+of completeness.  Therefore, we can represent single allelic variation,
+allelic complement, and multiple variants in a genotype that can each or
+collectively be associated with a phenotype.
+
+To enable standardized integration, this schema relies heavily on
```
malachig: There is a fair amount of jargon and abbreviation here in the docs that might not be immediately known to everyone.  Should we define each of these abbreviations at each use (and possibly include citations or URLs to the related resource?): `HPO`, `DO`, `SO`, `ECO`.

src/main/resources/avro/common.avdl
adamnovak: +1 on making these consistent.

I think `id` ought to be reserved for the fields that the implementation is going to use as primary keys or whatever for the various objects. In particular I am thinking that some implementations may want to restrict the `id`s of all objects to be decimal integer strings. Maybe `OntologyTerm` should have a `termId` and `ExternalIdentifier` should have an `externalId`?

I am thinking that these are value types, to be stored inline, and so don't really need `id` fields of their own.

src/main/resources/avro/genotypephenotype.avdl
adamnovak: Yeah, definitions or a glossary at the top would be super helpful. I have no idea what these things are.

src/main/resources/avro/genotypephenotype.avdl
```diff
@@ -0,0 +1,138 @@
...
+The context in which a genotype gives rise to a phenotype.
+This is fairly open-ended; as a stub we have a simple ontology term.
+For example, a controlled term for a drug
+*/
+record EnvironmentalContext {
+
+  /** TODO: define potential ontologies */
+  org.ga4gh.models.OntologyTerm environmentType;
+
+  /** A textual description of the environment. This is used to complement the structured description in the environmentType field */
```
adamnovak: This line is too long. In the Github diff view it is cut off at the right around "in the enviro", and the horizontal scroll bar is all the way at the bottom of the view for the file.

src/main/resources/avro/genotypephenotype.avdl
```diff
@@ -0,0 +1,138 @@
...
+*/
+record PhenotypeInstance {
+  /** The Phenotype ID. */
+  union { null, string } id;
+
+  /** HPO is recommended */
+  org.ga4gh.models.OntologyTerm type;
+
+  /** HPO is recommended.  Often this qualifier might be for abnormal/normal, or severity. */
+  union { null, org.ga4gh.models.OntologyTerm } qualifier = null;
```
adamnovak: Would anyone ever want to use multiple qualifiers?

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,115 @@
...
+This protocol defines genotype to phenotype relationships
+query system.
+
+The goal here is to allow users to query systems using one
+or more of Genotype, Phenotype, Environment, and Evidence.
+
+A query using one of the above items is to mean, by default,
+that the remainder of the query is as a "wildcard", such
+that all matches to just that query term would come back.
+Combinations of the above are to act like AND rather than OR.
```
adamnovak: Since each `FeaturePhenotypeAssociation` can have only one `evidence` specified, and multiple `EvidenceQuery`s are ANDed together, does that mean if I specify two types of evidence in my search I will get no results back (since no `FeaturePhenotypeAssociation` can have both the types of evidence I asked for)?

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,115 @@
+
+@namespace("org.ga4gh")
+/**
+This protocol defines genotype to phenotype relationships
+query system.
+
+The goal here is to allow users to query systems using one
+or more of Genotype, Phenotype, Environment, and Evidence.
```
adamnovak: Maybe this comment should be moved/echoed down where the query object is defined? So it shows up in the avrodoc for the query object?

src/main/resources/avro/methods.avdl
```diff
@@ -2,6 +2,35 @@
 
 protocol RPC {
 
+/** A query range suitable for inclusion in a positional range request. */
```
adamnovak: Do we only want to make queries on `Reference`s? Or do we want to also allow for using this on `Sequence`s that are introduced for new variation not in a `ReferenceSet`?

Why do we have both this and `Region` in `common.avdl`?

src/main/resources/avro/sequenceAnnotationmethods.avdl
```diff
@@ -0,0 +1,96 @@
...
+    */
+    array<string> parentIds = [];
+
+    /**
+    If specified, this query matches only annotations which overlap this range.
+    */
+    union { null, QueryRange } range = null;
+
+    // TODO: Fix this field. Clarify semantics around how OntologyTerm matching works, or
+    // change the Ontology term field on the feature.
```
adamnovak: Has this been done yet? Should this be done before merging?

src/main/resources/avro/sequenceAnnotations.avdl
```diff
@@ -0,0 +1,145 @@
...
+
+  /**
+  Type defining a collection of attributes associated with various protocol
+  records.  Each attribute is a name that maps to an array of one or more
+  values.  To work around Avro restrictions, three fields are define, one for
+  each type of value.  However attribute names must be unique across all three
+  fields.  If possible, multi-part string values should be split into array
+  elements instead of adding separator syntax that need to parsed.
+  */
+  record Attributes {
```
adamnovak: Since this is defining `org.ga4gh.models.Attributes`, shouldn't it be placed in `common.avdl` and used to replace all the `map<array<string>>` catch-all fields over in `variants.avdl`?

Or should this object maybe be `SequenceAnnotationAttributes` or something else more specific?

src/main/resources/avro/sequenceAnnotations.avdl
```diff
@@ -0,0 +1,145 @@
...
+  record WiggleSet {
+    /** Id of this annotation node. */
+    string id;
+
+    /** Set of additional attributes */
+    Attributes attributes;
+  }
+
+  /**
+  Node in the annotation graph that annotates a contiguous region of
```
adamnovak: Region of what?

src/main/resources/avro/common.avdl
```diff
@@ -353,6 +353,69 @@ record Sequence {
 }
 
 /**
+An abstraction for referring to a genomic region, in relation to some already 
+known reference. This will require some significant rework as we move to graph
+coordinates.  
+*/
+// TODO: Add support for paths through the graph.
```
adamnovak: Maybe this should be done now? This really has to be able to represent any contiguous subgraph.

Otherwise, `GenomicFeature` in `sequenceAnnotations.avdl` can't handle some important cases. For example, a genomic element may start on one `Sequence` and ends on a different one. Or a genomic element may be polymorphic (for example, a gene with a SNP), and both versions of the polymorphism need to be annotated as part of the element.

src/main/resources/avro/wip/variationReference.avdl
```diff
@@ -0,0 +1,263 @@
+@namespace("org.ga4gh.wip")
```
adamnovak: This whole file looks like it's getting added back incorrectly. I removed it when the graph reference stuff was accepted.

src/main/resources/avro/genotypephenotype.avdl
mbaudis: > > +      Genotype + Environment = Phenotype
Also, non-informative statements like this don't help & lead to unnecessary discussions (btw., "physical laws => everything").

And regarding this statement, the problem is not that it isn't true (it is - what parameters could be outside of the "environment", taking it to its extreme?); However, the problem lies on the right side of the equation: What is your granularity when describing a phenotype? Mean arterial blood pressure or left ventricular hypertrophy? Is there anywhere a clear definition separating observation | phenotype? And then phenotype | disorder | disease, in a systematic way? Just being curious here.

nlwashington: good idea @malachig, will make sure all abbreviations are defined.

src/main/resources/avro/wip/variationReference.avdl
jnguyenx: Hmm so should I remove it?

Looking at the graph branch, this file exists.
https://github.com/ga4gh/schemas/blob/graph/src/main/resources/avro/wip/variationReference.avdl

adamnovak: We abandoned using the graph branch, and actually did the graph stuff as a direct pull request against master, since I found a way to make it backward-compatible.

We probably should make an issue to close that topic branch.

src/main/resources/avro/genotypephenotype.avdl
adamnovak: I actually found the "Genotype + Environment = Phenotype" line helpful. That's the paradigm that the data model is modeling, simply expressed.

src/main/resources/avro/wip/variationReference.avdl
jnguyenx: I removed the file. Thanks for spotting that.

src/main/resources/avro/genotypephenotype.avdl
heckerma: > Genotype + Environment = Phenotype
What about epigenetics?

adamnovak: I guess that's environment here? It's not really mentioned.

src/main/resources/avro/common.avdl
jnguyenx: Good point, but I feel that this is something that needs to be done on master directly. It is a bit out of the scope of the pull request.

src/main/resources/avro/methods.avdl
jnguyenx: This should be asked on master, again I feel that this is out of scope for this pull request.

src/main/resources/avro/sequenceAnnotationmethods.avdl
jnguyenx: I will merge the PR with master, since sequence-annotation #246 has been merged into master.

src/main/resources/avro/sequenceAnnotations.avdl
jnguyenx: Same as above.

src/main/resources/avro/sequenceAnnotations.avdl
jnguyenx: Same as above.

src/main/resources/avro/common.avdl
jnguyenx: Same as above.

src/main/resources/avro/genotypephenotype.avdl
nlwashington: @heckerma good question.  I had not really considered epigenetic factors yet in the G2P schema, as I hadn't heard many folks discussing this with the GA4GH context yet (pardon me if i missed those conversations).  The sequence ontology captures different types of epigenetic markers as subclasses of [epigenetically_modified_region](http://sequenceontology.org/browser/current_svn/term/SO:0001720), so it is not difficult to see how those markers would fit into the <code>sequenceAnnotation</code> schema, and <code>GenomicFeatures</code> of those types could be added to the array of <code>GenomicFeatures</code>.  For the @monarch-initiative we have only started modeling epigenetic factors. I personally would consider epigenetic signatures as part of a genotype, as it is attached to the genome (though clearly could be altered due to external environmental perturbations).  That said, in Monarch we model genotypes to be comprised of two parts, intrinsic and extrinsic:  we include anything related to the DNA sequence as intrinsic and anything related to genome-targeting-reagents-resulting-in-altered-expression as extrinsic (such as addition of RNAi or morpholinos).  I can imagine that epigenetic markers would fit within a genotype, and probably the intrinsic genotype at that.  I will add a comment about being able to use <code>GenomicFeatures</code> for epigenetic markers, but we may want to discuss this in more detail in another git issue.

src/main/resources/avro/genotypephenotype.avdl
nlwashington: A reasonable suggestion.

src/main/resources/avro/genotypephenotypemethods.avdl
nlwashington: good idea.

src/main/resources/avro/genotypephenotypemethods.avdl
nlwashington: The current code permits only one evidence in the Association, and the Query also only allows for one Query (as opposed to an array) which would be ANDed to the Phenotype+Feature...so at the moment a user could not query with multiple Evidences.  But this suggests that we should update the code to 1) make the Evidence part of an association as an array of Evidences, and 2) that the user can then also query with an array of multiple evidences, if desired, to be consistent with the other Query types.

src/main/resources/avro/genotypephenotype.avdl
nlwashington: I have added definitions for abbreviations and suggestions for ontology usage throughout, in commit 73a72f7

src/main/resources/avro/genotypephenotype.avdl
nlwashington: I have hard-wrapped the comments now to 80 char in commit 73a72f7 (with the exception of urls... what is the guiding policy here?)

src/main/resources/avro/genotypephenotype.avdl
nlwashington: added in commit 73a72f7

src/main/resources/avro/genotypephenotypemethods.avdl
nlwashington: updated to permit a list of evidences, in commit 73a72f7

src/main/resources/avro/genotypephenotypemethods.avdl
nlwashington: done.  73a72f7

jnguyenx:
We modified the schemas according to the comments, can you consider accepting this pull request? :-)

mbaudis:
+1

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,154 @@
...
+Gets a list of `Variant` matching the search criteria.
+
+`POST /variants/search` must accept a JSON version of `SearchVariantsRequest`
+as the post body and will return a JSON version of `SearchVariantsResponse`.
+*/
+SearchFeaturesResponse searchVariants(
+  /** This request maps to the body of `POST /variants/search` as JSON. */
+  SearchFeaturesRequest request) throws org.ga4gh.methods.GAException;
+
+}
```
sarahhunt: Is this correct? It looks like a composite of variantmethods.avdl and g2p

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,154 @@
...
+Combinations of the above are to act like AND rather than OR.
+
+The "genotype" part of the query methods can be one or more
+genomic features.  Associations can be made at many
+levels of granularity (from whole genotypes down to individual
+SNVs), but users may use these methods with partial or
+inexact information.  Therefore, the query methods must be
+able to support query of some or all of the associated features.
+Furthermore, use of the relationships between genomic features
+means that when querying for a gene, any varants to that
```
sarahhunt: varants -> variants!

AAMargolin:
+1

sarahhunt:
+1 once the newlines are put back and typo fixed

Does SearchFeaturesResponse need a 'throws GAException' specified too?

@skeenan - does this need extensive voting as it is going into a topic branch rather than master?

pcingola:
+1

jnguyenx:
Since this pull request has been reviewed thoroughly, after it being accepted can we consider to merge the g2p branch into master?

skeenan:
+1

PR #269, opened 2015-03-30
Updating methods to work with side graphs.

adamnovak:
Since #250 went in, the methods have been out of sync with the graph representation. This is my suggestion for how to change the methods around to match the new side graph model.

Basically, I'm adding the necessary search methods for `Sequence`s and `Join`s, and getting rid of the special methods used to get novel sequences from a `VariantSet` (in favor of a `variantSetId` field when searching for `Sequence`s and `Join`s).

Also fixing a comment typo in common.avdl.


benedictpaten:
+1 LGTM. I think the join get method is a basic primitive we will need - there will be higher level methods, but this is still essential.

src/main/resources/avro/referencemethods.avdl
```diff
@@ -261,7 +260,69 @@ ListReferenceBasesResponse getReferenceBases(
...
 
-/****************  /sequence/{id}  *******************/
+/******************  /sequences/search  *********************/
+/**
+This request maps to the body of `POST /sequences/search` as JSON. Specifies a
+number of filters, all of which must be satisfird by each result returned.
+*/
+record SearchSequencesRequest {
+  /**
+  If not null, return only `Sequence`s that appear in the indicated
```
jeromekelleher: We should clarify that exactly one of referenceSetId or variantSetId is required to be non-null.

src/main/resources/avro/referencemethods.avdl
```diff
@@ -325,4 +386,99 @@ GetSequenceBasesResponse getSequenceBases(
     /** Additional request parameters to restrict the query. */
     GetSequenceBasesRequest request) throws GAException;
 
+/****************  /joins/search  *******************/
+/**
+This request maps to the body of `POST /joins/search` as JSON. Specifies a
+number of filters, all of which must be satisfird by each result returned.
```
jeromekelleher: typo: satisfird.

src/main/resources/avro/referencemethods.avdl
```diff
@@ -325,4 +386,99 @@ GetSequenceBasesResponse getSequenceBases(
...
+*/
+record SearchJoinsRequest {
+  /**
+  If present, return only `Join`s which belong to this reference set.
```
jeromekelleher: Again, we should clarify that exactly one of referenceSetId or variantSetId must be non-null. We should also use the same language (we use 'if present' here, and 'if non-null above').

src/main/resources/avro/referencemethods.avdl
```diff
@@ -325,4 +386,99 @@ GetSequenceBasesResponse getSequenceBases(
...
+
+  /**
+  If present, return only `Join`s which belong to this variant set.
+  */
+  union { null, string } variantSetId = null;
+
+  // TODO: Specify a proper sort of filter object instead of this. We want to be
+  // able to support a list of sequences, but also a single sequence and a
+  // range.
+
```
jeromekelleher: The semantics of this function is still obscure to me. Specifying a list of sequences seems fine, but the rest (starts, lengths, and strands)  seems very strange to me. What client operations are we trying to support here? Perhaps it would be better to break this into a number of different server API calls rather than trying to shoehorn all of this functionality into one call?

macieksmuga: @jeromekelleher the `SearchJoinsRequest` is intended to return a subset of joins present in a given (graph) reference or variant set, allowing the client to re-construct (a portion of) the graph from a list of joins and sequences. The `starts`, `lengths` and `strands` options allow the client to restrict the list of returned joins to a particular subgraph of interest (say, starting at a particular known location on a given reference sequence for a gene with several different alleles represented by a moderately complex subgraph, and limited to the length of that gene's "allele subgraph" along that reference sequence). Since their intent is to (cumulatively) restrict the scope of the search, I'm not sure how splitting these out into separate API calls would help.

src/main/resources/avro/referencemethods.avdl
adamnovak: OK, I did that.

adamnovak:
I've addressed the comments here, and added a stab at a subgraph query, where you can specify a `Position` and a radius and get all the `Sequence`s and `Join`s you can touch by going out that far.

src/main/resources/avro/referencemethods.avdl
```diff
@@ -113,21 +113,20 @@ as JSON.
...
   */
   union { null, string } referenceSetId = null;
 
   /**
-  If nonempty, return references which match any of the given `md5checksums`.
-  See `Reference::md5checksum` for details.
+  If nonempty, return only `Reference`s on `Sequence`s with IDs in the list.
   */
-  array<string> md5checksums = [];
+  union { null, string } sequenceIds = [];
```
macieksmuga: `string` or `array<string>` here?

src/main/resources/avro/referencemethods.avdl
```diff
@@ -325,4 +390,185 @@ GetSequenceBasesResponse getSequenceBases(
...
+
+  /**
+  If not null, return only `Join`s which belong to this variant set.
+  
+  If null, `referenceSetId` must not be null.
+  */
+  union { null, string } variantSetId = null;
+
+  /**
+  If not null, return only `Join`s with at least one `Side` on one of these
```
jeromekelleher: Should be "at least one `Side` on this sequence."

src/main/resources/avro/referencemethods.avdl
```diff
@@ -113,21 +113,20 @@ as JSON.
...
   /**
-  If present, return only references which belong to this reference set.
+  If not null, return only references which belong to this reference set.
   */
   union { null, string } referenceSetId = null;
 
   /**
-  If nonempty, return references which match any of the given `md5checksums`.
-  See `Reference::md5checksum` for details.
+  If nonempty, return only `Reference`s on `Sequence`s with IDs in the list.
```
jeromekelleher: Ditto @macieksmuga's question. Also, do we mean nonempty or non null?

jeromekelleher:
This is much better, thanks @adamnovak. You've done a great job in precisely stating the input requirements and semantics. I'm +1 modulo a few minor niggles.

src/main/resources/avro/referencemethods.avdl
adamnovak: This should still be array<string>, I think I mistook one object for another here. Not sure why this even builds...

src/main/resources/avro/referencemethods.avdl
adamnovak: It should be nonempty. This ought to be an array to match the other arrays in that request. I've fixed it now.

adamnovak:
OK, I've fixed all the bugs I introduced when I was fixing the other comments, and it's now rebased against master and squashed. I think this is ready to merge.

src/main/resources/avro/referencemethods.avdl
```diff
@@ -325,4 +390,185 @@ GetSequenceBasesResponse getSequenceBases(
...
+This request maps to the body of `POST /subgraph/radius` as JSON. Specifies
+a`Position` and a radius (in bases), and requests the subgraph of all `Join`s
+and `Sequence`s reachable within that number of bases from that position.
+
+Starting at the specified `Position`, and with a 0 radius denoting only that
+`Position`, walk outwards this many bases along all possible paths, traversing
+`Join`s only if necessary. All `Sequences` visited during this walk, and all
+`Join`s traversed, should be returned.
+*/
+record SubgraphRadiusRequest {
```
macieksmuga: Is there a POST method that corresponds to these Request/Response objects?

adamnovak: Yeah, the `subgraphRadius()` method below.

macieksmuga: Ah, I see... I tried generating a `_protocol_definitions.py` module from this code, and was missing the `subgraphRadius()` method from it because it (and its associated request/response objects) doesn't fit the typical `/something/search` pattern that the `process_schemas.py` relies on (will need to raise issue in server project to fix that script)

jeromekelleher: Is there a way we can make it follow the usual pattern without being too ugly?

adamnovak: I could just rename it to `searchSubgraphRadius()`. But do you think it's really a search method?

jeromekelleher: Do we need the radius bit? We could just call it `searchSubgraph` and follow the usual pattern. I don't think it's too clumsy --- we are searching for a particular subgraph, which we happen to define using the (point, radius) syntax.

adamnovak: I'm anticipating more ways to get subgraphs, but we do only have one now, so it could be `searchSubgraph()`.

macieksmuga: Sorry @adamnovak I think we need to think this through a bit more carefully: If it's not _really_ searching subgraphs (which it isn't), we shouldn't call it that.

You have a proposal to introduce the notion of a `Region` for graphs, and this method would return such a `Region` given a seed location and a radius around it. Perhaps develop this train of thought?

adamnovak: But I'm still not sure it's really a "search".

macieksmuga: @adamnovak You're right - it's more of a "get", and the idea is, and ideally we'll be able to obtain subgraphs through various operations (specifying seed and radius just being one such).

adamnovak:
Is there anything else I ought to do before someone merges this? Is there consensus around a name for the `subgraphRadius()` method that isn't `subgraphRadius()`? I'm reluctant to make it `searchSubgraph()` since I don't think it's a search, but since it's also the only way to get a subgraph right now I could change it to `getSubgraph()` if anyone prefers that.

macieksmuga:
@adamnovak I prefer `getSubgraph()`. It's the most descriptive of the available choices.

adamnovak:
OK, it's now `getSubgraph()`. I'm going to go ahead and mark this mergeable.

jeromekelleher:
I agree that `getSubgraph` is the best name, but this breaks two conventions: (1) all other getX methods are simple HTTP GETs of a single object; (2) all other POST methods with a complex search parameter use `search` as the prefix and have URL following a given pattern.

I also find the semantics of this method problematic. How do we manage the two lists (which I assume will be different lengths)? What order are the joins and sequences returned in? These details need to be nailed down.

I would suggest that you pull the problematic method out of this PR and make a separate PR for it, if you want to merge the rest quickly.

richarddurbin:
What about extractSubgraph as a name?  That gives the right indication that it is an extraction from a larger whole, not a simple get.

The lists would have different lengths.  The sequences are naturally ordered so will come in that order.  The join order should not matter.

Richard

> On 8 Apr 2015, at 09:30, Jerome Kelleher <notifications@github.com> wrote:
> 
> I agree that getSubgraph is the best name, but this breaks two conventions: (1) all other getX methods are simple HTTP GETs of a single object; (2) all other POST methods with a complex search parameter use search as the prefix and have URL following a given pattern.
> 
> I also find the semantics of this method problematic. How do we manage the two lists (which I assume will be different lengths)? What order are the joins and sequences returned in? These details need to be nailed down.
> 
> I would suggest that you pull the problematic method out of this PR and make a separate PR for it, if you want to merge the rest quickly.
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/ga4gh/schemas/pull/269#issuecomment-90841545>.
> 




-- 

jeromekelleher:
I think it might be simpler if we made two methods, `searchSubgraphJoins` and `searchSubgraphSequences`, which return pages of the objects in question. These would then precisely follow the pattern used in all other complex 'get' methods. I don't think it's stretching the imagination too much to call these searches: we're searching for the joins (or sequences) that exist in a given subgraph, defined by a position and a radius. This is no more odd than saying we are searching for the variants or reads that exist in a given genomic interval.

adamnovak:
I'm not convinced that it really is important to specify how the arrays will be filled in. The server can do whatever it wants, as long as when the client downloads and unions all the pages it gets the full result set. If we want to, we could add the constraint that neither array may be empty until all the results of that type have been returned.

Splitting it up into two methods might solve two problems at once: we wouldn't have to worry about the two-array paging semantics, and the two separate methods are much more like searches than this one is.

I don't really want to pull this method out, because I already pulled out the methods into this PR. I suppose I will if we can't get a consensus soon.

Does anyone not like the two-method solution?

richarddurbin:
+1 I am OK with the two method solution.

Richard

> On 8 Apr 2015, at 17:05, adamnovak <notifications@github.com> wrote:
> 
> I'm not convinced that it really is important to specify how the arrays will be filled in. The server can do whatever it wants, as long as when the client downloads and unions all the pages it gets the full result set. If we want to, we could add the constraint that neither array may be empty until all the results of that type have been returned.
> 
> Splitting it up into two methods might solve two problems at once: we wouldn't have to worry about the two-array paging semantics, and the two separate methods are much more like searches than this one is.
> 
> I don't really want to pull this method out, because I already pulled out the methods into this PR. I suppose I will if we can't get a consensus soon.
> 
> Does anyone not like the two-method solution?
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/ga4gh/schemas/pull/269#issuecomment-90960828>.
> 




-- 

macieksmuga:
I support @jeromekelleher's initial suggestion to split this method out from the current PR for now: Whether we go with the two-methods solution or something else, I don't think this should be a last-minute decision just to "get it in there" - it's a pretty critical method that exposes the challenges of integrating graphs into our current model.

FWIW, I'm not really in love with the split methods idea. A client would HAVE to issue both calls, and coordinate them on its end, in order to reconstruct a meaningful response. That may well work out in practice, but it means that our conventions are imposing some increasingly burdensome limitations on the kind of data we want to transfer.

benedictpaten:
Okay - I like this just the way it is now. I'm still 1+.

I don't think having two methods for getting the joins and sequences separately for a given radius query is a good idea - because in the restful API we have that potentially means doing the same query modulo a filter twice, once to get the sequences, once to get the joins. It is totally natural to get back two arrays, and the semantics of the paging are consistent with other methods we have.

I would like to see a general record type for subgraphs, and I think subgraphs should be segments + joins - but that's for another pull request.

jeromekelleher:
Two methods are better than one here I think. 

@adamnovak, we have to absolutely nail down the semantics of all calls completely for client code portability. If we leave things unspecified, you are right, the server can do what it likes. It will chose an ordering and structure for results that seems to make sense. Clients will write code against that server, and notice that, oh, the server is returning records in order X, I can use that property to make my algorithm more efficient. However, when the client then tries to run this code against another server which makes different assumptions, their code no longer works.

We want interoperability, so all methods must be fully and completely specified, including the order and structure of the results. (Or, if something is fundamentally unordered, this needs to be clearly documented.)

@macieksmuga, it's no big deal for the client to issue the two calls. From a client programmer perspective, I would like to get two iterators, on for  joins and and one for sequences. This maps naturally on to the two method, er, method. Paging happens much more naturally when we have two methods. I'm not necessarily going to consume the sequences and the joins at the same rate. So, what happens when my iterators become out of step when both lists are contained in the same paging structure? I'll have to cache the results from one list while I page ahead on the other. It's also conceivable that the client might only be interested in one or the other anyway, so why force them to get both?

@benedictpaten We're not getting back two arrays. If we were, I'd agree with you. We're getting back a set of pages over two different arrays. We're making things pointlessly complicated if we put two arrays into a page. We'd also break a pretty fundamental assumption in the reference implementation page handling code, so we'd need to make big special cases for this on both the client and server side.

So, having thought about this some more, I'm strongly in favour of the two method approach.

benedictpaten:
@jeromekelleher , @adamnovak  - I'll live with it being split if it makes the implementation that much easier.

adamnovak:
OK, here's an example with `searchJoinsSubgraph()` and `searchSequencesSubgraph()`.

I've specified that the orders of results are arbitrary, and that a given `Sequence` or `Join` may be returned multiple times (and it's up to the client to deduplicate them), because in discussion with @macieksmuga we concluded that server-side deduplication could potentially require a lot of server state (and/or a huge page token).

What do you think @jeromekelleher ? @macieksmuga ?

macieksmuga:
Could I make a compromise suggestion based on @richarddurbin's initial naming idea?

`extractSubgraphJoins()` and `extractSubgraphSequences()`

This way, the methods deliver what they promise.

richarddurbin:
I can see why you want the server to be allowed to return items from searchJoinsSubgraph() and searchSequencesSubgraph() in arbitrary order 
and multiple times, but I am not sure I agree it is a good idea.

You want this so you can do a depth first search returning items as you find them, and caching just where you are in the search.  
But it both puts more work onto the client, and also worse could become extremely inefficient in low complexity sequence, where you might traverse the 
same material many times, depending on the graph structure.

Much better would be to mark items you have already dealt with as you go, or keep a hash of them, so as not to repeat traverse them.  There is a
question about how long you keep the server state around for, but I guess that comes up for any of these search operations.

Richard

> On 9 Apr 2015, at 19:05, adamnovak <notifications@github.com> wrote:
> 
> OK, here's an example with searchJoinsSubgraph() and searchSequencesSubgraph().
> 
> I've specified that the orders of results are arbitrary, and that a given Sequence or Join may be returned multiple times (and it's up to the client to deduplicate them), because in discussion with @macieksmuga <https://github.com/macieksmuga> we concluded that server-side deduplication could potentially require a lot of server state (and/or a huge page token).
> 
> What do you think @jeromekelleher <https://github.com/jeromekelleher> ? @macieksmuga <https://github.com/macieksmuga> ?
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/ga4gh/schemas/pull/269#issuecomment-91312952>.
> 




-- 

src/main/resources/avro/referencemethods.avdl
```diff
@@ -284,6 +355,12 @@ record GetSequenceBasesRequest {
   union { null, long } end = null;
 
   /**
+  Specifies the maximum number of bases to return in a single page.
+  If unspecified, a system default will be used.
+  */
+  union { null, int } pageSize = null;
```
jeromekelleher: Sorry, I must have missed this before. When did we make a GET request that does paging? I thought the idea of the GETs was that they were simple requests that returned a single atomic object. 

In any case, this change is unrelated to the rest of the PR and should be put into it's own bugfix PR. It's not reasonable to expect people to re-review the entire PR each time there is an update.

src/main/resources/avro/referencemethods.avdl
```diff
@@ -325,4 +402,273 @@ GetSequenceBasesResponse getSequenceBases(
...
+a`Position` and a radius (in bases), and requests `Segment`s covering the
+portions of all `Sequence`s reachable within that number of bases from that
+position.
+
+Starting at the specified `Position`, and with a 0 radius denoting only that
+`Position`, walk outwards this many bases along all possible paths, traversing
+`Join`s only if necessary. All `Segment`s covering all bases visited during this
+walk should be returned.
+*/
+record SearchSegmentsSubgraphRequest {
```
jeromekelleher: Would `SearchSubgraphSegmentsRequest` be a bit more natural? Perhaps @macieksmuga's `ExtractSubgraphSegmentsRequest` would be better again...

jeromekelleher:
This is much better thanks @adamnovak. I'm coming around to @macieksmuga's point of using the `extract` prefix, as these are slightly different to the rest of the search requests. I don't have a strong opinion though, once the standard patterns are followed within the requests. 

I agree with @richarddurbin  that we should make the contract of returning each segment/join exactly once. I'd also like to see an ordering defined, but I agree this would probably be too strict for now. WRT to paging, I don't really know how we're going to do this. Up to now, we've managed to keep the reference server completely stateless by embedding enough information in the pageTokens to allow us to pick up any search where we left off. This is a great property to have when we start thinking about scale-out, because it means that requests for any page can go to any instance of the server without having to share state between instances. Hopefully we can keep this property in graph mode, but if we have to make it stateful for a prototype, then this isn't the end of the world.

src/main/resources/avro/referencemethods.avdl
adamnovak: That's probably true. I let @macieksmuga talk me into trying to fix what we thought was an oversight, but if we're going to be discussing it at all it needs to be its own issue.

macieksmuga: @jeromekelleher good point. I noticed a few places where `pageToken` hangs around without an associated `pageSize`. You're right that the latter may not need to be added in this case, but then, the former should be removed as well, right?

adamnovak:
OK, I've changed over to `extractSubgraphSequences()` at `/subgraph/sequences` and `extractSubgraphJoins()` at `/subgraph/joins`. I've also dropped the commit for paging sequence, as it's really quite off-topic here.

I have the server deduplication in here; if the reference server can't do it with a reasonable `pageToken`, I guess it will have to keep state (or return everything in one page?).

Are there any other suggestions?

adamnovak:
Hold on, I've reverted something I didn't want to and gone back to sequences instead of segments. I'll have that fixed in a bit.

ekg:
We can use both side graphs and string graphs with paths. The string graph with paths implies a a side graph against each path, so it provides a mechanism to express the relationship between sequences, annotations, and coordinate systems. However, many systems may benefit from the stability and simplicity of a side graph. So we should communicate both of them.

My point is that perhaps it makes sense to think about providing access to the more general structure as well. There is no reason to use only one. However, it makes things more complex.

One solution would be to provide an interface to querying the more general system. If we then provide one function that converts a string graphs to a side graph against a particular path, all the endpoints that return graphs can return either form.

adamnovak:
OK, I have fixed this to actually reflect what I think is the consensus: `extractSubgraphSegments()` (not `Sequences`) and `extractSubgraphJoins()` with server-side deduplication. What do you think @jeromekelleher ?

@ekg, the side graphs we are using are not required to be "against" any particular sequence; there's no requirement in the API anywhere that the graphs you produce be factored in terms of a primary path and a bunch of additions to it. So as far as I know, a string graph with annotated paths and a side graph with annotated paths are the same thing, with no semantic conversion necessary.

I think named path annotations would be useful, but I think they're out of scope for this particular PR. And I don't think on-the-fly conversions to make a path primary in the course of getting the results for a given method would be a good idea.

src/main/resources/avro/referencemethods.avdl
jeromekelleher: I'm not sure what the best thing to do is here --- we should tackle it in a separate PR though, and try to get this one merged as soon as we can.

jeromekelleher:
Looks good to me, thanks @adamnovak.

+1

ekg:
@adamnovak 

> the side graphs we are using are not required to be "against" any particular sequence

Side graphs _are_ built against a particular sequence. Their construction begins with a base sequence, then we add sides to it. Then, we can add sides to sides and so on, filling out a kind of coordinate inheritance tree where the initial base sequence is the parent of the entire graph. This provides a way of describing a stable coordinate system against an initial reference sequence.

Side graphs are attractive because they provide this stability. However, because they are optimized for describing things relative to an initial path, they do not make it easier to talk about multiple coordinate systems.

> there's no requirement in the API anywhere that the graphs you produce be factored in terms of a primary path and a bunch of additions to it

There is reference to joins and other concepts which are side graph specific: https://github.com/adamnovak/schemas/blob/1e5af2c28fef9d2846d6fceb549990ba8fb835e2/src/main/resources/avro/referencemethods.avdl#L398.

> So as far as I know, a string graph with annotated paths and a side graph with annotated paths are the same thing, with no semantic conversion necessary.

The difference is that the side graph has a primary root sequence, whereas the string graph + paths (variation graph) does not.

The whole point of the side graph is that we have one stable coordinate system. If we want to have multiple coordinate systems in the same graph, I fail to see the advantage of this novel graph concept.

benedictpaten:
This has three 1+'s, can someone outside of UCSC merge?

adamnovak:
@ekg I think you're conflating side graphs and insert graphs. A side graph is just a collection of sequences and joins; there's no concept there of one sequence being a parent of another. In the insert graphs we were using (before @lh3 and others made the case for side graphs), each sequence had at most one join on each end, and there were no free joins, leading to the sort of one-sequence-first hierarchical parent/child graph you are describing.

Either way, coordinates are stable under the addition of new stuff (sequences, and joins if available).

src/main/resources/avro/referencemethods.avdl
```diff
@@ -261,7 +260,73 @@ ListReferenceBasesResponse getReferenceBases(
   /** Additional request parameters to restrict the query. */
   ListReferenceBasesRequest request) throws GAException;
 
-/****************  /sequence/{id}  *******************/
+/******************  /sequences/search  *********************/
+/**
+This request maps to the body of `POST /sequences/search` as JSON. Specifies a
+number of filters, all of which must be satisfird by each result returned.
```
macieksmuga: satisfied

ekg:
I think I am clear on the differences. My concern is that I want to encode
multiple coordinate systems and annotations into the graph that I serialize
onto the wire. How does this happen with side graphs? Is there a way to do
so that does not make them equivalent to sequence graphs with paths (==
variation graphs)? If so, why is the variation graph model not appropriate?

richarddurbin:
This all seems fine to me.  I think we should be positive about the interrelationship of the side graph and a single stable coordinate system.
That is what we want GA4GH graphs for, to support stable exchange of information between sites.  If we want to change the coordinate
system we can do that as a mapping operation to a new graph.

Richard

adamnovak:
@macieksmuga OK, I have fixed that typo.

@ekg I'm not sure it's appropriate to be sending multiple coordinate systems and annotations over the wire in response to most of the methods. There should be one coordinate system used for, say, specifying the regions extracted in a subgraph.

I do think that the ability to name paths and to do conversions between named path coordinates and the underlying side graph coordinates will potentially be important. For example, if someone has data on the current GRCh38 alts, they will need to be able to do that conversion to lift over that data to a side graph representation of GRCh38. But the information used to for the conversion is annotation information and not integral to the structure of the side graph. We haven't yet defined any API machinery for querying or sending these named path annotations; it would be good to have a pull request for that.

A sequence graph with path annotations (which is also known as a variation graph?) is equivalent to a side graph with path annotations in which one path annotation is designated as primary. I would say that not having to designate a path as primary is better, because some assembly techniques do not naturally produce a primary path. Therefore, side graphs are slightly better here than variation graphs (if I am understanding variation graphs correctly), because side graphs do not need to have a primary path, while variation graphs do.

@ekg I am kind of confused as to what you want here. Do you want us to change from the side graph model to the variation graph model? Do you want us to add in API methods for named paths before merging in these methods?

jeromekelleher:
I think we're good to merge here, unless @ekg has some specific issues with the content of this PR.

Can you squash the commits please @adamnovak?

ekg:
@adamnovak

> A sequence graph with path annotations (which is also known as a variation graph?) is equivalent to a side graph with path annotations in which one path annotation is designated as primary.

I think it is reversed. The variation graph as I'm describing has no "primary" path as it is not necessarily constructed from an initial primary path. It can be an assembly graph. It van have cycles. It has nodes that are possibly short. Edges only connect the ends of nodes. Paths through the graph can be named. This provides a multiplicity of coordinate systems.

The side graph is not like an assembly graph. It is a DAG of joins and sequence additions applied to a base set of non overlapping sequences. These root sequences are what I am calling primary. I am sure it could be used to represent any kind of graph, but it has a design rooted in the idea that we are working from a linear, non-overlapping set of primary sequences. So the side graph has a primary sequence. I hope I have understood this correctly and if not I hope someone else can help correct my confusion.

If I'm confused then I apologize, but we don't need to be confused about this because you can see an implementation of variation graphs that does what I'm describing (vg). I am positive we would understand each other better if we were trying to get two implementations to work together and talk to each other. So to respond to @jeromekelleher, I think we should refrain from including things in the API until we have examples of them that can be inspected. Otherwise what is the point of standardization? We have one implementation and a other virtual one. If we focus this standard on the virtual one, we don't even see what we are building around.

adamnovak:
@jeromekelleher I've squashed everything down.

@ekg I think you are right and I was reversed. We do have a `Reference.isPrimary` field that defines some primary sequences, while a variation graph would have not one primary path but several named paths, right?

I think though that my original point still stands: the similarities between a side graph and a variation graph are more important than their differences. A variation graph requires joins to be at the ends of sequences, while a side graph allows joins anywhere along a sequence. A side graph has a set of primary sequences, while a variation graph has several named path annotations.

The side graph model is in right now, and this PR is just updating the methods to work with the models currently on the master branch. Changing over to a variation graph model would be relatively simple: delete `Reference.isPrimary`, constrain `Join`s to connect the ends of `Sequence`s, and add multiple named path annotations.

If we wanted to generalize our model to support servers serving either side graphs or variation graphs, all we would need to do would be to add the named paths feature, and say that some `ReferenceSet`s might contain no primary sequences. I think that that would probably be a good move, but I think you might be the right person to do that PR.

I do agree that eventually we should stop throwing unimplemented stuff into the schemas and dumping it on the reference server team to implement, but I think that they are catching up and we should wait until they are a bit more caught up to do that. Plus, this PR isn't *entirely* unimplemented: @macieksmuga has been working on the subgraph extraction methods over at <https://github.com/macieksmuga/server/tree/graphproto>.

pgrosu:
On Tue, Apr 14, 2015 at 1:29 PM, adamnovak <notifications@github.com> wrote:

> @ekg [...] I think you might be the right person to do that PR [...] I do agree that eventually we should stop throwing unimplemented stuff into the schemas and dumping it on the reference server team to implement.

+1 x 2

jeromekelleher:
Thanks @adamnovak --- since there are no -1s and a bunch of +1s I'm going to merge this.

PR #281, opened 2015-04-08
First stab at free `Allele`s, with a semantic search for them.

adamnovak:
These ideas probably won't be anywhere near ready until #253 and #269 are in. But I was asked by @benedictpaten to spec out how `Allele`s might work if they weren't restricted to a single `VariantSet`. What do people think of that idea?

Changes:

`Allele`s now exist outside of, and can be shared between,
`VariantSet`s. The parent object of an `Allele` is a `Dataset`.

Adding a `searchAllelesPath()` method that can use any of a number of
"algorithms" to get `Allele`s that are "relevant" to a `Path`, for some
algorithm-defined notion of relevance.

Also specifying that the `Allele`s used by a `VariantSet` may only use a
"simple" extension of the `Sequence`s and `Join`s in the relevant
`ReferenceSet`. That is, no novel `Join` may have both ends in novel
`Sequence`s.

### DIFF
--- a/src/main/resources/avro/variantmethods.avdl
+++ b/src/main/resources/avro/variantmethods.avdl
@@ -163,8 +163,17 @@ org.ga4gh.models.Variant getVariant(
 /******************  /alleles/search  *********************/
 /** This request maps to the body of `POST /alleles/search` as JSON. */
 record SearchAllelesRequest {
-  /** Required. The IDs of the variant sets to search over. */
+  /**
+  If nonempty, look only at `Allele`s used in at least one of the specified
+  `VariantSet`s. `Allele`s that are not used in any `VariantSet`s may only be
+  returned if this is empty.
+  */
   array<string> variantSetIds = [];
+  
+  /**
+  If present, look only at `Allele`s in the specified dataset.
+  */
+  union { null, string } datasetId = null; 
 
   /**
   Required. Only return `Allele`s on the sequence with this ID.
@@ -238,6 +247,75 @@ org.ga4gh.models.Allele getAllele(
   The ID of the `Allele`.
   */
   string id) throws GAException;
+  
+/******************  /alleles/search/path  *********************/
+/** This request maps to the body of `POST /alleles/search/path` as JSON. */
+record SearchAllelesPathRequest {
+  /**
+  If present, look only at `Allele`s in the specified `Dataset`.
+  */
+  union { null, string } datasetId = null; 
+
+  /**
+  If present, look only at `Allele`s used in the specified `VariantSet`.
+  */
+  union { null, string } variantSetId = null; 
+
+  /**
+  Specifies a search algorithm to use, which defines the exact semantics of what
+  it means for an `Allele` to be "relevant" to a `Path`. If the server does not
+  support the sp[ecified algorithm, it will throw an error.
+  */
+  string algorithm;
+
+  /**
+  Specifies a path. Only alleles "relevant" to the specified `Path` under the
+  specified algorithm will be returned.
+  */
+  org.ga4gh.models.Path query;
+
+  /**
+  Specifies the maximum number of results to return in a single page.
+  If unspecified, a system default will be used.
+  */
+  union { null, int } pageSize = null;
+
+  /**
+  The continuation token, which is used to page through large result sets.
+  To get the next page of results, set this parameter to the value of
+  `nextPageToken` from the previous response.
+  */
+  union { null, string } pageToken = null;
+}
+
+/** This is the response from `POST /alleles/search/path` expressed as JSON. */
+record SearchAllelesPathResponse {
+  /**
+  The list of relevant alleles. An `Allele` should be returned if the specified
+  search algorithm deems it relevant to the specified path.
+  */
+  array<org.ga4gh.models.Allele> alleles = [];
+
+  /**
+  The continuation token, which is used to page through large result sets.
+  Provide this value in a subsequent request to return the next page of
+  results. This field will be empty if there aren't any additional results.
+  */
+  union { null, string } nextPageToken = null;
+}
+
+/**
+Gets a list of `Allele`s relevant to a path.
+
+`POST /alleles/search/path` must accept a JSON version of
+`SearchAllelesPathRequest` as the post body and will return a JSON version of
+`SearchAllelesPathResponse`.
+*/
+SearchAllelesPathResponse searchAllelesPath(
+    /**
+    This request maps to the body of `POST /alleles/search/path` as JSON.
+    */
+    SearchAllelesPathRequest request) throws GAException;
 
 /******************  /callsets/search  *********************/
 /** This request maps to the body of `POST /callsets/search` as JSON. */
--- a/src/main/resources/avro/variantmethods.avdl
+++ b/src/main/resources/avro/variantmethods.avdl
@@ -164,16 +164,21 @@ org.ga4gh.models.Variant getVariant(
 /** This request maps to the body of `POST /alleles/search` as JSON. */
 record SearchAllelesRequest {
   /**
-  If nonempty, look only at `Allele`s used in at least one of the specified
-  `VariantSet`s. `Allele`s that are not used in any `VariantSet`s may only be
-  returned if this is empty.
+  If nonempty, look only at `Allele`s used in the specified `VariantSet`. This
+  includes both `Allele`s novel to that `VariantSet` and `Allele`s from the
+  associated `ReferenceSet` with `AlleleCall`s in that `VariantSet`.
+
+  Either this field or referenceSetId must be set.
   */
-  array<string> variantSetIds = [];
-  
+  union { null, string } variantSetId = null;
+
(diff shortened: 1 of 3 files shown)


richarddurbin:
Thanks @adamnovak.  This looks good to me.  

Does "relevant" mean overlap, or containment (allele in path)?  Or identity?

Am I right that the ReferenceSet is defined by the DataSet?

In particular I support the "simple extension" requirement.  i.e.  that novel Joins involved in defining Alleles must connect to the ReferenceSet at at least one end.

> On 9 Apr 2015, at 00:57, adamnovak <notifications@github.com> wrote:
> 
> These ideas probably won't be anywhere near ready until #253 <https://github.com/ga4gh/schemas/pull/253> and #269 <https://github.com/ga4gh/schemas/pull/269> are in. But I was asked by @benedictpaten <https://github.com/benedictpaten> to spec out how Alleles might work if they weren't restricted to a single VariantSet. What do people think of that idea?
> 
> Changes:
> 
> Alleles now exist outside of, and can be shared between,
> VariantSets. The parent object of an Allele is a Dataset.
> 
> Adding a searchAllelesPath() method that can use any of a number of
> "algorithms" to get Alleles that are "relevant" to a Path, for some
> algorithm-defined notion of relevance.
> 
> Also specifying that the Alleles used by a VariantSet may only use a
> "simple" extension of the Sequences and Joins in the relevant
> ReferenceSet. That is, no novel Join may have both ends in novel
> Sequences.
> 
> 
>   https://github.com/ga4gh/schemas/pull/281 <https://github.com/ga4gh/schemas/pull/281>
> Commit Summary
> 
> Making `Alleles` first-class objects.
> File Changes
> 
> M src/main/resources/avro/variantmethods.avdl <https://github.com/ga4gh/schemas/pull/281/files#diff-0> (80)
> M src/main/resources/avro/variants.avdl <https://github.com/ga4gh/schemas/pull/281/files#diff-1> (33)
> Patch Links:
> 
> https://github.com/ga4gh/schemas/pull/281.patch <https://github.com/ga4gh/schemas/pull/281.patch>
> https://github.com/ga4gh/schemas/pull/281.diff <https://github.com/ga4gh/schemas/pull/281.diff>
> —
> Reply to this email directly or view it on GitHub <https://github.com/ga4gh/schemas/pull/281>.
> 




-- 

adamnovak:
"Relevant" is intentionally vague, because different algorithms will have different notions of relevance. The one that is most natural to me would be to find all Alleles that either occur along the query path, or which are alternatives to sequence on the query path. However, whether that ought to include, say, alleles which involve deleting or duplicating a huge region around the query path is a hard question. I think it's never really going to be settled, and is properly outside the scope of the standard, hence the ability for many different algorithms to be available.

Ideally, someone will make one really good algorithm, the semantics of which everyone finds useful, and there can be a de facto standard around that.

src/main/resources/avro/variantmethods.avdl
```diff
@@ -320,6 +329,75 @@ org.ga4gh.models.Allele getAllele(
...
+
+  /**
+  If present, look only at `Allele`s used in the specified `VariantSet`.
+  */
+  union { null, string } variantSetId = null; 
+
+  /**
+  Specifies a search algorithm to use, which defines the exact semantics of what
+  it means for an `Allele` to be "relevant" to a `Path`. If the server does not
+  support the sp[ecified algorithm, it will throw an error.
```
macieksmuga: typo

src/main/resources/avro/variantmethods.avdl
```diff
@@ -320,6 +329,75 @@ org.ga4gh.models.Allele getAllele(
...
+  */
+  string algorithm;
```
macieksmuga: Perhaps there should be a method for the server to return a list of supported algorithms, then...

macieksmuga: Also, perhaps define it as follows:
```
union {null, string} algorithm = null;
```
to allow the server to select some default algorithm if none is specified by the user?

diekhans: Maciek Smuga-Otto <notifications@github.com> writes:
> to allow the server to select some default algorithm if none is specified by
> the user?

general comment: server should never select defaults.  It just
servers data.  Defaults in data are hugely error prone.

macieksmuga: @diekhans True, that. 
But then, the server will have to bear the full responsibility of advertising exactly what algorithms it offers, and exactly what they do with alleles along paths.

benedictpaten:
Okay - I am plus 1 on this, except for one thing. 

In the proposed pull request all alleles are part of a dataset, even alleles that use joins/sequences that are contained in a variant set (as in a simple extension). This is odd and creates some ugly corner cases. For example, you could have an allele that includes variants/joins from multiple variant sets. I think it would be better therefore to have alleles to be either part of a dataset (so called "named alleles") or part of a variant set (allele calls specific to a given variant set). I think such a distinction is quite natural. I would use the rule that if an allele uses a join/sequence that is part of a variant set, it must be part of that variant set, else it can be part of the dataset.

benedictpaten:
Hmm, more thoughts (spelling out a discussion with Adam offline).

Given that all joins and sequences belong either to a reference set or a variant set, I now think that an allele should be allowed to belong to either a reference set, or a variant set, but not to a data set, for the following reasons (generalising my earlier argument). If an allele belongs to a data set, it may include sequences/joins defined in any combination of reference sets and variant sets (a data set can contain more than one of both reference sets and variants sets). This will be really hard to manage and communicate, I sense.

Instead, I would argue we impose the following: 

(0) Reference sets should only be allowed to inherit from one parent reference set (currently, I believe, we allow for multiple inheritance - which is a mistake for obvious reasons).

(1) Alleles should not be allowed to include joins/sequences defined across multiple reference sets, except for joins/sequences defined in a child reference set and its parent, and, recursively, ancestor reference sets.

(2) Alleles should not be allowed to include joins/sequence defined across multiple variant sets.

(3) In the case that an allele includes joins/sequences defined in a variant set, it should belong to that variant set. 

(4) In the case that the allele includes joins/sequences defined only in a reference set (and its ancestors), it should belong to that reference set.

This still allows for "named alleles", on reference sets. Consider the current reference set for GRCh38 - which is composed simply of a set of primary and alt sequences. It is easy easy to name alleles for this reference set, and include them across variant sets. This plan therefore admits the status quo. It also allows for named alleles to be created on extensions of the GRCh38 graph in a natural way.

haussler:
+1

adamnovak:
OK, here it is with `Allele`s that can belong either to `VariantSet`s or `ReferenceSet`s.

I have a single `variantSetId` and a single `referenceSetId` on the `Allele` object, because I thought that that would make it clearer with the relationship was. You are allowed to use both, but you can't have an `Allele` in two `ReferenceSet`s unless it was inherited from their common ancestor. This differs from the behavior of `Reference`s and `Join`s, but I think it might be better.

Thoughts?

adamnovak:
Actually, I have decided on my own that that was a bad idea. Here's a version in which `Allele`s work just like `Sequence`s and `Join`s in terms of what they can belong to and how they are inherited. You can't put an `Allele` in a `ReferenceSet` or `VariantSet` unless that container has all the `Sequence`s and `Join`s the `Allele` uses.

src/main/resources/avro/variants.avdl
```diff
@@ -365,13 +370,18 @@ it is also common for it to be a mix of `Segment`s on reference and novel
...
-sequence graph. For example, if  an `Allele` that spans a novel deletion is
-required, a new `Join` should exist to describe that deletion, available through
-`searchJoins()`.
+To get the `Allele`s in a `ReferenceSet` or `VariantSet`, use `searchAlleles()`.
+
+Note that `Path`s cannot follow `Join`s not represented in the relevant sequence
+graph. For example, if  an `Allele` in a `VariantSet` describes a novel deletion
+relative to the applicable `ReferenceSet`, the `Join` describing that deletion
+must be a novel `Join` in that `VariantSet`, and thus available through
+`searchJoins()` on the `VariantSet`.
```
macieksmuga: Just to make it clear, an `Allele` could take a path through the `Join`s and `Sequence` representing, say, a SNP, that's described in the underlying `ReferenceSet`, thus not novel to the `Allele`'s `VariantSet`, correct?

adamnovak: Yes, and that allele could be in the ReferenceSet or in the VariantSet.

Maybe I should add the proviso that it can't be in a VariantSet if it's in that VariantSet's ReferenceSet...

macieksmuga: Thanks! The architecture of the graph reference server allows for either case (it doesn't enforce constraints on allele membership in a variantSet).

jeromekelleher:
Can this PR be closed @adamnovak?

adamnovak:
Sure. I think some of this might come back when the graph stuff does, but I don't want this merged into master at the moment.

Issue #282, opened 2015-04-09
Remove sourceURI from Reference


macieksmuga:
current Reference record has a `sourceURI` field:
```
  union { null, string } sourceURI = null;
```
that seems at cross purposes to the architecture of this web service API.

delagoya:
Closing in favor of the PR that is being discussed as part of #283

PR #283, opened 2015-04-09
Removed sourceURI from Reference record.

macieksmuga:
Clients should be able to pull all required metadata and sequence bases
using the methods defined in referencemethods.avdl.

Addresses issue #282

### DIFF
--- a/src/main/resources/avro/references.avdl
+++ b/src/main/resources/avro/references.avdl
@@ -69,15 +69,6 @@ record Reference {
   string name;
 
   /**
-  The URI from which the sequence was obtained.
-  Specifies a FASTA format file/string with one name, sequence pair.
-  In most cases, clients should call the `getSequenceBases()` or
-  `getReferenceBases()` methods to obtain sequence bases for a `Reference`
-  instead of attempting to retrieve this URI.
-  */
-  union { null, string } sourceURI = null;
-
-  /**
   All known corresponding accession IDs in INSDC (GenBank/ENA/DDBJ) ideally
   with a version number, e.g. `GCF_000001405.26`.
   */
@@ -179,9 +170,6 @@ record ReferenceSet {
 
   /** Public id of this reference set, such as `GRCh37`. */
   union { null, string } assemblyId = null;
-
-  /** Specifies a FASTA format file/string. */
-  union { null, string } sourceURI = null;
 
   /**
   All known corresponding accession IDs in INSDC (GenBank/ENA/DDBJ) ideally


macieksmuga:
(I amended the PR to remove the sourceURI from ReferenceSet as well, for the same reason).

jeromekelleher:
I definitely agree with this change, but I think we'd need to replace these fields with something. I think the main reason they exist is for provenance (I hope that no-one was suggesting we download FASTA files out-of-band), so we'd need to keep that in some way.

macieksmuga:
@jeromekelleher I agree that provenance is absolutely critical, but isn't that already being accomplished through appropriately curated metadata, such as maintained in the `sourceAccession` array and the `ncbiTaxonId`?

jeromekelleher:
I tend to agree @macieksmuga --- perhaps we could get some feedbacl from the original designers? Maybe git blame would be helpful  to track this down...

calbach:
I believe this was originally added by @richarddurbin. Though I vaguely remember the API having more `sourceURI`s at some point, so perhaps it was included for the sake of consistency.

I don't think the intent was ever to have the user attempt to download the `sourceURI` directly. The field seems mostly innocuous to me, but I guess I'm not opposed to removing it. There have been several instances where people assumed the primary access pattern for a Reference was to download the `sourceURI`, perhaps that's sufficient reason alone for removal.

Tentative +1..

richarddurbin:
It was there for provenance.  If there is somewhere else to record provenance then I am happy to remove this.

Richard

macieksmuga:
What particular metadata is required to track provenance? Is there some canonical store at NCBI or elsewhere of "known good" FASTA files, that this value was assumed to point to? The reason I'm asking this is that as we begin to assemble graph references, we may end up using quite a few data sources for (short) sequences (such as SNP databases), and even if they're all gathered into a collection of FASTA files to serve out, I'm not sure there's a guarantee that those files would properly encode the provenance of each sequence.

richarddurbin:
I don't think it is realistic to systematise where people get their sequences from, although many will come from standard places.  The idea here is to 
allow flexible indication of what arbitrary source was used, in the same sense as text in the methods section of a paper, not to support automated 
reconstruction of extraction.  You could rename it to provenanceURI, with a comment that this is not for automated retrieval of a sequence.  Or you 
can remove it, and people can record this sort of information in accompanying README files. This is not a high priority for us to construct a formal
solution for - it does not affect functional use of the system.

Richard

delagoya:
It seems we have a consensus, but I am not at all sure what that consensus is, so let me restate what I think it is and see if we can come to closure:

- There is a recognition that the source of a sequence is not always a standard repository, and that an optional field to designate a non-repository source is a Good Thing To Have. 
- The name `sourceURI` is an acceptable name, and we should not change it for change's sake unless there is a strong opinion to do so.
- A possible alternative PR would be to update the documentation to provide proper context for use of `sourceURI`

macieksmuga:
@richarddurbin thanks for clarifying the intended uses of this field! @delagoya based on the feedback here, the course you suggest seems reasonable. Unless I hear otherwise over the course of the next day, I'll close this PR and open a new one with documentation fixes as described above.

macieksmuga:
Closing this PR for reasons stated in discussion.

PR #284, opened 2015-04-14
Refine comments to Unix epoch

delagoya:
Several files have timestamp fields defined as "The time at which this RECORD_TYPE was last updated in milliseconds from the epoch."

This PR refines the definition of the "epoch" to the [Unix time epoch](http://en.wikipedia.org/wiki/Unix_time)


delagoya:
Request for input. This one should be non-controversial, as it is just a documentation refinement. But since it touches three working groups I would like to have input from someone in each.

src/main/resources/avro/metadata.avdl
```diff
@@ -27,13 +27,13 @@ record Individual {
   union { null, string } description = null;
 
   /**
-  The time at which this individual was created in milliseconds from the epoch.
+  The time at which this individual was created in milliseconds from the [Unix time epoch](http://en.wikipedia.org/wiki/Unix_time).
```
jeromekelleher: Are these lines < 80 chars?

jeromekelleher:
I still think an [ISO date](http://en.wikipedia.org/wiki/ISO_8601) would be better, but this is definitely an improvement.

+1 after fixing the line lengths.

mbaudis:
The "playground" metadata version defines all times in ISO 8601. However, I'm not sure if this works in the current AVRO version used here?

```
  /**
  The time at which this individual's record was created. 
  Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
  */
  union { null, timestamp-millis } recordCreated = null;
```

diekhans:
We should use ISO dates and times and GMT.  This is far
more common in the data exchange world.

We can do that as a normalization pass over the whole
schema.  Avro 1.8 will have a data/time type.

delagoya:
@jeromekelleher good catch, reformatted just now.

RE: ISO date format, avro does not currently offer a basic type for dates and timestamps, but it appears that version 1.8.0 will support it: https://issues.apache.org/jira/browse/AVRO-739

I say punt until it becomes an officially supported data type.

jeromekelleher:
Thanks @delagoya, looks good. Assuming others are OK with this change, can you squash the commits please?

I agree on waiting until Avro 1.8 before we revisit the timestamp format issue.

mbaudis:
+1

For the next larger metadata PR, we'll either use your updated version + an AVRO 1.8 => ISO comment, or go straight ISO - if it takes this long ...

delagoya:
@jeromekelleher my Git-fu is not all that great, but I rebased the ga4gh/schema#master then push out the commit. Will that suffice?

jeromekelleher:
@delagoya --- the easiest thing to do is to `git rebase -i upstream/master`, and then 'squash' all the commits except the first when the editor window pops up. Once that's done, use git push -f origin/<branchname> to force push the changes.

delagoya:
Folks I am going to close this PR in favor of a new one, considering that PR #301 may conflict. Once that issue is addressed, will resubmit a clean pull request. 

Plus I had a Rebase Gone Wrong just now and don't have time to debug ;-)

PR #286, opened 2015-04-16
Removed outdated definition of Experiment from reads.avdl

macieksmuga:
`reads.avdl` contained a redundant and outdated copy of the `Experiment` record (now contained in `metadata.avdl`, that was causing the server schema tests to fail on post-v0.5.1 schema definitions.


diekhans:
+1  The experiment object was in reads just because the metadata wasn't done

mbaudis:
+1

benedictpaten:
+1

PR #291, opened 2015-04-23
Better CI for the schemas repo

dcolligan:
The point of this change is to have more stringent CI on schema repo checkins so we don't encounter problems downstream parsing and using the schemas in the server repo.

This will be accomplished by copying -- in a modified form -- some of the code that we have developed in the server repo for manipulating avro schemas into the schemas repo.  We can add additional checks with python scripts as we uncover issues.

This change is discussed in the server repo here: https://github.com/ga4gh/server/issues/366

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.py[cod]
 target
 *~
 #*
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,1 +1,12 @@
-language: java
+language: python
+python:
+  - "2.7"
+
+install:
+    - pip install -r requirements.txt
+
+script:
+    - flake8 scripts tests
+    - nosetests tests
+    - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
+    - mvn test -B
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,5 @@
+PyYAML
+flake8
+humanize
+nose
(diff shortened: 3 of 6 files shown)


dcolligan:
Actually, this is ready for review now.  This commit merely establishes the framework change (all of the old tests are still run).  Subsequent PRs will add new tests along with the fixes for the schema issues which the tests identify as needing fixing.  Proceeding in this way both allows incremental progress and not breaking the CI at any point.

jeromekelleher:
I like this idea --- we really do need stronger CI on the schemas, and this is a way of getting there. How do people feel about bringing Python code into the schemas repo?

macieksmuga:
This will go a long way towards ensuring that the schemas remain usable in real-world applications (such as the reference server) as they evolve. 

Enthusiastic +1 on the concept - but you may want to include basic documentation (including links to `pip` and such) on how to run these tests for people unfamiliar with Python development.

I also noted a lot of extra chaff output when running the test harness on my mac, almost all of it generated by a highly verbose maven.

fnothaft:
Strong +1 as well.

jeromekelleher:
Looks like people are happy with including Python code: excellent!

+1.

jeromekelleher:
OK, we have enough +1s and no-one has complained about this, so I'm going to merge. Thanks @dcolligan.

PR #292, opened 2015-04-27
Change ternary SAM flags to binary flags

dcolligan:
See discussion that originated this PR in https://github.com/ga4gh/server/issues/362

### DIFF
--- a/src/main/resources/avro/reads.avdl
+++ b/src/main/resources/avro/reads.avdl
@@ -243,10 +243,10 @@ record ReadAlignment {
   The orientation and the distance between reads from the fragment are
   consistent with the sequencing protocol (equivalent to SAM flag 0x2)
   */
-  union { boolean, null } properPlacement = false;
+  union { null, boolean } properPlacement = null;
 
   /** The fragment is a PCR or optical duplicate (SAM flag 0x400) */
-  union { boolean, null } duplicateFragment = false;
+  union { null, boolean } duplicateFragment = null;
 
   /** The number of reads in the fragment (extension to SAM flag 0x1) */
   union { null, int } numberReads = null;
@@ -263,7 +263,7 @@ record ReadAlignment {
   union { null, int } readNumber = null;
 
   /** SAM flag 0x200 */
-  union { boolean, null } failedVendorQualityChecks = false;
+  union { null, boolean } failedVendorQualityChecks = null;
 
   /**
   The alignment for this alignment record. This field will be
@@ -284,7 +284,7 @@ record ReadAlignment {
   By convention, each read has one and only one alignment where both
   secondaryAlignment and supplementaryAlignment are false.
   */
-  union { boolean, null } secondaryAlignment = false;
+  union { null, boolean } secondaryAlignment = null;
 
   /**
   Whether this alignment is supplementary. Equivalent to SAM flag 0x800.
@@ -300,7 +300,7 @@ record ReadAlignment {
   `alignedSequence` and `alignedQuality` fields in the alignment record will
   only represent the bases for its respective alignment.
   */
-  union { boolean, null } supplementaryAlignment = false;
+  union { null, boolean } supplementaryAlignment = null;
 
   /**
   The bases of the read sequence contained in this alignment record.


fnothaft:
+1

jeromekelleher:
Looks good to me, but I don't understand the data well enough to have an opinion. @lh3, is there a reason why we might want to make these fields nullable?

lh3:
These are not my design, but I think most of them make sense. For example:

* `properPlacement`: not applicable to single-end
* `duplicatedFragment`: what if markDuplicate is not applied (e.g. PCR-free)?
* `failedVendorQualityChecks`: information not available?

In addition, when these fields are not null, you have to fill them for every ReadAlignment record, which is clumsy and inefficient. There's also backward compatibility issue. Overall -1.

fnothaft:
@lh3 I agree about `duplicatedFragment` but disagree about `properPlacement` . E.g., if a read is single ended, there is no placement invariant for it to violate, which would reasonably imply that the read is properly placed. Additionally, since default values are already given, you only need to set the fields if you know that you might not match the defaults (e.g., you know you might not match the default for `properPlacement` if you're mapping a paired end read). I'm not familiar enough with the `failedVendorQualityChecks` field to comment intelligibly there.

In general, unless there is a need for a field to be ternary, I would prefer the fields to be binary. Else, more work has to be done to validate the ternary fields when reading the records.

lh3:
Almost no one uses `failedVendorQualityChecks` nowadays. Then you will fill every record with `failedVendorQualityChecks=false`, which is unnecessary. In SAM, `properPlacement` is for paired-end data only.

jeromekelleher:
@dcolligan --- I think this is mostly motivated by issues encountered when processing the schemas, isn't it? From what @lh3  says, we'll need to find a different way to fix (at lease some of) these problems. What are the alternatives?

macieksmuga:
With all due respect, what's wrong with filling every field with `failedVendorQualityChecks=false` as opposed to filling the field with `null`? (it takes the same amount of space and effort, and complicates the model significantly - thus this PR) Isn't that flag set to false (by default) on every original SAM file record to begin with, thus making the filling of this field automatic when a record is read? Or am I missing something about SAM binary flags?

fnothaft:
+1 @macieksmuga

lh3:
As I understand, when a field is null, you don't need to output it in JSON.

jeromekelleher:
What about the future though @macieksmuga? Perhaps we'll come to a point where read data isn't going through the SAM/BAM format at all, and setting these flags wouldn't make sense. Making these nullable makes sense from a forwards compatibility perspective if we intend to eventually deprecate them (like `failedVendorQualityChecks`?).

dcolligan:
@jeromekelleher we have two options:
1) change the way `process_schemas.py` parses schema files
2) change the schemas

This PR is one way of doing (2).  Another way of doing (2) would be to recode the SAM flags like:
```
union { null, boolean } supplementaryAlignment = null;
```
That is, switching the order of the types and defaulting to null instead of false.

jeromekelleher:
@dcolligan The `union { null, boolean } supplementaryAlignment = null;` option seems perfectly reasonable to me...

macieksmuga:
@dcolligan +1 on your suggestion to flip null and boolean in this case.

lh3:
Yes, `supplementaryAlignment` etc should default to `null` actually (meaning not available), not to `false`.

lh3:
> Yes, supplementaryAlignment etc should default to null actually (meaning not available), not to false.

Hmm.. probably need more people to comment whether this is the right way.

diekhans:
The core problem here is that the motivation for allow this
field to be NULL was not documented.  This left us to guess if
it was a well thought out decision or a copy and paste of some
other field.  It also left no guidance to those wanting to use
the field.

I am not pointing fingers here; I do this too and have create a
lot few lines of schema than most.  However, it's vital
important to capture the `why' in the schemas.

We need the GA4GH documentation to be normative.  We need
developers to be able to accurately use or implement the schemas
base only on documentation and code in the repositories.

macieksmuga:
@diekhans well said!

I gather that changing the union order and setting default to null is what's called for here, then?
If so, could this be changed/accepted ASAP, as it's just as important as #293 for server development to proceed.

diekhans:
If the reason for NULL is to have three states:
  success, fail, not-yet-tested

then a default to false is deceptive

It might be clearer to represent these values as an enum rather
that a null+boolean.

pgrosu:
@dcolligan, @jeromekelleher, @macieksmuga It might be a good idea to update `process_schemas.py` to ensure that it is Avro-compatible.

@lh3 I might be better to not consider `null` as ignorable, but maybe more along the lines of:

1) Not yet tested as @diekhans mentioned, or
2) The confidence in this setting is not high for `true` or `false`, and might need retesting.

If we ignore it, it could fail the test of the following definition if things were not properly set:

```
By convention, each read has one and only one alignment where both
  secondaryAlignment and supplementaryAlignment are false.
```

+1 @diekhans I think `enum` might be appropriate here, though `true/false` might probably work better for some things instead of `success(pass)`/`fail` than others.  Maybe updating the variable names to  `IsA_secondaryAlignment` and `IsA_supplementaryAlignment` might help, though we can think more about it so they would communicate the intent as clearly as possible.  I also agree that the documentation should be normative.

jeromekelleher:
@dcolligan, could you update this PR (or open a new one) with the proposed defaulting to null behaviour? I think a null default is by far the most appropriate here. With a null default, if a server doesn't set the value, it is returned as null (which is correct). With a False default if a server doesn't set the value (for whatever reason), we return False which is giving the client incorrect information.

I think {null, boolean} is perfectly reasonable, for the reasons @lh3 gives: we have three values, True, False, or not set/not available.

In practise, these default values don't really mean much, since these are object that the server is filling out, and it's really up to the server what it returns. The only place the default values really matter is in the SearchRequest classes, where we are defining the default behaviour of the server in the cases in which the client does not provide full information.

diekhans:
@dcolligan please include documentation on the meaning of NULL.  I can
help if needed.

dcolligan:
Okay, just updated the PR to go with the "change default of ternary flags to null" implementation.

Once this PR is merged `process_schemas.py` runs correctly, which means we can add a modified version of it to the schemas CI.

jeromekelleher:
Great, thanks @dcolligan. We should file an issue on the lack of documentation here and fix in a follow-up.

+1.

dcolligan:
@lh3 you were the only one that gave a -1 to the previous design.  Do you approve of this new design?

macieksmuga:
Since the nature of the PR changed significantly, here's a shiny new +1 from me.

fnothaft:
+1 as well.

dcolligan:
Okay, since nobody seems to be objecting to the new approach, I think this is good to merge.

jeromekelleher:
Technically, this should go through the new approval process in #296. @delagoya, can you advise please?

delagoya:
-1 

As per @jeromekelleher, this represents an major API Change and the latest commit needs to be cleared by the following people before it can move forward:

@jeromekelleher (done)
@fnothaft (done)
@dglazer 
@vadimzalunin 
@saupchurch 
@lh3

vadimzalunin:
+1

saupchurch:
I too, like defaulting to null as false could be interpreted as informative.

+1

dglazer:
+0 -- we're fine with this change.

I have some slight concerns that keep me from being +1, but don't want to block merging unless anyone else is more concerned than I am. Specifically:
- do we have a definition of how this maps to JSON serialization? For example, if a client gets back a `readAlignment` that doesn't specify anything about (e.g.) `secondaryAlignment`, does this PR mean they should now change their interpretation from "false" to "unknown"? Or does a blank return still mean "false", with explicit JSON needed to mean "unknown" ? 
- what do we expect clients to do with "unknown" ? For some fields like `failedVendorQualityChecks` I think it's pretty low-impact. But for other fields like `secondaryAlignment`, if they're not willing to assume that "unknown" means "false" (which is contrary to the spirit of this PR), they'd have to return an error, and therefore backends would never be able to return "unknown" (which again is contrary to the spirit of this PR).

Put those together, and I expect that, in practice, clients will almost always treat missing flags as "false". That's fine, and matches today's behavior, which is why I'm +0. But it seems confusing, which is why I'm not +1.

fnothaft:
> * do we have a definition of how this maps to JSON serialization? For example, if a client gets back a readAlignment that doesn't specify anything about (e.g.) secondaryAlignment, does this PR mean they should now change their interpretation from "false" to "unknown"? Or does a blank return still mean "false", with explicit JSON needed to mean "unknown" ?

I think this would be the correct explanation.

> * what do we expect clients to do with "unknown" ? For some fields like failedVendorQualityChecks I think it's pretty low-impact. But for other fields like secondaryAlignment, if they're not willing to assume that "unknown" means "false" (which is contrary to the spirit of this PR), they'd have to return an error, and therefore backends would never be able to return "unknown" (which again is contrary to the spirit of this PR).

I think this is application specific. E.g., if I'm doing a QC check, I might not care, or might issue a warning message. If I'm downloading the data set of reads so that I can align/assemble the dataset, I may be filtering only on the `primaryMapping`/`unmapped` fields (to get a set of "unique" reads), before throwing away all prior alignment info, or I might check to see if all alignment flags are null (the read has not been previously aligned). If I'm calling variants on aligned reads, I might go throw an error if I get reads with bad alignment flags.

Keep in mind that it is not invalid to have all alignment flags set to null. If I uploaded raw FASTQ using the API, I would expect all alignment flags to be null.

dglazer:
@fnothaft -- not sure which of the two suggested interpretations you think is correct. If a client gets back a `readAlignment` that doesn't specify anything about (e.g.) `secondaryAlignment`, they previously would have assumed that it was false. Should they now:
a) change their interpretation from "false" to "unknown", OR
b) continue to interpret a blank return as "false", with explicit JSON needed to mean "unknown" ?

fnothaft:
@dglazer I'm going with (a). The application then decides whether this "unknown" value is permissible for the analysis.

dglazer:
Got it, thanks. (I still expect that, in practice, most clients will decide "unknown" is permissible, will handle it exactly the same as "false", and that will be fine in all circumstances I can think of.)

dcolligan:
Okay, so it looks like this has all the necessary approval except for @lh3 ... @lh3 , can you weigh in?

lh3:
I said:

>> Yes, supplementaryAlignment etc should default to null actually (meaning not available), not to false.

> Hmm.. probably need more people to comment whether this is the right way.

So, I am okay with this as others have not raised concerns.

dcolligan:
Great.  So @delagoya can you take off the -1 and merge?

delagoya:
Looks good to me. +1. Merging.

PR #294, opened 2015-04-28
Modify sequences/{id} endpoint semantics for serving out sequence bases.

macieksmuga:
The `sequences/{id}` endpoint was modified as follows:

- changed endpoint name to `sequences/{id}/bases` the better to express its semantics
- request object was removed (since this is a `GET` method)
- method definition changed to reflect passing `start`, `end` parameters explicitly instead of via request object.
- `nextPageToken` removed from response object, as client knows length of sequence a priori, can request range explicitly.

### DIFF
--- a/src/main/resources/avro/referencemethods.avdl
+++ b/src/main/resources/avro/referencemethods.avdl
@@ -326,41 +326,19 @@ SearchSequencesResponse searchSequences(
   */
   SearchSequencesRequest request) throws GAException;
 
-/****************  /sequences/{id}  *******************/
+/****************  /sequences/{id}/bases  *******************/
 /**
-The query parameters for a request to `GET /sequence/{id}`, for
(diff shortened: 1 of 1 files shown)


macieksmuga:
I should've mentioned: I need this PR to for implementing the graph reference server so that it can be used to evaluate the first batch of side graphs.

jeromekelleher:
+1 --- totally agree with this, thanks @macieksmuga!

adamnovak:
+1, this is definitely a better way to do it.

delagoya:
+1

jeromekelleher:
Can you rebase this off upstream/master please @macieksmuga? It's ready to merge then I think.

macieksmuga:
@jeromekelleher rebased as requested.

jeromekelleher:
Thanks @macieksmuga, merging.

PR #298, opened 2015-05-01
G2p

jnguyenx:
This protocol defines the associations between genotype and phenotype (G2P). Associations can be made as a result of literature curation, computational modeling, inference, etc., and modeled and shared using this schema.

Here, we follow the dogma of:
```
  Genotype + Environment = Phenotype
```
where a G2P association is between the G(enotype) in the context of some E(environment), which gives rise to a P(henotype). These associations have further evidence, provenance, and attribution.

We leverage the GenomicFeature in the sequenceAnnotation schema here as it can accomodate any genomic feature from a single nucleotide variation (SNV), up through a gene, and/or complex rearrangements. Each can be modeled as genomic features, and generally linked to a phenotype.
Collections of these features can represent a genotype at different levels of completeness. Therefore, we can represent single allelic variation, allelic complement, and multiple variants in a genotype that can each or collectively be associated with a phenotype.

To enable standardized integration, this schema relies heavily on OntologyTerms, both for typing phenotype (for example, using the HPO or DO) but also genomic features (such as using SO) and levels of evidence (ECO).


src/main/resources/avro/genotypephenotype.avdl
```diff
@@ -0,0 +1,176 @@
...
+  union { null, string } id = null;
+
+  /**
+  Examples of some environment types could be drawn from:
+  Ontology for Biomedical Investigations (OBI): http://purl.obofoundry.org/obo/obi/browse
+  Chemical Entities of Interest (ChEBI): http://www.ontobee.org/browser/index.php?o=chebi
+  Environment Ontology (ENVO):  http://www.ontobee.org/browser/index.php?o=ENVO
+  Anatomy (Uberon): http://www.ontobee.org/browser/index.php?o=uberon
+  */
+  org.ga4gh.models.OntologyTerm environmentType;
```
adamnovak: I think you're already in the `org.ga4gh.models` namespace, so this doesn't need to be fully qualified.

src/main/resources/avro/genotypephenotype.avdl
```diff
@@ -0,0 +1,176 @@
...
+  union { null, string } id;
+
+  /** HPO is recommended */
+  org.ga4gh.models.OntologyTerm type;
+
+  /** 
+  PATO is recommended.  Often this qualifier might be for abnormal/normal, 
+  or severity.
+  For example, severe: http://purl.obolibrary.org/obo/PATO_0000396 
+  or abnormal: http://purl.obolibrary.org/obo/PATO_0000460
```
adamnovak: Given one of these URLs, what library code/API can be used to turn it into a human-readable description?

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,144 @@
...
+the subclass closure.
+
+Each query can be made against a string, an array of external 
+identifers (such as for gene or SNP ids), ontology term ids, or 
+full feature/phenotype/evidence objects.
+*/
+record SearchFeaturesRequest {
+
+  union {null, string, ExternalIdentifierQuery, OntologyTermQuery, 
+         GenomicFeatureQuery} feature = null;
```
adamnovak: @jeromekelleher How is the server going to handle deserializing unions of all these things?

AAMargolin:
+1

adamnovak:
Other than what appears to me to be gratuitous namespacing, I am +1 on this too.

jnguyenx:
Can someone grant me access to the repository? I'd like to commit on the branch to fix @adamnovak comment.

skeenan:
added

jnguyenx:
@skeenan thanks!

skeenan:
It seems to me it seems all concerns are dealt with.  Can we get some +1s on this so that we can get it merged?

bwalsh:
+1, lets merge

shadowtramp:
Looks awesome! +1

src/main/resources/avro/genotypephenotypemethods.avdl
pgrosu: @jeromekelleher The server would need to be updated for processing them, which wouldn't be too much of a change.

fnothaft: I don't like the implementation choice of having a five way union for these three fields. Since the `GenomicFeatureQuery`, `PhenotypeQuery`, and `EvidenceQuery` structures (as well as `ExternalIdentifierQuery` and `OntologyTermQuery`) are all new, why not just nest the `ExternalIdentifierQuery`, `OntologyTermQuery`, and `string` parts of the union under each of `GenomicFeatureQuery`, `PhenotypeQuery`, and `EvidenceQuery`? E.g., `GenomicFeatureQuery` would change from:

```
record GenomicFeatureQuery {
  array<org.ga4gh.models.Feature> features;
}
```

to:

```
record GenomicFeatureQuery {
  array<org.ga4gh.models.Feature> features = [];
  union { null, ExternalIdentifierQuery } externalId = null;
  union { null, OntologyTermQuery } ontologyTerm = null;
  union { null, string } searchString = null;
}
```

This would greatly simplify the data model (and in turn, simplify parsing the query), without losing any richness.

pgrosu: Or that :)

fnothaft:
I am -0.9 on the way the `SearchFeaturesRequest` is constructed with 5 way unions, as that makes the schema more complex. As I indicated in my in-line comment, these 5 way unions can be simplified to a two way `{ null, Schema }` union by nesting schemas. If this can be fixed, I would be +1. I don't think that the 5 way union is necessary (as described in my comment), but will change my vote if persuaded that it is functionally necessary.

src/main/resources/avro/genotypephenotypemethods.avdl
jnguyenx: The motivation is to emphasis the fact that only one of those can be passed per query. Having several unions suggests that more than one of those fields can be set, which can be confusing.

Having this 5 way union is self-explanatory in my opinion.

fnothaft: I think it'd be preferable to have a simpler schema and to document that the nullable fields are exclusive. If someone runs a query that defines mutually exclusive fields, you'd respond with an error, as you would if the user provided a query that was invalid for any other reason. While the five way union does implicitly disallow the combination of multiple query objects, it hadn't occurred to me that this was the primary motivation of the union, so I wouldn't _personally_ call it "self explanatory".

diekhans: We absolutely need to document motivation in the data structures.

If the intention is to only allow one parameter per query,
a different queries would be clear than overloading
the request object.

dglazer:
Following up on the ongoing concall -- see `SearchVariantsRequest` (in variantmethods.avdl) for a similar use case, where you can specify either `referenceId` or `referenceName`. Not exactly the same semantics, but similar.

nlwashington:
thanks @dglazer .  will investigate and revise with @jnguyenx today.

src/main/resources/avro/genotypephenotypemethods.avdl
nlwashington: note that this pattern is not new...  we followed a pattern that @diekhans put into the [sequenceAnnotation.avdl at line 55](https://github.com/ga4gh/schemas/blob/1cbf5bab6451b84c96841a59ac8e7d241a4e67ab/src/main/resources/avro/sequenceAnnotations.avdl#L55)

nlwashington: an alternative is to have distinct queries depending on the object type being supplied in the query.  this would be the least ambiguous. 
so instead of 
<pre>
SearchFeaturesRequest {
         GenomicFeatureQuery} feature = null;
} 
</pre>

we might proliferate the searches to:
<pre>
SearchGenomicFeatureByIDRequest {
  ExternalIdentifierQuery feature;
} 
SearchGenomicFeatureByOntologyTermRequest {
  OntologyTermQuery feature;
} 
SearchGenomicFeatureByStringRequest {
  string feature;
} 
SearchGenomicFeatureByGenomicFeatureRequest {
  GenomicFeatureQuery feature;
} 
</pre>

this would be the cleanest implementation.  we would leave these as their ```*FeatureQuery``` record types to allow extension in the future (for example, new parameters specific for the object type).

bwalsh: From my point of view, Nicole's distinct queries has a lot of merit.  It is
explicit and keeps the schema simple.

-Brian Walsh

jnguyenx: Unfortunately this query can have up to 3 parameters (each with the 5-way union). The number of distinct queries will explode which is not a viable option. It's pity because I found this really clean.

Here's the current problematic piece of code:
<pre>
record SearchFeaturesRequest {
         GenomicFeatureQuery} feature = null;
         PhenotypeQuery} phenotype = null;
         EvidenceQuery} evidence = null;
}
</pre>

We're bumping into this issue because Avro does not offer inheritance.
From my readings, the technique to mimic that is to use a n-way union (from [here](http://www.infoq.com/articles/ApacheAvro) and [here](http://technicaltidbit.blogspot.com/2013/02/avro-tips.html)).

I'm keen to keep the current implementation, and as stated above I'm not a big fan of the proposed fix.

@diekhans @fnothaft do you have any other ideas?

fnothaft:
I spoke with a few folks at the plenary and would like to remove my -0.9 for the time being.

skeenan:
Thanls @fnothaft. Merging.

PR #299, opened 2015-05-04
Split ReadAlignment.alignment ...

dcolligan:
... into alignment and graphAlignment

Rationale:
Currently, we can't deserialize JSON data into a python object if its schema has a union with more than one potential non-null type.  When we encounter the union's JSON, there is no type information present, so we would need to hazard a guess about the type, which is error-prone.

### DIFF
--- a/src/main/resources/avro/reads.avdl
+++ b/src/main/resources/avro/reads.avdl
@@ -266,14 +266,20 @@ record ReadAlignment {
   union { boolean, null } failedVendorQualityChecks = false;
 
   /**
-  The alignment for this alignment record. This field will be
-  null if the read is unmapped.
+  The alignment for this alignment record served by a server in "classic"
+  mode. This field will be null if the read is unmapped.
 
-  If an API server supports "classic" mode, it must not return `GraphAlignment`
-  objects here. If the API server supports the "graph" mode and does not support
-  the "classic" mode, it must not return `LinearAlignment` objects here.
+  This attribute can only be non-null if `graphAlignment` is null.
   */
-  union { null, LinearAlignment, GraphAlignment } alignment = null;
+  union { null, LinearAlignment } alignment = null;
+
+  /**
+  The alignment for this alignment record served by a server in "graph"
+  mode.
+
+  This attribute can only be non-null if `alignment` is null.
+  */
+  union { null, GraphAlignment } graphAlignment = null;
 
   /**
   Whether this alignment is secondary. Equivalent to SAM flag 0x100.


macieksmuga:
+1
This is a big help for keeping implementations of the API straightforward.
I'm fine with the naming (`alignment` and `graphAlignment`) but am interested in hearing if others want different names for these fields.

adamnovak:
I would prefer `linearAlignment` and `graphAlignment`, so that graph isn't the weird one.

But either way it's a +1 from me.

pgrosu:
-1 I'm not sure why it would be a problem to update the code for handling multiple types in a union.  It's more natural to express them like it was originally - as follows - and coding it shouldn't be too difficult:

```
union { null, LinearAlignment, GraphAlignment } alignment = null;
```

~p

fnothaft:
I am very strong +1 to this change.

> -1 I'm not sure why it would be a problem to update the code for handling multiple types in a union.

Strongly disagreed. I think I said this on the original PRs that started to add these multitype fields, but there's many reasons that having multiple types in a union is a poor decision. While optional fields (e.g., `union { null, T } field`) are well supported across most infrastructure (e.g., SQL-based systems, code generation tools), moving to a `union { null, T1, T2 } field;` sacrifices straightforward type checking, is incompatible with many COTS SQL-based systems (e.g., Hive, Spark-SQL, Impala, etc). IMO, it is less work for the programmer to process two `union { null, T} field;`'s with different `T`s, as you do a check to see if the field is null, while the `union { null, T1, T2 } field;` requires a type check and cast at runtime. While that check and cast is _straightforward_ to do in some languages (e.g., Scala), it is not straightforward to do in other languages.

pgrosu:
No problem, but you might like the following regarding Hive:

http://docs.hortonworks.com/HDPDocuments/HDP1/HDP-1.2.2/ds_Hive/language_manual/datatypes.html#LanguageManualTypes-UnionTypes

~p

macieksmuga:
@adamnovak good point, however the current naming scheme has the advantage of backwards compatibility. Not sure how vital this is in practice.

dcolligan:
The schemas are undergoing such huge churn... has backward compatibility ever been a priority?  I mean, it's a nice-to-have, but I think getting the schemas correct is more important if this is pre-1.0.  I vote for adopting @adamnovak 's suggestion.

dcolligan:
Just pushed a new commit with @adamnovak 's suggestion

diekhans:
Getting the schemas right is the priority, not backwards
compatibility.

The lower-level data, such as reads, is derived from very stable
practice (BAM) and will are more sane that other parts.

Other components barely fit together, have redundancy and are
inconsistent.

The only way we are going to get to a release that is worth
having backwards compatibility with is to get it build and use
it.

pgrosu:
Agreed with @diekhans, that stability and having the correct massively-parallel-enabled schemas is most important.  It will enable a level of analysis that has not been yet available to NGS as of yet - think of 1000 genomes as just a new dataset that would be loaded up every 15 minutes or less.  I agree with checking them in practice as being the best way, but I would add that the proof-of-concept would need to be also tested on massive data - besides the reference server - to ensure we're not implementing something that already is out there currently.

delagoya:
-1

This represents an major API Change and the latest commit needs to be cleared by the following people before it can move forward:

@jeromekelleher
@fnothaft
@dglazer 
@vadimzalunin 
@saupchurch 

Also @pgrosu have they addressed your concerns? You need to explicitly state "+1" since you did have a "-1" earlier.

macieksmuga:
@delagoya I strongly disagree. The change this PR introduces is explicitly backwards compatible (more so than what it's replacing), and it touches on a part of the API that is being actively built and refined - it affects only graph-based reference and variant sets.

Putting -1's on changes like these (which are in effect refinements and fixes on the initial graph reference/variation PR accepted just recently) only to require unanimous consensus is significantly blocking our ability to  fix issues and iteratively refine this brand-new part of the API.

pgrosu:
@delagoya Even though a -1 was initially my vote I'll switch to +1, since I unfortunately don't have the energy to convince folks of the benefits. It might be a philosophical hurdle that we're facing, but I'm worried it is deeper than that.  I'm concerned that for this project the deep understanding of currently available computer science concepts is not being carefully examined and/or ubiquitously understood - which I tried to do through my small contributions over the past year.  It would be understandable if people might be scared of what they don't think is possible to change - the "unknown unknowns", which we can easily elaborate on - but rather more worrisome, that they might be ignored in order to choose the easier path, while not fully exploring the impact it might have down the line.  I'm unfortunately not part of the conference calls to know what is being discussed or decided, though it would be nice to be.  I understand we have to start small to deliver something that is conducted through a test-driven development (TDD) process via [server](https://github.com/ga4gh/server/), but we should at least periodically take pause and ask ourselves if we might be limiting the bigger picture of how this will all work together on a larger scale, while examining all the pros and cons.  As an example here, if in the future we will have additional `union`s to check, how complex do we want to get with separate checks which can be easily encapsulated.  At least we need to carefully understand the impact how it might limit us - with actual examples of analysis on real systems - before we make a major change through a quick vote.

Paul

delagoya:
@macieksmuga The above vote reflects a new process that was discussed on the Data Working Group call a few weeks ago about making changes to the main artifacts of teams that no longer hold regular sessions. In this case the ReadsTaskTeam and `reads.avdl`. The process is described in  `CONTRIBUTING.md`. 

The proposed PR adds an attribute to an existing record type. I consider adding an attribute a major change, but opinions may differ and I would like to ask for discussion of that point. 

If the group comes to consensus that backwards compatible changes (such as adding null attributes) should not be considered as a `Major Enhancement` I am happy to remove the -1 vote and extra consensus process.

macieksmuga:
Thanks for the clarification, @delagoya! 
For what it's worth, even though this change is to the `reads.avdl` file, it reflects the work of the Reference Variation Task Team, not the Reads Task Team - and this team is currently very active and is holding regular sessions.

pgrosu:
@macieksmuga The ReadsTaskTeam was the one that spent a lot of time last year - including me - on tweaking it to its most necessary and compact components that would minimize the amount of information being transmitted.  Adding a new `union` adds additional data that would need to be transmitted over the wire, which can have a huge impact when we sent billions or trillions of reads over the wire.

adamnovak:
@delagoya , do you want to take the lead on wrangling the people who have to evaluate this PR in order to get it in? It looks like, for example, @saupchurch has only two GitHub contributions in the last year, and so some form of contact other than repeated @-mentioning might be required to actually get his attention. I would guess that you have lines of communication that @dcolligan and I don't.

I think we're likely to run into more cases where the server people need a change in the schemas to make things implementable, and some of the still-not-fully-implemented schemas belong to retired task teams. I suppose this will be a test of the Major Enhancement process, to see if it can effectively address those cases.

fnothaft:
@pgrosu the tuples are nullable, and it is semantically invalid to populate both, so the impact on serialized object size is negligible in serialization frameworks that are null-aware.

dglazer:
I'd be a quick +1 if we went back to `alignment` and `graphAlignment`, as in the original PR. (I.e. I'm +1 on splitting the optional union field into two optional individual fields.)

If people feel strongly about changing `alignment` to `linearAlignment`, I'd like to understand why. The short version of my concern is that it would gratuitously break backwards compatibility, so that (for example) if someone wanted to build an RNA analysis app, they couldn't seamlessly test interop against both an existing 0.5.1-compatible server and a new HEAD-compatible server. I'll say more if needed.

fnothaft:
> I'd be a quick +1 if we went back to alignment and graphAlignment, as in the original PR. (I.e. I'm +1 on splitting the optional union field into two optional individual fields.)

OK with me.

> If the group comes to consensus that backwards compatible changes (such as adding null attributes) should not be considered as a Major Enhancement I am happy to remove the -1 vote and extra consensus process.

I think this would be a reasonable, low overhead approach.

dcolligan:
@dglazer Ok, I've pushed a commit that implements the backwards-compatible naming scheme.

pgrosu:
@fnothaft By working off the idea that null-aware frameworks fit with the idea of null not being considered a major enhancement, then I'm good with that too for a +1. 

@dglazer I'm okay for now with `alignment` being the primary one, and we can explore that again in the future as this technology/analysis develops even further.  It might not be a bad idea to have that one flexible in your implementation in case it needs to be updated down the line, in case it begins to impinge on clarity.  For instance, you could implement an interface that maps between the two: `alignment <-> linearAlignment`. But more specifically what do you have in mind in your roadmap of putting together?  This way we would be in sync - without having divergent downstream surprises - to which we could also provide input.

Thanks,
Paul

macieksmuga:
Remember when "camera" used to mean "film camera" by default and you needed to specify "digital camera" if that's what you meant?

delagoya:
+1 

@macieksmuga I've added the `RefVariationTaskTeam` label to the issue and updated the comment to reflect that an active working group is on the case. Since this has been reviewed by a few folks from ReadTaskTeam, going to +1 status.

delagoya:
@macieksmuga Also can you guys label your issues? This will signal to me that an active group is reviewing and prevent me from applying an automatic down vote. Thanks!

pgrosu:
@macieksmuga Well, cameras would begin to approach the perspective :)  I'm thinking more along the lines of [ChIP-Seq](http://en.wikipedia.org/wiki/ChIP-sequencing), which is also an alignment besides the obvious approaches of whole-genome sequencing, exome sequencing, RNA-Seq, etc.  It's just that we should remain flexible and nimble by becoming comfortable with the idea that today, the way we approach the fields of genetics and computation is drastically different than 10 years ago - which will progress into the future.  I'm not saying to have dynamic schemas without any controlled vocabulary, but that we should plan for integration in the future, since migrating that kind of data will not be feasible and the schema is a better way (i.e. think of pointers in C).  I remember when I was in class many years ago, and using @richarddurbin's [Biological Sequence Analysis](http://www.amazon.com/Biological-Sequence-Analysis-Probabilistic-Proteins/dp/0521629713) as our textbook.  Those concepts have a specific framework as an introduction that are still applicable today, but because of newer and cheaper technology we re-categorize those elements based on the framework of our experiments.  In one experiment our read is an alignment, while in another we're looking at it through the lens of a peptide sequence.  We want to find those connections, and we're still adding to the vocabulary :)  And regarding the area of computer science - though some concepts have remained constant - the same principles as in genetics has allowed us to augment our categories and processes we apply to them.  Though it is usually more consistent and simpler by working in that domain from the perspective of mathematics.

~p

macieksmuga:
@delagoya thank you, and yes, I'll gladly label the issues raised by the ref var team as you requested!

dglazer:
@dcolligan -- thanks for the backwards-compatible naming -- +1

@macieksmuga  -- I agree the References Variation task team is very active. (And I'm very happy with the progress you all are making.) However, I still think @delagoya is correctly using the new process here. As he said, the trigger is  *"changes to the main artifacts of teams that no longer hold regular sessions. In this case the ReadsTaskTeam and reads.avdl."* I think it's working well to make forward progress without unnecessary backwards pain. 

@delagoya -- thank you for helping us get this new process working, as we work to find the right balance of safe and fast.

@macieksmuga -- remember when "car[riage]" needed to be modified by "horseless" ? :)

jeromekelleher:
I'm +1 after we update the documentation to explain the new setup. Exactly one of these fields is non-null, and which one depends on the server's 'mode'?

pgrosu:
@jeromekelleher  Yes the server preforms this before sending it over the wire, and then the client after receiving it just sets the appropriate elements from their default values.  Below is a link describing how this is done via Jackson with the `Include.NON_EMPTY` and `Include.NON_NULL` settings:

http://oraclejavazone.blogspot.com/2014/11/jackson-ignore-null-empty-values-in.html

~p

saupchurch:
I see you edited to remove the invocation of new methods, but wanted to add my +1 as well and second @jeromekelleher on documentation update

pgrosu:
Sean, sorry for my confusion but which documentation location, and in what place was the invocation of new methods removed?  I'm trying to be in sync and I have several places I'm looking at.  I'm also +1 with @jeromekelleher on the documentation update - it usually makes things more predictable for me.

Thanks,
Paul

saupchurch:
Sorry, I wasn't clear. @delagoya edited his original -1 and removed the MajorEnhancement tag.  That's the change I was referring to.

pgrosu:
Ah, no problem :)  Just didn't wanted to assume something incorrectly.

Thanks for the clarification,
~p

dcolligan:
@jeromekelleher by documentation, do you mean just editing the text above the attributes, or something else?

jeromekelleher:
@dcolligan Yes, just update the local docstrings. We should have a docstring for both attributes, and explain how they are related to each other.

dcolligan:
@jeromekelleher alright, updated the docstrings

jeromekelleher:
Excellent, thanks @dcolligan.

+1

dcolligan:
Is this ready to merge now?  @vadimzalunin is the only one that hasn't given a +1 in @delagoya 's list of people that needed to approve, but I haven't seen him on this thread...

delagoya:
Looks like all issues have been resolved. Merging.

PR #301, opened 2015-05-05
metadata.avdl time formats

mbaudis:
This PR addresses the previously discussed use of ISO8601 for date/time attributes.

One problem here is the lack of a specific dateTime format in the current AVRO implementation (this will be addressed in v1.8(?)). For now, the PR clarifies the ISO use instead of the vague "timestamp" description, and recommends the proper formatting with the attribute type set to string.

Other changes :

* .created => . recordCreated
* .updated => recordUpdated
* Experiment.runDate => Experiment.runTime
* Sample.age => Sample.ageAtSampling

### DIFF
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -32,13 +32,13 @@ record Individual {
   TODO: The format definition is temporary and will be updated to 
   the upcoming AVRO time format (timestamp-millis) when this is supported.
   */
-  union { null, string } recordCreateTime = null;
+  string recordCreateTime = null;
 
   /**
   The time at which this record was last updated.
   Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
   */
-  union { null, string } recordUpdateTime = null;
+  string recordUpdateTime = null;
 
   /**
   The species of this individual. Using
@@ -119,13 +119,13 @@ record Sample {
   The time at which this record was created. 
   Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
   */
-  union { null, string } recordCreateTime = null;
+  string  recordCreateTime = null;
 
   /**
   The time at which this record was last updated.
   Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
   */
-  union { null, string } recordUpdateTime = null;
+  string recordUpdateTime = null;
 
   /**
   The time at which this sample was taken from the individual.
@@ -193,13 +193,13 @@ record Experiment {
   The time at which this record was created. 
   Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
   */
-  union { null, string } recordCreateTime = null;
+  string recordCreateTime = null;
 
   /**
   The time at which this record was last updated.
   Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
   */
-  union { null, string } recordUpdateTime = null;
+  string recordUpdateTime = null;
 
   /**
   The time at which this experiment was performed.
@@ -276,13 +276,13 @@ record IndividualGroup {
   The time at which this record was created. 
   Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
   */
-  union { null, string } recordCreateTime = null;
+  string recordCreateTime = null;
 
   /**
   The time at which this record was last updated.
   Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
   */
-  union { null, string } recordUpdateTime = null;
+  string recordUpdateTime = null;
 
   /** The type of individual group. */
   union { null, string } type = null;
@@ -312,13 +312,13 @@ record Analysis {
   The time at which this record was created. 
   Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
   */
-  union { null, string } recordCreateTime = null;
+  string recordCreateTime = null;
 
   /**
   The time at which this record was last updated.
   Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
   */
-  union { null, string } recordUpdateTime = null;
+  string recordUpdateTime = null;
 
   /** The type of analysis. */
   union { null, string } type = null;
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -27,15 +27,18 @@ record Individual {
   union { null, string } description = null;
 
   /**
-  The time at which this individual was created in milliseconds from the epoch.
+  The time at which this record was created. 
+  Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
+  TODO: The format definition is temporary and will be updated to 
+  the upcoming AVRO time format (timestamp-millis) when this is supported.
   */
-  union { null, long } created = null;
+  union { null, string } recordCreated = null;
 
   /**
-  The time at which this individual was last updated in milliseconds
-  from the epoch.
+  The time at which this record was last updated.
+  Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
   */
-  union { null, long } updated = null;
+  union { null, string } recordUpdated = null;
 
   /**
   The species of this individual. Using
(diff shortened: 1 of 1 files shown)


adamnovak:
I can +1 this. But I think `ageAtSampling` does need to have a unit specified, since it did before.

mbaudis:
@adamnovak Well, we can do it the easy, the medium, and the ontology-crazy way. 

Easy:

```
/**
The age of the individual (not of the sample) at time of sample collection.
Unit is years. This field may be approximate.
*/
union { null, long } ageAtSampling = null;
```

Medium:

```
record TimeSpan {
  long timedInterval;
  string unitOfTime;
}

/**
The age of the individual (not of the sample) at time of sample collection.
This field may be approximate.
*/
union { null, TimeSpan } ageAtSampling = null;
```

Ontology-crazy:

```
record TimeSpan {
  OntologyTerm timedInterval;
  OntologyTerm unitOfTime;
}

/**
The age of the individual (not of the sample) at time of sample collection.
This field may be approximate.
*/
union { null, TimeSpan } ageAtSampling = null;
```

Now, I don't want to say that I'm opposed to the "ontology-crazy" way; this depends on the way we hash out the OntologyTerm object definition. We can talk about my ideas for that any time (or e.g. on this Thursday's MTT call or in Leiden ...).

Just remember - "easy" will require refactoring of data from e.g. lots of pediatric cancer sources (though I'm all for that), but also provides a rather bleak description of murine samples ...

But the best practical approach right now would be "easy" with "year" definition, right? I could make this explicit.

adamnovak:
We could also do easy with days or seconds. Since it can be approximate, you can just up-convert.

delagoya:
This schema is pulled into the reads.avdl. As such it is a major API change to reads implementors. @jeromekelleher can you take a look to see if there needs to be any review by `ReadsTaskTeam` implementors?

jeromekelleher:
I don't think this affects existing implementations, since I think the metadata.avdl is new since 0.5.1 @delagoya. Can you confirm this please @mbaudis?

delagoya:
@jeromekelleher true, but you will need to support it in major revisions later as one of the main API builders. Would this type of change be considered something that the group must weigh in on?

mbaudis:
There will be future changes, as soon as the AVRO spec. supports a dateTime format (in my reading timestamp-millis based on ISO8601 ?).

Anyway:

- the "milliseconds from the epoch" wasn't a good definition (see the other discussions referring to timestamp)
- there is broad consensus that ISO is the correct reference

So IMHO it is better to do the first step getting the data reference right before waiting for the AVRO update.

As for the attribute renaming: IMHO relevant to remove ambiguities (e.g. "created" could be used for the sample collection or experiment date, while "recordCreated" specifically refers to the data itself).

As for @adamnovak's point about the unit for "ageAtSampling": See above. It was a recommendation/question (months) before, now is a recommendation/question (years). Should be resolved in one of the ways described in my previous comment; so let me know if I should make the "years" for now as required unit per comment.

adamnovak:
OK, years for now will probably work fine. We will be able to change it later when we decide we need this to work for mice and fruit flies.

diekhans:
Michael Baudis <notifications@github.com> writes:
> So IMHO it is better to do the first step getting the data reference right
> before waiting for the AVRO update.

Completely agree.  We need to work towards consistency cross the entire
API.

delagoya:
+1

mbaudis:
@diekhans so, maybe +1 this, so I feel more comfortable merging it?

diekhans:
I like the approach, although it might be good to puzzle for a few minutes
over the naming.

recordUpdateTime would be clearer than recordUpdated, which sounds like a boolean.

pgrosu:
I would say more simple that that.  Just remove `record` and call them the following:

`DateCreated`
`DateModified`

If you prefer the `Date` can be substituted with `Timestamp`, but I don't feel it sounds as natural.

~p

diekhans:
Paul Grosu <notifications@github.com> writes:
> DateCreated
> DateModified

-1

The goal is to distinguish the metadata record change times from
the data it describes.  These need to all be times, not just
dates, as there can be multiple modifications on the same day.

pgrosu:
Ok, thanks for the clarification.  So then, how about the following:

`CreatedOn`
`ModifiedOn`

diekhans:
Michael, here is a counter proposal

   string metadataCreateTime;

name is very explicit.

I don't see a use case for these fields to be null.

fnothaft:
@diekhans RE:

> I don't see a use case for these fields to be null.

A case I could imagine is if you are importing metadata from a system that doesn't log creation/update time. Is that a possible case, or one we should disregard? I am OK with disregard personally, but wanted to play devil's advocate.

diekhans:
@fnothaft good example.  I am sure this happens, as ever possible
combination of bad data happens.

However, since they could substitute import time with out really
causing much confusion, I am not sure it's worth the extra
schema complexity.

I am fine with NULL and see what happens.

fnothaft:
> However, since they could substitute import time with out really
causing much confusion, I am not sure it's worth the extra
schema complexity.

+1, let's just make sure that's documented.

mbaudis:
@diekhans @fnothaft Regarding the creation vs. import possibility, IMHO the attributes describe the system-local status, so import would either lead to creation time => recordCreated & recordUpdated, or only the latter (if merging data for an existing UUID). If data is properly propagated through compatible systems, there would be pointers to the records of origin e.g. (through ```derivedFrom``` or ```accessions ``` in upcoming metadata refinement.

Regarding NULL: I'm actually against having this option (value insertion/modification should be basic system behaviour), but left it in from previously => Support for removing?

@diekhans @pgrosu Mark's proposal ```recordUpdateTime``` is the most fitting (we even could do ```recordUpdateDateTime```), but what about unnecessary verbosity? IMHO attribute names should be logical & as descriptive as the context requires; so recordUpdated seems fine when considering the accomp. description (otherwise we end up with ```dateOfBirthISO8601```). However, only using ```created``` would invite at least logical if not namespace conflicts. So I'm still for the ```recordUpdated```.

diekhans:
Hi @mbaudis, verbose when it leads to easy to understand is
good, even a goal.  This stuff doesn't get typed near as often
as it is read. IDEs that completely names is common.  Making a
mistake in understanding a value wastes far more time than
typing.

So recordUpdated sounds like a boolean to me.  Has the record
been updated?  Naming identifiers so they are intuitive when read
as native language really helps understanding, in my experience.

I find recordUpdateTime much clearer.  I think
recordUpdateDateTime is unnecessary, as time implies data for
compute data.

Using `record' I find maybe a little confusing.  Hence the
metaDataCreateTime (could have dataCreateTime to go with it).
However, I am not sure of this.

pgrosu:
@mbaudis, @diekhans No problem, I'm okay with the longer name.  I was just thinking since these timestamps are encapsulated in either `Individual`, `Sample`, `Experiment`, `IndividualGroup` or `Analysis`, they would infer the type of record they represent.  But I agree with Mark, that they ensure people will fully understand their meaning while typing without the risk of misrepresentation.  With dynamic schemas we can always update the fieldnames later on as they become second nature.

mbaudis:
@diekhans O.k.; let's do it with ```recordUpdateTime``` etc. I prefer this style to the ```dataCreateTime``` option, since there will be (many) instances were the original data creation time might be different & non-recoverable. And there are/will be separate attributes for e.g. ```samplingDateTime```, ```Experiment.runTime``` etc. which will more clearly reference "data origin" time points.

Before I edit this in: comments on NULL for ```recordCreateTime```, ```recordUpdateTime```? I'm in favour of de-nulling...

diekhans:
I am not 100% sure about the naming, but it's a big step forward
so I will +1 recordUpdateTime which can not be NULL.  It can
evolve.

about dataCreateTime was more conceptual, not that particular
name. It is very important to the make clear the different
between times that apply to the record and those that apply to
the data.

Although, I feel strongly that all of GA4GH needs to be a
immutable model.  That is, a record has a create time and then a
link to the previous version, which has it's own create time.

pgrosu:
@diekhans So then how about also having a `history` variable that stores the array of `id`s associated to the particular record.  Then we can treat them like a stack or a queue, depending on if you want to perform LIFO or FIFO.  In either case it would be ordered.  If that is too much to transmit, we can have it be like a reversed linked-list with a field called `previousRecordId` that would be a `union { null, string }` that would be pointing to the previous record's `id` in its particular historical sequence.

helenp:
I have been following this with interest. It makes sense to me to generate all the creation/update possibilities and explain how to use them. Experience suggests that the default information is typically about the record and we should make this cleasr. So recordCreation, recordUpdate, and dataCreation/dataUpdate would be my preferred naming - time could be appended to all of these if that makes it clearer.

mbaudis:
@pgrosu @helenp @diekhans We had discussed the tracking/mutability issues previously, with especially @diekhans being a strong proponent of immutable data objects (which implies a strong tracking/versioning system). The way you describe it would be roughly one of the possible implementations. The main basic concepts (IMHO) would be 

* a system where you keep the UUID for all iterations of the data object, and leave the versioning to the implementation (having provided the basic requirement with ```recordUpdateTime```)
* a system with truly immutable objects, where each change of the data object leads to a new one with new UUID, and the tracking is done via something like @pgrosu's ```history``` (I would rather name it ```representationOf``` or something in that vein) - this wouldn't even need the ```recordUpdateTime```, but require a tight system & storage overhead

While I see both as viable options, the 2nd way imposes stronger restrictions on the implementation; this amy be good or bad, depending on your POV / position in the implementation chain ...

mbaudis:
Since there had been enough support for the time attribute changes, and IMHO concerns have been addressed accordingly, I'll merge.

pgrosu:
@mbaudis Memories are coming back and I agree :)  We will see what would be preferred, since with a real implementation more people/teams will weigh in on that.  Usually deltas of the changes are kept as snapshots to reconstruct the data-object(s) to a specific point in time, thus avoiding the need to store complete gigabyte/petabyte versions of the data-object(s).  And as usual there are many avenues that we can be explore in that domain :)

`p

P.S. I'm also +1 for merging.

Issue #165, opened 2014-10-19
Format of ids for GAOntologyTerm


cmungall:
Current docs state:

```
  /**
  The ID defined by the external onotology source.
  (e.g. `http://purl.obolibrary.org/obo/OBI_0001271`)
  */
  string id;
```

This is fairly open ended and we can imagine confusion and inconsistent usage here.

For the ontologies currently referenced in the metadata schema, e.g.
- HP
- UBERON
- CL
- OBI

Terms are typically referenced in two ways.
## URIs/IRIs

For many biological ontologies these are typically obolibrary purls, which follow:

```
http://purl.obolibrary.org/obo/<IDSPACE>_<NUMERICFRAGMENT>
```

See: http://www.obofoundry.org/id-policy.shtml
## OBO-Style identifiers

Typically follow the form

```
<IDSPACE>:<NUMERICFRAGMENT>
```
## Options
1. The schema should mandate URIs only (using the URI form recommended by the source ontology)
2. The schema should mandate OBO-Style IDs
3. The schema should have separate 'id' and 'iri' fields
4. The schema should have a flexible field

Option 1 is probably the conceptually simplest. Option 2 is not very future proof as it doesn't allow open-ended expansion to any ontology out there on the semantic web. Option 3 is probably overkill.

I would advocate option 4. To elaborate, we allow the field to contain either a URI or a CURIE (https://en.wikipedia.org/wiki/CURIE see also http://www.w3.org/TR/curie/), without the brackets. We then assume the existence of a number of implicit qname prefixes. E.g.

```
@prefix UBERON http://purl.obolibrary.org/obo/UBERON_
@prefix CL http://purl.obolibrary.org/obo/CL_
@prefix OBI http://purl.obolibrary.org/obo/OBI_
@prefix NCBITaxon http://purl.obolibrary.org/obo/NCBITaxon_
```

This could potentially live in a separate JSON-LD context file.

This is also consistent with the translation in the OBO-Format spec: http://oboformat.googlecode.com/svn/trunk/doc/obo-syntax.html#5.9.1

I would be happy to branch and make a pull request, but I thought it worthwhile polling for opinions. Need this to be future-proof, consistent - but also not over-engineered.

buske:
The MME is hoping to converge on a compatible representation. Curious if there are any updates?

cmungall:
[...]

PR #302, opened 2015-05-07
Variant annotation support

sarahhunt:
These three new protocols form an initial proposal for variant annotation support. 

VariantAnnotation and AlleleAnnotation records hold different types of annotation derived by comparing a Variant or Allele to a set of reference data.  AnnotationSets group VariantAnnotation/ AlleleAnnotation records and hold full details of all software and reference data sets used. 

The effect of alternate alleles on transcript sets is the first type of annotation to be considered in detail.

Two methods protocols are proposed. alleleAnnotationmethods.avdl supports the mining of pre-calculated annotation data and annotateAllelemethods.avdl supports annotation as a service.


reece:
Verbatim copy of GA4GH_prototype_annotations2.docx, sent by Sarah Hunt to ga4gh-dwg-annotation@googlegroups.com on May 7 is available as a gist at https://gist.github.com/reece/0604e25e78f5c63e1a5f .

pgrosu:
Thank you for sharing @reece - it definitely helps having a concrete example.  I'm not on that email list, since it would be helpful to see how we got to this design.  Is it free to join?

Thanks,
Paul

reece:
@pgrosu I think you need to contact @skeenan to be added to the annotation list.

pgrosu:
Ah, thank you @reece.  @skeenan any possibility?  It would just help with being in sync.

Thank you,
Paul

reece:
@sarahhunt - Thanks for the proposal. That helped tremendously!

Below is a long list of comments specific to your proposal and, more importantly, some design goals that I believe we should consider.

Putting money where my mouth is, I created an example (https://gist.github.com/reece/880aa4283d404a2563cd) that illustrates some of these goals. (N.B. My example has it's own flaws, like not being able to store annotations from multiple versions of the same tool. This group should consider it a strawman alternative, not a proposed solution.)

## Specific suggestions about the proposal:

* Use consistent syntax. I'm especially thinking about camelCase v. underscore v. all-caps. Assuming we're standardizing on cameCase, "IMPACT" should be "impact", "variantannotations" should be "variantAnnotations", etc. I'm fine with notable exceptions like "HGVSg". (I don't care at all about which convention, just that there is exactly one.)

* Drop the concept of annotationset in the annotation itself. This is better modeled as an annotation set containing annotations (rather than annotations referring to their container).


## Goals:

* Files structures containing variants and variant annotations should be easily mergable. The primary use case is to merge precomputed variant annotations with a patient's set of annotations. 

* Predictions should be typed. That is, annotations from a single source or of a particular type should follow a similar schema.

* Use stable external refs wherever possible. For example, "effects" should use SO ids. (The human readable names are not guaranteed stable and have changed at least once to my knowledge.)

* All locations must have a reference sequence specified by accession (not name, like "chr22"). Doing otherwise is as good as a address number without a street name. I think it's essential that this annotation spec make the sequence reference unambiguous.

pcingola:
Hi @rece, thank you for your comments, here are some answers (at least my opinion):

i) Consistent syntax: The examples you mentioned seem to in proper cameCase in Sarah's PR ('variantannotations' is lowercase only in POST URLs, and I did not find the "IMPACT" in all caps). May be I'm looking at the wrong place...?

ii) The concept of annotationSet in the annotation itself is analogous to the concept of 'readGroupId' in 'ReadAlignment' itself. This was added as a suggestion form the group (the initial version did not have it) and we all thought it was quite useful.

Goals:
i) Your comment about "variants and variant annotations should be easily mergable", in my opinion has been take care of. "Merging" variant with their respective annotations is a function of the annotation algorithm, which is included in Sarah's proposal (see AnnotateVariantsRequest). The API user should NEVER have to this manually because doing so is error prone and annotation algorithms take care of many corner cases that users are usually unaware of.

ii) I'm not sure what you meant with "Predictions should be typed", can you clarify?

iii) SO IDs: We could include them, but in my opinion this makes the record less useful ('missense_variant' is much more readable than 'SO:0001583'). In any case, I think that what you mention is a problem about versioning ontologies. A problem that we don't intend to solve here, furthermore, the structure of the ontology can also change, so using IDs doesn't solve it. We should leave this as it is and modify it later, when GA4GH solves ontologies versioning issues and then use an 'OntologyTerm'.

iv) Annotations are anchored to variants, which already have reference sequence ID, so asking for "chr22" makes a lot of sense because this is a context specific query (may be I misunderstood your point here). In any case, there should perhaps be an option for using sequence IDs instead of chromosome names, but I would suggest to add it after the PR has been accepted.

helenp:
Hi @pcingola iii) you could use the SO term, id and version. If you don't do this then ids need to be looked up and there is an annotation-ontology mapping problem and an ontology versioning problem to deal with. We will work on the ontology component to make some recommendations for discussion in Leiden so that this is consistent across different components.
All ontologies can change structure and there are diff tools that allow tracking of changes - moving between ontology versions is a complex problem and I don't think GA4GH can solve this though there are some tools available.

sarahhunt:
Hi @reece  and @pcingola 

A few additional comments - 

The avro schemas follow the standard syntax: https://github.com/ga4gh/schemas/blob/master/CONTRIBUTING.md#syntax_style, apart from the HGVS attributes, which maybe should really be all be lower case too. @reece - I suspect you are looking at the output from the prototype implementation that I mailed round the group rather than the schemas. As mentioned, it is a work in progress and you have spotted a couple of compliance errors! Maybe it was too early to share this output.

The annotationSet id is in the annotation record to support the extraction of a subset of annotation - maybe over a specific gene - and the annotationSet record which contains information on software and reference data versions used. An annotationSet could be a database of 'all current variants vs all current transcripts', so if the set referenced the annotations it could get quite unwieldy.   

Thanks for the recommendation @helenp. The effects are currently represented by an array of OntologyTerm's. There is no version in the OntologyTerm at the moment, ( just source, id, name) but it would make sense to add it.

pcingola:
+1

pgrosu:
+1 I had a chance to look through it, and I like it.  it would be nice to see it integrated with the rest of the schemas in an implementation.

Thanks,
`p

nlwashington:
note that ontology discussions are happening at #165.  i would strongly advocate for using OntologyTerm (or at least the curie ids) in the schema early, as it is always hard for people to adopt using more structure later.  I will be adding more comments on the overall schema shortly.

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,244 @@
...
+import idl "sequenceAnnotations.avdl";
+import idl "ontologies.avdl";
+
+
+
+/******************  /variantannotations/search  *********************/
+/** This request maps to the body of `POST /variantannotations/search` as JSON. */
+record SearchVariantAnnotationsRequest {
+  /** Required. The IDs of the annotation sets to search over. */
+  array<string> annotationSetIds = [];
```
calbach: Per https://github.com/ga4gh/schemas/pull/253 this should now be singular.

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,244 @@
...
+  /**
+  If non empty, will restrict the query to annotation sets within the
+  given datasets.
+  */
+  array<string> datasetIds = [];
+
+  /**
+  If non empty, will restrict the query to the specified annotation sets
+  */
+  array<string> annotationSetIds = [];
```
calbach: This shouldn't be necessary, since you already have getAnnotationSet

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,233 @@
...
+  */
+  array<TranscriptEffect> transcriptEffects = [];
+
+
+  /**
+  The names of known variants which are co-located with this variant.
+  these can use used to look up disease associations, ClinVar statuses,
+  allele frequencies in reference panels, etc
+  */
+  array<string> coLocatedVariants =[];
```
calbach: Should these be variant IDs?

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,233 @@
...
+}
+
+
+
+
+/**
+A `VariantAnnotation` record represents the result of comparing a variant
+to a set of reference data.
+*/
+record VariantAnnotation {
```
calbach: Could you add more documentation here about when this record should be used? Specifically, when would I use a `VariantAnnotation` over an `AlleleAnnotation`?

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,244 @@
...
+  overlapping reference alleles should be returned.
+  */
+  long end;
+
+
+  /**
+  Only return variant annotations for any of these features
+  If null, return all variant annotations in specified window.
+  */
+  union { null, array<org.ga4gh.models.Feature> } features;
```
calbach: How does the matching work? Is this matching feature IDs of TranscriptEffects or is this doing a live JOIN against the regions annotated by those features?

If it's the former, this should be named more specifically and should refer to the feature_id. If it's the latter, I'm not sure what this buys you beyond the above start/end parameters.

src/main/resources/avro/alleleAnnotations.avdl
calbach: Is one for linear-only and one for graph-only?

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,244 @@
...
+  /** This request maps to the body of `POST /variantannotations/search` as JSON. */
+  SearchVariantAnnotationsRequest request) throws GAException;
+
+
+
+/******************  /alleleannotations/search  *********************/
+/** This request maps to the body of `POST /alleleannotations/search` as JSON. */
+record SearchAlleleAnnotationsRequest {
+  /** Required. The IDs of the annotation sets to search over. */
+  array<string> annotationSetIds = [];
```
calbach: Per #253 this should now *maybe* be singular. Though this one is definitely higher value..

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,244 @@
...
+
+/****************** /annotationsets/search *********************/
+/** This request maps to the body of `POST /annotationsets/search` as JSON. */
+record SearchAnnotationSetsRequest {
+
+  /**
+  If non empty, will restrict the query to annotation sets within the
+  given datasets.
+  */
+  array<string> datasetIds = [];
```
calbach: Per #253 this should now be singular.

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,244 @@
...
+    /**
+    This request maps to the body of `POST /alleleannotations/search` as JSON.
+    */
+    SearchAlleleAnnotationsRequest request) throws GAException;
+
+
+
+/****************** /annotationsets/search *********************/
+/** This request maps to the body of `POST /annotationsets/search` as JSON. */
+record SearchAnnotationSetsRequest {
```
calbach: Paginate?

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,233 @@
...
+
+  //TODO: derive unique id from digest of data [location, allele, transcript?]
+
+
+  /** The id of the transcript feature the annotation is relative to
+  */
+  string featureId;
+
+
+  /** Alternate allele - a variant may have more than one alternate allele, each of which
```
calbach: In this case, would the variant have two TranscriptEffect records? One for each allele?

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,233 @@
...
+
+
+  /** The ID of the transcript effect record
+  */
+  string id;
+
+  //TODO: derive unique id from digest of data [location, allele, transcript?]
+
+
+  /** The id of the transcript feature the annotation is relative to
```
calbach: Are there some kind of limitations around what type of feature this should be?

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,233 @@
...
+
+}
+
+/**
+An `AlleleAnnotation` record represents the result of comparing a allele
+to a set of reference data.
+*/
+record AlleleAnnotation {
+
+  /** The allele ID. */
```
calbach: Is this the ID of the allele, or the ID of the AlleleAnnotation?

diekhans: Please, all answer should result in updated to the
documentation.  We have someone going through some of the
documentation now.  Even the more well documentation schemas
have a lot of ambiguity and uncertainty.

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: The TranscriptEffect record is intended just for transcripts. We will need to add other record types in future, such as one for regulatory features, but these will have different attributes.

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: It would have 2 for each transcript. Does the new comment at line 25 make this clear?

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: Yes that's right. There's a new comment at the top of the file to this effect.

I'm aware of the disussion re splitting graph and linear modes into different branches and it won't be hard to split this PR as neccessary.

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: Thanks - added!

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: Yes - that makes more sense.

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: It's the former. Do the new comments at the top of the file make this clear?

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: I've switched them all to singular for now - we can revisit later.

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: I've added ids for both variant and allele annotation records - these were not present initially.

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: Fine to revisit later, especially while the dataset discussion continues.

sarahhunt:
Thanks for the input @calbach!

@diekhans  - I've added to the documentation, but if you, or anyone else, find anything else too vague do let me know.

src/main/resources/avro/alleleAnnotations.avdl
calbach: Yes, thanks.

src/main/resources/avro/alleleAnnotations.avdl
calbach: Perhaps it makes sense to just strip out the graph pieces for now.

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,255 @@
...
+
+  /** The ID of this VariantAnnotation. */
+  string id;
+
+  /** The variant ID. */
+  string variantId;
+
+  /** The ID of the annotation set this record belongs to. */
+  string annotationSetId;
+
```
calbach: nit: Please remove these double newlines throughout, unless this style convention is used elsewhere(?).

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,274 @@
...
+  long end;
+
+
+  /**
+  Only return variant annotations for any of these features
+  Features may include specific transcripts, genes or regulatory elements
+  This or a location (referenceName/referenceId plus start and end) must be supplied
+  If null, return all variant annotations in specified window.
+  */
+  union { null, array<org.ga4gh.models.Feature> } features;
```
calbach: I still think it's too confusing to embed the entire feature here. The coordinates here will have a very complicated interaction with the above proposed referenceName, start, end.

`feature_ids` I think would be better, or just remove this field entirely for the first draft.

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,274 @@
...
+  union { null, string } referenceName = null;
+
+  /**
+  Only return variants with reference alleles on the reference with this ID. One
+  of this field or `referenceName` or `features` is required.
+  */
+  union { null, string } referenceId = null;
+
+  /**
+  Required if referenceName or referenceId supplied.
```
calbach: Is this copied from somewhere else? I believe the other endpoints say the opposite:
if start or end are provided -> you must also provide refName or refId.

The current documentation means I can't do a query which returns all variant annotations on a given reference.

calbach:
I've added some more comments inline. In general things look sane, but I haven't dug too much into the details of the contents of the actual annotations. I would like it if someone from a more bio-heavy background could do a thorough review of those fields.

src/main/resources/avro/alleleAnnotationmethods.avdl
diekhans: yes, reference by ids rather than copying features will results
a simpler data model.

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: The required start/end was copied from master variants/search as an attempt at consistency. I see the reads/search has optional start and end which does make more sense.

The new refVar-linear branch, which is presumably soon to be master,  now searches variants by segment and reads by referenceId and optional start/end. @adamnovak - will reads/search be updated to take segments, meaning this is the standard variant annotations should also use?

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: Updated to take feature_ids now

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: Thanks - that's tidied up now.

jacmarjorie:
+1

AngieHinrichs:
+1 Thanks for laying the foundation!

src/main/resources/avro/alleleAnnotationmethods.avdl
adamnovak: I think we do want to standardize on Segments for specifying intervals. And in our graph prototype we found a compelling reason to make the search interval optional for variants/search too, so we should probably settle on optional Segments.

rajgopals:
+1

skeenan:
Merging to master.

stevenbrenner:
+1

Issue #306, opened 2015-05-13
CI yielding warning lines


dcolligan:
We have CI now in the schemas repo... yay!

We should make it stricter; there are a bunch of warning lines getting spewed out in the current CI:

https://travis-ci.org/ga4gh/schemas

```
$ nosetests -v tests
testMaven (test_maven.TestMaven) ... ok
[WARNING] Avro: Invalid default for field recordCreateTime: null not a "string"
[WARNING] Avro: Invalid default for field recordUpdateTime: null not a "string"
...
```

dcolligan:
Fixed in #307

PR #307, opened 2015-05-13
Fail CI if avro-tools issues warnings

dcolligan:
- change metadata time types from string to union { null, string } to
    eliminate warnings
- also remove compile_schemas verbosity flag, plus some other
    refactoring

Issue #306

### DIFF
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -32,13 +32,13 @@ record Individual {
   TODO: The format definition is temporary and will be updated to 
   the upcoming AVRO time format (timestamp-millis) when this is supported.
   */
-  string recordCreateTime = null;
+  string recordCreateTime;
 
   /**
   The time at which this record was last updated.
   Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
   */
-  string recordUpdateTime = null;
+  string recordUpdateTime;
 
   /**
   The species of this individual. Using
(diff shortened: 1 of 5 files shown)


jeromekelleher:
Great work @dcolligan, +1.

macieksmuga:
+1, thanks @dcolligan!

mbaudis:
-1
This has been a design decision in https://github.com/ga4gh/schemas/pull/301. I would rather support keeping it that way; basic requirements for any record should be (uu)id, creation time, update time, no matter what it is about (one could argue about update if considering an object immutable).

dcolligan:
@mbaudis the warning I'm fixing here stems from having an attribute set to a default value which is impossible for that attribute's type to accommodate.  Alternatively, an empty string could be the default.  Would you be satisfied with that?

macieksmuga:
@mbaudis @dcolligan the other alternative - potentially closer to the spirit of #301 - is to leave the definition as a string, but not to provide any default value (the user is then required to supply an explicit value in all cases).

mbaudis:
>  a default value which is impossible for that attribute's type to accommodate
 So it should be just

```string recordCreateTime;```

... right? Sorry for this. Agree & change?

dcolligan:
@mbaudis sounds good to me.  Just pushed a change to that effect.

macieksmuga:
@dcolligan @mbaudis thanks for working through this - it was blocking me as well.

macieksmuga:
+1 for the updated fix.

mbaudis:
+1 now, and retraction of my -1 above.
However, I feel not comfortable enough to merge it in myself due to the test suite changes in this commit - could somebody else do this, please?

jeromekelleher:
Looks like we have agreement; merging.

PR #310, opened 2015-05-20
Move fragment data into the fragment object (WIP)

afirth:
I'll leave this here as a starting point for discussion. This reopens the can of worms from #212.
By keeping the fragmentId in the ReadAlignment, construction of a bam equivalent object should remain extremely simple.

### DIFF
--- a/src/main/resources/avro/readmethods.avdl
+++ b/src/main/resources/avro/readmethods.avdl
@@ -79,6 +79,84 @@ SearchReadsResponse searchReads(
   /** This request maps to the body of `POST /reads/search` as JSON. */
   SearchReadsRequest request) throws GAException;
 
+/******************  /fragments/search  *********************/
+/**
+This request maps to the body of `POST /fragments/search` as JSON.
+
+If a reference is specified, all queried `ReadGroup`s must be aligned
+to `ReferenceSet`s containing that same `Reference`. If no reference is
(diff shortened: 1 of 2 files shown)


dglazer:
@afirth  , do you plan to add method changes into  this PR also? I think all the tradeoffs here -- benefit of this change to app developers, cost of this change to server developers -- depend on the proposed methods.

(The details matter, because the fragment-centric methods I've seen proposed in the past are potentially useful for app developers, but likely very expensive for server developers.)

dglazer:
Or is the idea that no new methods are needed, because the new `readAlignments[]` array in the `Fragment` object provides all the extra capability that apps would need?

If so, that's a nice clean change. I'd still want to understand how an app would use this (ideally with an example of a real-world app that can't efficiently use the API today, but would be able to after this change), and what the cost to server developers would be to support this.

afirth:
Hi David @dglazer,

If you want to quantify RNA transcript abundance, a common operation is to get fragments, not just reads, which are compatible with the transcript structure.
This is efficiently done if you have the fragment info. If you do not have this info then you must find the mate on client side since next_mate_position is not sufficient to say if the mate also aligns with the transcript. This is a severe shortcoming in the SAM format. The intention is to develop against the reference server, but we wanted to propose this against /schema as soon as possible for comment. It's reasonable to imagine a server implementation which caches compatible fragment alignments, and it doesn't make sense to me to do that inside the ReadAlignment framework.

PR #319, opened 2015-06-04
Create Dataset in metadata.avdl

mbaudis:
This PR introduces  ```Dataset``` record type (at the expense of the previous limited ```IndividualGroup```). Please see #248 for some background on this.
```Dataset``` is seen with inherent *evolvability*. There was consensus among the metadata task team that the best way would be to put this forward in a skeleton format, with supposed future feature refinement.

### DIFF
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -264,6 +264,7 @@ Represents a group of data objects of one or more types (e.g. all Individuals, S
 associated with a clinical study; or e.g. a trio in genetic diagnostics.).
 This concept may be expanded in the future (ontology for describing the type of dataset ...).
 This record replaces and extends the previous "IndividualGroup" record type.
+TODO: Determination of scope, structure, specific attributes.
 */
 record Dataset {
   /** The dataset UUID. This is globally unique. */
@@ -288,29 +289,12 @@ record Dataset {
   string recordUpdateTime;
 
   /**
-  The type of dataset. Examples could be "trio", "metaanalysis", "GWAS"...
-  */
-  union { null, string } datasetType = null;
-
-  /**
-  The uuid's of included records.
-  */
-  array<string> recordsIncluded = [];
-
-  /**
-  The type of included records (ga4gh.Sample, ga4gh.reads ...).
-  TODO: Define necessity of this since types can be derived from the 
-  referenced records.
-  TODO: Define if records are limited to a single type or heterogeneous.
-  */
-  union { null, string } recordsType = null;
-
-  /**
-  A map of additional individual group information.
+  A map of additional information.
   */
   map<array<string>> info = {};
   
 }
+
 /**
 An analysis contains an interpretation of one or several experiments.
 (e.g. SNVs, copy number variations, methylation status) together with
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -260,16 +260,19 @@ record Experiment {
 }
 
 /**
-Represents a group of individuals. (e.g. a trio)
+Represents a group of data objects of one or more types (e.g. all Individuals, Samples, Experiments 
+associated with a clinical study; or e.g. a trio in genetic diagnostics.).
+This concept may be expanded in the future (ontology for describing the type of dataset ...).
+This record replaces and extends the previous "IndividualGroup" record type.
 */
-record IndividualGroup {
-  /** The individual group UUID. This is globally unique. */
+record Dataset {
+  /** The dataset UUID. This is globally unique. */
   string id;
 
-  /** The name of the individual group. */
+  /** The name of the dataset. */
   union { null, string } name = null;
 
-  /** A description of the individual group. */
+  /** A description of the dataset. */
   union { null, string } description = null;
 
   /**
@@ -284,15 +287,30 @@ record IndividualGroup {
   */
   string recordUpdateTime;
 
-  /** The type of individual group. */
-  union { null, string } type = null;
+  /**
+  The type of dataset. Examples could be "trio", "metaanalysis", "GWAS"...
+  */
+  union { null, string } datasetType = null;
+
+  /**
+  The uuid's of included records.
+  */
+  array<string> recordsIncluded = [];
+
+  /**
+  The type of included records (ga4gh.Sample, ga4gh.reads ...).
+  TODO: Define necessity of this since types can be derived from the 
+  referenced records.
+  TODO: Define if records are limited to a single type or heterogeneous.
+  */
+  union { null, string } recordsType = null;
 
   /**
   A map of additional individual group information.
   */
   map<array<string>> info = {};
+  
 }
-
 /**
 An analysis contains an interpretation of one or several experiments.
 (e.g. SNVs, copy number variations, methylation status) together with
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -260,20 +260,16 @@ record Experiment {
 }
 
 /**
-Represents a group of data objects of one or more types (e.g. all Individuals, Samples, Experiments 
-associated with a clinical study; or e.g. a trio in genetic diagnostics.).
-This concept may be expanded in the future (ontology for describing the type of dataset ...).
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by dglazer** [comment]
Per previous discussion in #248, I'm -1 on this change as is, since it
combines a non-controversial code cleanup (moving the existing Dataset
object to a new more global home) with a needs-discussion introduction of
new semantics (e.g. type, list of contents).  I'd prefer to quickly do the
first change, and then discuss the second set separately.

**Comment by richarddurbin** [comment]
I also support doing the first change, with the same semantics but in the long 
term home, then having a separate PR and discussion on changed semantics.

Richard

> On 4 Jun 2015, at 16:59, David Glazer <notifications@github.com> wrote:
> 
> Per previous discussion in #248, I'm -1 on this change as is, since it
> combines a non-controversial code cleanup (moving the existing Dataset
> object to a new more global home) with a needs-discussion introduction of
> new semantics (e.g. type, list of contents). I'd prefer to quickly do the
> first change, and then discuss the second set separately.
> —
> Reply to this email directly or view it on GitHub <https://github.com/ga4gh/schemas/pull/319#issuecomment-108946876>.
> 




-- 

**Comment by mbaudis** [comment]
@dglazer @richarddurbin Well, it seemed to me as a rather *skeleton* version already, with only few new attributes thought of getting discussions on track.

Considering your opinions, I'll pare ```metadata.Dataset``` down further to basically minimum record requirements (including the universal/generic ```recordCreateTime```, ```recordUpdateTime```, ```description```, ```name```, ```info```). The specific attributes ```datasetType ```, ```recordsIncluded``` and ```recordsType``` will be removed.

The extended version can be found at https://github.com/ga4gh/metadata-team/blob/master/avro-playground/Dataset_test.avdl.

Does this sound reasonable?

**Comment by diekhans** [comment]
Even if we are simple moving the schema to another, it would be good to fix the original problem in that this isn't documented. Perhaps those involved in define it could provide documentation.

**Comment by helenp** [comment]
Dataset - a data item that is an aggregate of other data items of the same type that have something in common.
http://purl.obolibrary.org/obo/IAO_0000100

**Comment by diekhans** [comment]
It appears that the goal was not of the same type.  More akin to `study'.

**Comment by mbaudis** [comment]
@diekhans Well, that is one of the things that have to be defined. There seems to be agreement about a fixed content (i.e. uuid list, though possibly with updates) compared to dynamic queries (as previously stated, I would tentatively include both options).
So maybe one could go for 2 different record types: 

* ```Dataset```, which captures a group of records of the same type which have some common feature (equal to @helenp s reference) - and which may be dynamically generated (e.g. all reads from breast cancer samples)
* ```Study```, which is a wrapper for a static collection of record of possibly different type, which all are part of an external concept, with rich annotations and clear provenance etc.

However, we may keep in mind that this PR here is now a technical procedure, fixing the ill-placed ```reads.Dataset``` into a skeleton ```metadata.Dataset```; so checks and +/- on this are appreciated ...

**Comment by diekhans** [comment]
OK, +1 on you moving it...

**Comment by helenp** [comment]
+1 @diekhans dataset and study separation,

**Comment by helenp** [comment]
+1 on moving dataset

**Comment by dglazer** [comment]
I remain +1 on moving the existing Dataset object, and -1 on adding/changing any fields to it as part of the move PR.

If you want to include comments describing the original intent, I'd suggest something like:
```
/**
A Dataset is a data-provider-specified collection of related data of multiple types.
Logically, it's akin to a folder -- it's up to the provider what goes into the folder.
*/
```
Possibly extended with something like:
```
/**
For server implementors, they're a useful level of granularity for implementing
administrative features such as access control (e.g. Dataset X is public;
Dataset Y is only available to lab Z's collaborators) and billing (e.g. the costs
of hosting Dataset Y should be charged to lab Z).

For data curators, they're 'the simplest thing that could possibly work' for grouping
data (e.g. Dataset X has all the reads, variants, and expression levels for a
particular research project; Dataset Y has all the work product from a particular
grant).

For data accessors, they're a simple way to scope exploration and analysis
(e.g. are there any supporting examples in 1000genomes?
what's the distribution of that result in the data from our project?)   
*/
```

pgrosu:
@dglazer  Why the hesitation to move it?  Isn't a Dataset a form of metadata with a collection of records that are associated to?  I understand the need for an administrative portion, and maybe we can call that a DataGroup or RecordCollection - what do you think?  

On a separate note, I think we probably should begin to make a list of defined terms before we categorize them to a schema.  Could we have such a documentation somewhere?  We could start with is a Dataset, Study, etc.

dglazer:
No hesitation on moving -- as I say above
> I remain +1 on moving the existing Dataset object

pgrosu:
@dglazer Ah, thank you for highlighting - not enough coffee in the system yet :)  If the body could just naturally produce a bare minimum of caffeine in the morning...coffee bean genome + human genome...hmm...

mbaudis:
@dglazer I'm still not sure about your opinion regarding the move.

As noted above, the current version of the PR just moves to metadata, and adds the generic  ```recordCreateTime```, ```recordUpdateTime```, ```name``` and ```info```. The timestamps, like the id, are basic parts of metadata record definitions; ```name``` and ```info``` are also standard but can be null/empty (this could be discussed as part of general object structure discussions, but right now is default). So, do you have any specific objections to that?

As for your comment draft: While this would be very helpful, it currently describes a hybrid Dataset/Study type. This clashes with the definition provided by @helenp above, which is much more restrictive - see also my note above (I'm rather neutral on the split, but understand the wish for a static, documented "Study" type).

dglazer:
@mbaudis -- sorry I wasn't clear:
- I am +1 on moving the Dataset object, **unchanged**, from reads.avdl to metadata.avdl
- I am -1 on making any changes to it as part of the move, including lightweight ones
- I am +0 on including comments that describe how it currently works

I'm not trying to be difficult here -- I just think that getting metadata right is important, and I don't want to combine a code cleanup with other changes. Those extra fields you're proposing may be standard in metadata.avdl, and seem reasonable from first principles, but they're not yet standard across all the objects in /schemas, and I haven't seen (maybe I missed?) discussion of if/how/when we want to standardize on them, if/how they should affect the various search methods, etc.

Re my comment draft -- I was responding to @diekhans ' request that we document *the original intent*, not weighing in on any future changes to that intent. 

(Happy to talk in person at some point if that would help us get in sync.)

akiani:
+1 on moving dataset to metadata too. Apologies if this is a repeat question, but is there a diagram somewhere (on the repo) that visualizes the current data ontology? I can see the maintenance overhead reason for not having something like that yet.

richarddurbin:
+1 from me.  Let’s move on.

> On 5 Jun 2015, at 17:49, Michael Baudis <notifications@github.com> wrote:
> 
> @dglazer <https://github.com/dglazer> I'm still not sure about your opinion regarding the move.
> 
> As noted above, the current version of the PR just moves to metadata, and adds the generic recordCreateTime, recordUpdateTime, name and info. The timestamps, like the id, are basic parts of metadata record definitions; name and info are also standard but can be null/empty (this could be discussed as part of general object structure discussions, but right now is default). So, do you have any specific objections to that?
> 
> As for your comment draft: While this would be very helpful, it currently describes a hybrid Dataset/Study type. This clashes with the definition provided by @helenp <https://github.com/helenp> above, which is much more restrictive - see also my note above (I'm rather neutral on the split, but understand the wish for a static, documented "Study" type).
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/ga4gh/schemas/pull/319#issuecomment-109360074>.
> 




-- 

mbaudis:
I have reverted the PR to the "minimal move" type supported by @dglazer. I understand/support his arguments from a software management POV.
Merging it now due to abundance of +1. This will be followed shortly by new PRs addressing the metadata record conformity and then probably attribute/documentation add-ons for discussion.

pgrosu:
@mbaudis I thought there was no hesitation to the move, but rather any changes within its interpretation.

mbaudis:
@pgrosu see @dglazer above:

>I am -1 on making any changes to it as part of the move, including lightweight ones

... which I can understand/accept from a purely software management perspective, though there are other concepts. Anyway done ow, for the 1st part.

pgrosu:
@mbaudis Ok, no problem - thanks for clarifying the interpretation.

PR #324, opened 2015-06-07
Add technical attributes to metadata.avdl Dataset record

mbaudis:
Adding "technical" attributes to the ```Dataset``` record type; at the same time removing ```IndividualGroup``` from metadata, since this is a special case of ```Dataset``` and/or a possibly upcoming ```Study``` record type.


mbaudis:
Closing to clean up commits/branches.

PR #325, opened 2015-06-07
document the current behavior of the Dataset object

dglazer:
This comment-only PR is in response to @jeromekelleher 's request (in #248) for clearer documentation of the original intent of the `Dataset` object, and to the discussion (in #319) that led to moving it to metadata.avdl (thanks @mbaudis).

My main goal here is to capture the current behavior in a formal visible way, as a baseline for discussion of any changes. I'd prefer to merge this in as is, and then discuss any changes to behavior or fields in a separate PR, but I'm flexible on process.

### DIFF
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -294,6 +294,8 @@ record IndividualGroup {
 }
 
 /**
+NOTE: there's discussion about changing the role of Dataset, possibly as follows:
+
 Represents a group of contextually related data objects of (e.g. all Individuals, Samples, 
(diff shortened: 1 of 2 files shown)


mbaudis:
Thanks @dglazer; this is an excellent approach especially for letting people figure out the scope & structure of the ```Dataset``` record type.

Btw., there is also now https://github.com/ga4gh/schemas/pull/324.

skeenan:
Can we get some +1s on this comment only PR for merging.

dglazer:
Closing in favor of #389 (which has the same content, but without merge conflicts).

PR #329, opened 2015-06-10
add fastq object

saupchurch:
This adds the ReadSequence object which is intended to store the raw read sequence as described by a fastq file (with quality string) or a fasta file (without).  Storing the raw read sequence accomplishes several things:

1) retrieval of mate pair sequence #310 would now be achievable with an id search on ReadSequence instead of ReadAlignment
2) this allows direct raw sequence retrieval for quantification methods that do not use aligned reads (ex: http://pachterlab.github.io/kallisto/about.html)  Previously, the fastq would have to be reconstructed downstream from the ReadAlignment CIGAR and Reference sequence.
3) processing of read data has to be done to get a ReadAlignment which means some decisions have been made for anyone requesting that aligned data.  Access to raw sequence allows an investigator to make their own decisions regarding alignment parameters, etc.
4) unless I'm mistaken this would be the only way to access unaligned reads in the initial data as even the (un-implemented) read is described as having alignment data.  This is a further complication for item (2).

### DIFF
--- a/src/main/resources/avro/reads.avdl
+++ b/src/main/resources/avro/reads.avdl
@@ -148,6 +148,24 @@ record ReadGroupSet {
   // referenceSet.
 }
 
+/**
+A read sequence corresponds to one read entry in a fastq file, if the
+quality string is present, or a fasta entry, if absent.
+*/
+record ReadSequence {
+  /** Sequence identifier. */
+  string id;
+
+  /** Description. */
+  union { null, string } description = null;
+
+  /** Raw read sequence. */
+  string sequence;
+
+  /** Quality values. */
+  union { null, string } quality = null;
+}
+
 /** A linear alignment can be represented by one CIGAR string. */
 record LinearAlignment {
   /** The position of this alignment. */


pgrosu:
+1 @saupchurch

vadimzalunin:
+1

fnothaft:
I'm -0 on this. IMO, it is preferable to store an unaligned read via the `ReadAlignment` record, instead of adding a new record. Admittedly, this is somewhat semantically imprecise. Is the implication of this pull request that anyone implementing a read server would be required to store both unaligned/aligned reads? As far as I can see, to do the actions you're interested in you'd have to store both the unaligned and the aligned reads.

diekhans:
A read server would have to deliver unaligned reads, it can do on-the-fly
create of the objects.  GA4GH is a transfer protocol, there is no expectation
on storage formats.

It would be useful to understand the queries one would have to
get unaligned reads.

fnothaft:
@diekhans IMO, it's _subtly_ more complex than that. The way I see it, the nit is that the `AlignedRead` object stores the _aligned_ sequence and quality scores, not the original sequence and quality scores. I believe we only mask bases if the bases are hard clipped (I don't think we remove soft clipped bases). I agree with you that we don't place an expectation on storage formats, but by storing the aligned bases/quality scores in the aligned read model, you can't guarantee that you can convert on-the-fly between aligned/unaligned without a loss of fidelity.

diekhans:
Unlike BAM, the AlignedRead object doesn't allow storing unaligned reads.   Maybe allowing this would be be an approch.

Yes if one uses hard clipping in creating the mapping, the fastq can not be recreated.  TCGA only keeps BAMs for DNA and hard clipping in not used.

TCGA found at some labs had used an RNA-seq mapper that modified reads and thus had to go back to get fastq files.



-- Sent from my computer phone

-----Original Message-----

fnothaft:
@diekhans isn't it wrong for the `AlignedRead` object to disallow unaligned reads? Fundamentally, in a set of aligned reads, knowing that read _n_ did not align is important information.

> TCGA only keeps BAMs for DNA and hard clipping in not used.

This would disallow chimeric mappings, then.

pgrosu:
So then why not just have an `UnalignedRead` record with a `transmitOnlyAlignedReads` flag that can be used to block transmission of it whenever necessary - or we can use request specific fields in the transmission.

diekhans:
@fnothaft, absolutely!  We can't lose unaligned reads.  Perhaps
someone from the Reads Task Team can comment on why unaligned
reads are not included.

I believe chimeric mappings are represented using soft clipping

dglazer:
-1 -- I completely agree we need to support unaligned reads, but unless I'm missing something, we already do. See below for the results of a search for "unmapped" in the schemas/ repo. If there are problems with the current representation let's discuss and fix, but we don't have to start from scratch.

From `reads.avdl`:
```
/**
  The alignment for this alignment record served by a server in "classic"
  mode. This field will be null if the read is unmapped.

  This attribute can only be non-null if `graphAlignment` is null.
  */
  union { null, LinearAlignment } alignment = null;
```

And from `readmethods.avdl`:
```
record SearchReadsRequest {
  ...
  /**
  The reference to query. Leaving blank returns results from all
  references, including unmapped reads - this could be very large.
  */
  union { null, string } referenceId = null;
```
and
```
record SearchReadsResponse {
  /**
  The list of matching alignment records, sorted by position.
  Unmapped reads, which have no position, are returned last.
  */
  array<org.ga4gh.models.ReadAlignment> alignments = [];
```

fnothaft:
@dglazer that was my thought as well.

pgrosu:
@dglazer, @fnothaft - I understand but the scope is different here.  Think of the `ReadSequence` id as your term id in an inverted index of term-vs-document matrix - as one would do when building a simple search-engine - where it points to its presence in different documents, and in this case the reads would be the documents showing their association to those terms:

terms-docs-matrix

Just like Sean (@saupchurch) mentioned above, and also Alastair (@afirth) mentioned in the following post [https://github.com/ga4gh/schemas/pull/310#issuecomment-105922422], we do not want be only dependent by the program that generated the alignment to search the transmitted/stored read data structure.  Regarding querying only for unaligned reads, that cannot be possible now since we're retrieving everything and otherwise we would need to search for the mate, which is not straightforward.  We want the query to be fragment-focused.

Hope it helps,
~p

diekhans:
@dglazer  The path (or reference name in the 0.5 version) are not nullable.  

It is probably is a good economy of representation to not add an new record. It would be good if the reads group to fix the schema and document the intentions:

* There needs to be a query to return only the unaligned reads without getting all of the aligned reads. This is commonly done when searching for viral sequences.
* We need to document how a readgroup of unaligned reads is represented.  We really should be able to share data without aligning it.

dglazer:
@diekhans , thanks for getting into the details of things we could improve. To your three points:
- which reference field is not nullable? (The `referenceSetId` in `ReadGroup` is nullable, with a comment saying `Required if there are any read alignments.`, which I believe handles this use case well.)
- I agree, it would be useful to for `SearchReadsRequest` to have a way to request **only** the unmapped reads. (It's not needed for FASTQ files, since everything there is unmapped, but it is sometimes useful.) Do you have a query syntax in mind? I suggest doing a separate PR for that, and making sure to get input from our current server implementors (@calbach and @jeromekelleher).
- I think it is documented -- per the comment in my first bullet, you just don't populate the `referenceSetId` field. But better documentation is certainly better -- what would you like to see clarified?

pgrosu:
@dglazer I believe Mark is referring to [lines 186-187 in reads.avdl](https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/reads.avdl#L186-187):

```
record GraphAlignment {
  /** The `Path` against which the read is aligned */
  Path path;
```

Currently `Path` is not part of in [any search method](https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/readmethods.avdl) for `Read`.  The definition for `Path` is in [commons.avdl on lines 388-410](https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/common.avdl#L388-410).

~p

saupchurch:
Thank you @dglazer for pointing out what I missed in the ReadAlignment.  I agree with the points raised and suggest the following:

1) I will modify this PR (title and content) into a comment only PR with the goal of better documentation on how to represent unaligned reads in the existing ReadAlignment.
2) SearchReadsRequest needs a way to request only unmapped reads - to be done in a separate PR
3) The GraphAlignment option needs a way to store unmapped reads

dglazer:
Sounds good @saupchurch -- I like all three suggestions.

fnothaft:
@saupchurch I would be +1 to that.

saupchurch:
Closing as this is out of date with respect to work being done in the documentation branch.

PR #330, opened 2015-06-10
Removing OntologyTerm (duplication of ontologies.avdl)

mbaudis:
Removing OntologyTerm (duplication of ontologies.avdl)


diekhans:
+1 and thanks @mbaudis for clean up my mistake for not deleting
this when in the sequence annotation pull request.

nlwashington:
+1

PR #335, opened 2015-06-12
Changing the type of the exists field in a beacon's response to boolean.

mcupak:
I believe the general consensus is that the "overlap" response type should be dropped in favour of simple true/false responses.


mfiume:
+1 We had an extensive discussion on the group thread about the beacon response, and decided that beacons respond either yes or no. In the case of an error, the exist field should be omitted.

mcupak:

PR #338, opened 2015-06-12
Removed beacon's email field.

mcupak:
It seems there is very little value in having an email directly in the schema. Contact info can be listed on the beacon home page, possibly with more information and a better way of contacting the maintainer than an email.


mcupak:

PR #339, opened 2015-06-12
Initializing beacon's frequency field to null as it's optional.

mcupak:


mcupak:

PR #345, opened 2015-06-19
Introducing the "Disease" record type

mbaudis:
When keeping the ```OntologyTerm``` record type sparse, there is no way to annotate associated features (fallback/buffer text descriptions, time of observation etc.). This commit creates a wrapper ```Disease``` record type and changes the content of the ```diseases``` attribute accordingly

### DIFF
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -69,9 +69,8 @@ record Individual {
 
   /**
   Diseases with which the individual has been diagnosed.
-  TODO: Is this the right representation?
   */
-  array<OntologyTerm> diseases = [];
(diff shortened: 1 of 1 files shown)


mellybelly:
Ok so I think we really need to understand the use case for defining a disease here. There are very many definitions of disease in the world. There is a lot of work going on this in the G2P schema already, which has already been accepted. see https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/genotypephenotype.avdl#L77

I think this can be either a disease or a phenotype. 

I also don't think the ontology term should be optional, this is trouble downstream. We can always annotate up. @jjfarrell ICD can be used as an OntologyTerm.

@nlwashington @AAMargolin can comment further

PR #351, opened 2015-07-02
Proposal for quantitative ageOfOnset

mbaudis:
ageOfOnset needs a quantitative representation, since needed granularity is arbitrary & unpredictable for many applications. This proposal

* uses the previous ```ageOfOnset``` attribute name for an ISO8601 encoded age string
* renames the ontology based age object to ```ageOfOnsetClass``` (open for different recommendations)


mbaudis:
Closing this for clean-up; g2p has to be re-evaluated.

PR #352, opened 2015-07-07
Move Linear Model to Master Branch

adamnovak:
These commits get rid of the entire "mode" system, and also all of the graph semantics. Variants in this model are straightforward string replacements that occur at linear coordinate positions.

The architecture of Sequences, Segments, and References are all still around, but Joins, Paths, and Alleles are gone.

Since master has, while this was in development, acquired an Allele annotation branch, some manual integration will be required to adjust what's on master now to the non-existence of the Allele object. Perhaps the Allele annotation stuff would be better housed on the refVar-graph-unary branch, where Alleles are alive and well?


jeromekelleher:
This looks good, but I'm struggling to understand what's been reverted and what's new. Would it be possible to get a quick summary of how this differs from 0.5.1 (minus the GA prefix noise)?

dcolligan:
I second @jeromekelleher 's comments.  It would be good to get a list of backwards-incompatible changes that have been introduced since 0.5.1.

sarahhunt:
@adamnovak - I see what you mean about the Allele Annotation schema. I've sent a PR through to the refVar-linear branch with the graph dependent parts removed. Let me know if there's a better way to do it!

adamnovak:
I've now merged @sarahhunt 's stuff into this PR and gotten this able to be merged with the master branch.

There's an interactive diff from 0.5.1 available at: https://github.com/ga4gh/schemas/compare/v0.5.1...refVar-linear

In addition to all the build system changes, Travcis stuff, the addition of test code, and so on, there are a number of new APIs (Allele Annotations, G2P, Metadata, SequenceAnnotation). There are also many changes to Beacon. References to bases by reference name have been phased out in favor of sequence ID pointers, and we have Sequences, Strands, Segments, Sides, and Positions added and used pervasively. Some new search criteria have been added to existing methods, and new methods have been added to support the new objects.

adamnovak:
Everybody on the call seemed to like the "alpha topic branches" idea for splitting features like graphs out until they're implemented in the reference server. What's still in the way of merging this?

jeromekelleher:
It's still not clear what's changing here @adamnovak --- can you try to break this down and explain what's going on please? This is a big change and people need to be able to understand clearly what's happening before they can vote for it.

dcolligan:
(Jerome and I are, it should be said, very much in favor of the spirit of this change -- that is, getting the master schemas branch down to the minimum needed to run the non-graph server.)

adamnovak:
Relative to the current master, the following has changed:

* Graphs and graph mode have been removed.
    * The Graph Mode FAQ has been removed.
    * The introduction to side graphs at the top of `common.avdl` has been removed.
    * The `/mode/{mode}/` endpoint has been removed.
    * The `Join` object has been removed.
    * The `/joins/search` endpoint has been removed.
    * `ReferenceSet`s can no longer include `Join`s (since they are removed).
    * The `Path` and `Region` objects have been removed, with uses replaced by `Segment`s.
    * The `GraphAlignment` has been removed from the Reads API.
    * The subgraph search endpoints (`/subgraph/segments` and `/subgraph/joins`) have been removed.
    
* `Alleles` have been removed, in favor of the traditional edit-based `Variants`.
    * The `Allele` object has been removed.
    * The `/alleles/search` and `/alleles/{id}` endpoints have been removed.
    * The `AlleleAnnotation` and references to it have been removed.
    * The `/alleleannotations/search` endpoint has been removed.
    * The `AlleleCall` object has been removed.
    * The `alleleIds` field of `Variant` has been removed.
    * Consequently, the `referenceBases` and `alternateBases` fields of `Variant` are now required.
    
* `Segment`s have been standardized as the way to refer to areas of the genome.
    * The `sequenceId`, `start`, and `length` fields on `Reference` have been replaced with a `Segment`.
    * The `referenceId`, `start`, and `end` fields on `SearchVariantsRequest` have been replaced with a `Segment`.
    * The `referenceName`, `start`, and `end` fields on `Variant` have been removed and replaced by a `Segment`.

* IDs have been standardized as the way to refer to `Sequence`s and `CallSet`s.
    * The `referenceName` field of `Position` has been removed. `Position`s must now point to `Sequence`s by ID.
    * The Sequence Annotations API now identify locations with `Segment`s instead of `Path`s.
    * The `referenceName` field on `SearchVariantsRequest` has been removed.
    * The `callSetName` field has been removed from `Call`s; they should now use the `callSetId` field instead.

* `Calls` have been revised.
    * `Call`s now have an `id` field.
    * The `variantId` field has been removed from `Call`s. `Call`s are to be contained in `Variant`s and cannot be retrieved independently.
    * The `phaseSet` of a `Call` can now only be null or a string, not an array of strings.
    
* Other miscellaneous changes
    * The `sourceAccessions` field on `Reference` is now required to include a version number.
    * `Feature`s now include a `Strand`.
    * `VariantSet`s can no longer contain `Sequence`s, and `/variantsets/{id}/sequences/search` and `/variantsets/{id}/sequences/{id}` have been removed. The correct way to augment an existing `ReferenceSet` is to create a new `ReferenceSet` that inherits from it.

I don't think it's correct to try and prepare a whole changelog from the last release here. We might want that document when we actually do the release (maybe we should keep it in the repository and update it when we make changes?), but most of those changes have nothing to do with this pull request, and involve the additions of other API features that we're wanting to strip out anyway, or things that are already merged into master.

benedictpaten:
+1 This needs to be merged asap, in my view.

diekhans:
+1 this is holding up a lot of development.

jeromekelleher:
Thanks for the summary @adamnovak, this makes it much easier. I'm happy with the vast majority of the changes, as I think they are good forwards-compatible ideas. 

I'm not quite sure I follow what's happened to Calls though. Is this a straightforward revert to non-graph behaviour, or is there new stuff? The id field looks new to me --- why is this here?

benedictpaten:
@dglazer @calbach  Any comments here? This PR, and subsequent PRs to strip down master to what is implemented in the reference server should get us within spitting distance of implementing https://github.com/ga4gh/schemas/pull/353

calbach:
Process note: This is massive. Is this an accumulation of changes from many unrelated PRs to the refvar topic branch which have already been individually reviewed? This is problematic as reviewing the actual code in this PR reduces to me having to reread almost the entirety of every avro file in the repo (in many cases without a diff). I will just review the change summary to start.

> Graphs and graph mode have been removed.
> The Allele object has been removed.

I'm in favor of this.

> The sequenceId, start, and length fields on Reference have been replaced with a Segment.
> The referenceId, start, and end fields on SearchVariantsRequest have been replaced with a Segment.
> The referenceName, start, and end fields on Variant have been removed and replaced by a Segment.

Is there justification for these changes beyond laying the framework for future graph changes? These are pretty significant breaking changes so there should be a tangible benefit.

Also, I need to come back to this point and look through the code, but I'm not sure about the reference+sequence relationship. IIRC, a reference ID is also a sequence ID. In the linear branch though, all sequences are references, right? I guess the goal might also be to lay a framework for the future graph changes?

> IDs have been standardized as the way to refer to Sequences and CallSets.
> The referenceName field of Position has been removed. Positions must now point to Sequences by ID.
> The referenceName field on SearchVariantsRequest has been removed.
> The callSetName field has been removed from Calls; they should now use the callSetId field instead.

I'd say I'm +0 on this. It's cleaner from a programming perspective, but removal of referenceName and callsetName (i.e. sampleId) functionality will be a barrier to clients when switching from files -> API.

> Calls now have an id field.

-1 for performance reasons without hearing some very compelling argument. This is a good example of a case where we should have a separate PR which clearly explains the rationale.

> The variantId field has been removed from Calls. Calls are to be contained in Variants and cannot be retrieved independently.

Sounds reasonable.

> The phaseSet of a Call can now only be null or a string, not an array of strings.

This seems like another breaking change which needs justification. It may be fine - I just don't have the context.

macieksmuga:
@calbach the `phaseSet` of a `Call` being a null or string is a simple reversion to pre-graph behavior. Array of strings was specifically introduced to support graphs.

macieksmuga:
+1 modulo removing the `id` field from  `Call` .

adamnovak:
OK, I've removed IDs in `Call`s, since they didn't really make all that much sense, seeing as nothing else was using them. That's in #366 onto the refVar-linear branch, and should show up here when that's done with CI and can be merged.

Also Maciek is right about the phase sets; we added an array of strings as an option for a phase set so that we could talk about phasing by assigning individual instances of `Allele`s to the same molecules as other instances of `Allele`s. With `Allele`s gone, it makes no sense to have that kind of a phase set; it's sufficient to phase `Call`s as a whole with each other, like we were doing before the graph stuff went in.

I think the `Segment`-ization and removal of name references is worth doing, even if it breaks backward compatibility. Removal of `Path`, for example, meant I had to replace one location with a `Segment`, and then it seemed very obviously inconsistent to use `Segment`s to talk about some kinds of locations and groups of three fields to talk about others.

As for `Reference`s and `Sequence`s, a `Sequence` basically defines the string and coordinate space, while a `Reference` brings part of that sequence/coordinate space into one or more `ReferenceSet`s. A `Sequence` may share an ID with an associated `Reference`, but it is not required to. I don't think that that assertion means anything in practice, since any two objects of different types can share an ID.

lh3:
-1 on replacing (refName,start,end) with `Segment`. `Segment` has an extra strand field, which makes little sense in VCF. In addition, `Segment` is a complex object: Segment->Side->{Strand,Position}. I don't see we need this level of complexity if all we want is (name,start,end).

Re reference vs. sequence: it is also my impression that in the linear API, every sequence is a reference sequence. There was no "sequence" object in 0.5.1. This PR is changing our reference model which deserves more discussions. I prefer the v0.5.1 version.

lh3:
Some other small differences in comparison to v0.5.1:

 * v0.5.1 Reference doesn't have an isPrimary field.
 * v0.5.1 Variant.alternateBases and calls default to empty arrays.
 * v0.5.1 ReferenceSet doesn't have includedReferenceSets.
 * Position is defined differently: one with strand and the other without strand.

I have only briefly checked common.avdl, references.avdl and ariant.avdl. I could have missed some differences. As @calbach said, it would be good to make the new linear branch largely compatible with v0.5.1 unless there is a compelling reason against 0.5.1.

jeromekelleher:
I don't think our goal for this should be to revert entirely to 0.5.1, otherwise we lose essentially a year's worth of work. I think it's inevitable that we'll have some breaking changes (many of which have already been discussed, debated and voted on) when we get to the next release. However, these should be well defined and minimised where possible. 

To me, the point of this PR is to back out of the graph changes as simply as we can, and get a version of the schemas in master that we can iterate on to move towards the next release. Perhaps this PR isn't the place to make structural changes like adding Segments and changing reference models and so on. It would certainly be easier to review and discuss Segments if they were added in a follow up PR on master.

lh3:
My point is that 0.5.1 is an agreed and well discussed linear model, but the current model is not. When we have doubts, we should stay close to 0.5.1. On the differences between 0.5.1 and this branch, I see 0.5.1 is mostly better.

calbach:
> Phasesets

OK, great - I did not realize this was a graph by-product. Also I see that it removes a union type, so I'm definitely in favor.

> Segment-ization

I agree that we should be consistent. As @lh3 points out, the existing convention in 0.5.1 would be to **not** have segments or sequences. Note that I'm not saying we need to maintain 0.5.1 at all costs - I'm mainly concerned that Segments and Sequences were added to support the graph model (or the hybrid model) and I'm not convinced they are adding value in a linear-only version of the API. I also think we'll need to make a large number of breaking changes when we switch to the graph model down the road (I believe we're no longer planning to have an intermediate hybrid model), so I think it makes sense to favor usability of the linear API over setting ourselves up for future graph API changes. 

Per today's call I think the plan is to revert back further (to something closer to 0.5.1), and then propose the segment/sequence changes from a topic branch on top of that. I think this is more-or-less the suggestion in @jeromekelleher's comment.

Also thanks @adamnovak for putting the PR together.

adamnovak:
@lh3 and @calbach (and averyone else), have a look at #367. In there, I wind back to something a bit closer to 0.5.1, which is as close to what the server is currently implementing as I think I can get it with respect to the things that the graph changes touch. I do indeed remove `Segment`s and `Sequence`s, and go back to pointing to things by `Reference` ID. I also undid a bunch of small syntactic changes on the relevant objects that the server never implemented.

With this, the server shouldn't have to do anything else to get in sync with the schemas for the core references, reads, and variants APIs (although a strict compliance test will still probably turn up some bugs).

If we merge that into this PR, and then this PR into master, we can then go ahead and a topic branch for the reintroduction of `Sequence`s and `Segment`s, and the replacement of name references with guaranteed-to-be-unique ID references, and the other syntactic improvements that I think we can mostly agree are actually improvements---but this time with a corresponding server branch in which it is implemented, so the server doesn't get left behind again.

diekhans:
CH Albach <notifications@github.com> writes:
> I'd say I'm +0 on this. It's cleaner from a programming perspective, but
> removal of referenceName and callsetName (i.e. sampleId) functionality will be
> a barrier to clients when switching from files -> API.

I believe it's a far better approach to address simplifying use
of the API via have more sophisticated queries rather than have
a more complex data model.

For example, there would be a query that takes referenceName
converts to referenceId and return the reference sequence
without having to explicit store referenceName along with
referenceId.

adamnovak:
OK, @lh3 et al., come have a look at it now. I've pared the synatx back to what is currently supported by the server, modulo a couple of things that @jeromekelleher wanted me to leave in because the server is (presumably) about to adopt them.

Additional changelog:

* Removed `Segment`s.
* Replaced `Segment`s with triples of a `Reference` name, start, and length.
* Removed `Sequence`s, and all methods thereof.
* Changed references to `Sequence` IDs to use `Reference` names instead.
* Removed `Side`s.
* Added `Strand` back to `Position`.
* Removed `sequenceIds` and `referenceNames` from `SearchReferencesRequest`.
* Removed `isPrimary` from `Reference`.
* Removed `ReferenceSet` inheritance and the `includedReferenceSets` field.
* Replaced `SearchVariantsRequest`'s `name` with the old `variantName`.
* Added `callSetName` back to `Call`.
* Fixed a couple arrays to default to empty.

dglazer:
Thanks Adam. Do you know if there's an easy way to see diffs against 0.5.1?
I think we're going from San Francisco to San Jose by way of New York,
which makes it hard to review carefully. ( sorry if I missed it somewhere -
on a phone now.)

diekhans:
SourceTree is a usefu app to review and understand changes
  https://www.sourcetreeapp.com/

https://answers.atlassian.com/questions/26552/sourcetree-equivalent-of-git-diff-master-feature

lh3:
I was mainly worrying about `Segment` and the reference model. As these are closer now, I am +1 to this PR. The changes requested by @jeromekelleher in #367 are reasonable.

src/main/resources/avro/variantmethods.avdl
```diff
@@ -61,98 +61,19 @@ org.ga4gh.models.VariantSet getVariantSet(
...
   */
   string variantSetId;
 
-  /** Only return variants which have exactly this name (case-sensitive, exact match). */
-  union { null, string } name = null;
+  /**
+  Only return variants which have exactly this name (case-sensitive, exact
+  match).
+  */
+  union { null, string } variantName = null;
```
calbach: Was this revert intentional? The field name was changed in #239

src/main/resources/avro/variantmethods.avdl
```diff
@@ -161,21 +82,12 @@ record SearchVariantsRequest {
...
-  Only return variants with reference alleles on the reference with this name.
-  One of this field or `referenceId` is required.  (case-sensitive, exact match)
-  */
-  union { null, string } referenceName = null;
-
-  /**
-  Only return variants with reference alleles on the reference with this ID. One
-  of this field or `referenceName` is required.
-  */
-  union { null, string } referenceId = null;
```
calbach: No action required: would be good to add this field back in later.

src/main/resources/avro/variantmethods.avdl
```diff
@@ -243,91 +155,10 @@ org.ga4gh.models.Variant getVariant(
...
-org.ga4gh.models.Allele getAllele(
-  /**
-  The ID of the `Allele`.
-  */
-  string id) throws GAException;
-
 /******************  /callsets/search  *********************/
 /** This request maps to the body of `POST /callsets/search` as JSON. */
 record SearchCallSetsRequest {
-  /**
```
calbach: Revert?

src/main/resources/avro/variants.avdl
```diff
@@ -117,37 +104,27 @@ particular `Variant`.
...
   The number of results will also be the same.
   */
-  union { null, string } callSetName = null;
-
-  /**
-  The ID of the `Variant` that this `Call` belongs to. Must be set in any `Call`
-  that is not being returned from the server already contained within its
-  `Variant`.
-  */
-  union { null, string } variantId;
```
calbach: No action required: would be good to reapply this change later (removal of variantId + callsetId).

calbach:
+1. Please also take a look at a few minor inline comments I added. Thanks!

src/main/resources/avro/variantmethods.avdl
adamnovak: Yes. As far as I can tell, the server is currently set to parse (and the client library to produce) the `variantName` field.

src/main/resources/avro/variantmethods.avdl
adamnovak: Yes, that star should have stayed. I'll put it back.

adamnovak:
It seems like everyone is happy now. Are there any other comments? Did I miss something? Or does someone want to merge this?

adamnovak:
@dglazer Github will give you the diff here: https://github.com/ga4gh/schemas/compare/v0.5.1...refVar-linear

We should probably wait for your comment before merging this, actually.

dglazer:
+1.

Thanks @adamnovak for pulling this together, and for the link to diffs from 0.5.1. I didn't look at all the changes carefully, but I very much like the goal of getting our schemas and implementations back in sync, and this PR is a big step in the right direction. Notes below on some of the details.

* Changes to 0.5.1 Features (9 files)
  * I think these edits are the most important part of the cleanup, and they look good, especially since @jeromekelleher and @calbach, who know the most about existing implementations, are happy. More comments below, but nothing blocking.
  * reads.avdl: per discussion late in #259, I don't think the Fragment record adds any value as is (since without Fragment queries you can't use it for anything, and it's not clear Fragment queries are worth implementing). It's easy to ignore, so doesn't hurt much, but I'd be tempted to pull it out either as part of this general cleanup, or in a subsequent smaller PR.
  * beacon.avdl: it doesn't seem stylistically consistent with the rest of the schema, and I don't know if the reference server has support. But I haven't been following the Beacon conversation, so don't have a strong opinion.  

* Changes to 0.5.1 WIP Features (3 files)
  * I don't know enough to have an opinion here. These edits promote metadata support from WIP to top-level. I know there's been lots of work on metadata, but I haven't been following it closely -- does making it top-level imply that the reference server now supports metadata, and we think the objects/methods are stable enough to lean on? (In all cases we now have moved some common objects into /metadata, so should probably keep it as is for now, and then do a separate cleanup PR if needed.)

* New Features (9 files)
  * -0 for process reasons. The point of this PR is to get the reference server and schema back in sync, and I don't think the reference server has imminent plans to support the annotations or g2p features introduced here. (That's not a comment on the substance of those changes of course, just on the process we're moving towards. If others are comfortable, I don't object strongly.)

* Modified Infra (6 files)
  * +0 -- all seem like good changes, but I've only skimmed.

* New Infra (12 files)
  * +0 -- all seem like good changes, but I've only skimmed. (I was a little surprised to see 23,000 lines of presumably-generated UML here, but assuming people have discussed that, not a problem.)

 
#### Full list of changed files
##### Changes to 0.5.1 Features (9 files)
```
    175  src/main/resources/avro/beacon.avdl
     69  src/main/resources/avro/common.avdl
N    16  src/main/resources/avro/methods.avdl
    110  src/main/resources/avro/reads.avdl
    159  src/main/resources/avro/readmethods.avdl
     72  src/main/resources/avro/references.avdl
    126  src/main/resources/avro/referencemethods.avdl
     97  src/main/resources/avro/variants.avdl
    146  src/main/resources/avro/variantmethods.avdl
```
##### Changes to 0.5.1 WIP Features (3 files)
```
    198  src/main/resources/avro/[wip/]metadata.avdl
    118  src/main/resources/avro/[wip/]metadatamethods.avdl
X   189  src/main/resources/avro/wip/variationReference.avdl
```
##### New Features (9 files)
```
N   186  src/main/resources/avro/alleleAnnotations.avdl
N   182  src/main/resources/avro/alleleAnnotationmethods.avdl
N    86  src/main/resources/avro/annotateAllelemethods.avdl
N   176  src/main/resources/avro/genotypephenotype.avdl
N   144  src/main/resources/avro/genotypephenotypemethods.avdl
N    26  src/main/resources/avro/ontologies.avdl
N   103  src/main/resources/avro/sequenceAnnotationmethods.avdl
N   182  src/main/resources/avro/sequenceAnnotations.avdl
N   337  src/main/resources/avro/wip/matchmaker.avdl
```
##### Modified Infra (6 files)
```
      4  .gitignore
     11  .travis.yml
     24  CONTRIBUTING.md
     10  README.md
     32  GeneratingDocumentation.md → doc/GeneratingDocumentation.md
    171  pom.xml
```
##### New Infra (12 files)
```
N  1781  contrib/Doxyfile
N   124  contrib/avdlDoxyFilter.py
N    47  contrib/make_avrodoc.sh
N 14073  doc/uml.dia
N  9498  doc/uml.svg
N     6  requirements.txt
N    66  scripts/run_tests.py
N   132  scripts/utils.py
N   126  tests/compile_schemas.py
N    31  tests/test_maven.py
N    62  tests/test_protocol.py
N   129  tests/utils.py
```

PR #363, opened 2015-07-16
Further edits from Eagle, looking at clinical intervention/observatio…

eleanorstanley:
…n studies

A few suggestions to the schema using a different dataset.
I hope I have followed correct protocol: I corrected typos/grammar in comments; I wrapped text at 80 characters.
Thanks, Eleanor

### DIFF
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -464,7 +464,7 @@ protocol Metadata {
     Atlantic ocean, at Latitude 26.4, Longitude 69.3, Altitude -100m)
     - whole blood preparation of an individual
   */
-  record Biosample {
+  record BioSample {
 
     /**
     The biosample's id.


mbaudis:
```
[Eagle: Stage IV of what staging system? Before the staging system was defined
 but not the stage, now we have the stage but not the system. Am I missing 
something, are both not required? If you only want one term maybe the example
should be: TNM staging system, T2]
```
@eleanorstanley You seem to mis-read this a bit. It is

```
union { null, OntologyTerm } stageAtDiagnosis = null;
```

... where the relevant type/version etc. are defined inside the ```OntologyTerm``` object (which is a wrapper around ```ontologySourceID``` and additional information).

mbaudis:
@eleanorstanley I merged some of your notes (typos, examples) in the current metadata branch metadata.avdl.

diekhans:
Thanks @eleanorstanley!!.  We need to to start collecting data files to use a use cases for metadata to include in our
test suite.  Would you be able to donate data from the test cases that you are using?  We don't need lots, just examples of various cases.

eleanorstanley:
@diekhans I am unable to share the data file I used from sysVASC target discovery project (http://www.sysvasc.eu) outside the consortium, apologies. The data I used to look at the intervention/observation record update is from https://clinicaltrials.gov, I used random accession numbers and pulled them together to build an idea of the data captured, for example
Observational: https://clinicaltrials.gov/ct2/show/NCT00215969?term=NCT00215969&rank=1
Interventional: https://clinicaltrials.gov/ct2/show/NCT02367495?term=NCT02367495&rank=1

PR #367, opened 2015-07-22
Wind back graph syntax changes to what the current reference server implements

adamnovak:
This is an attempt, as discussed on the call this morning, to undo some of the syntax changes introduced alongside the graph features. I'm mostly targeting the schemas as they were at e6d6074 which @dcolligan recommended, but I'm keeping some of the newer features which the server does actually implement (like a couple of the get by ID methods, and the use of `Positions` instead of what we used to call `Side`s).

I've only touched a few of the files: the ones that adding graph support actually modified, and a couple of the other APIs that depended on the stuff I removed and had to be tweaked in order to compile. Those other APIs still need to be backed out onto topic branches; @macieksmuga is going to handle that after we finish backing out the graph features.

If we take this into master, I think this leaves the server more or less correctly implementing everything it actually does implement. There are some places where the spec specifies an array, and the server only supports a 1-element array. There are also still a few new features in the files I touched that the server may or may not have picked up on, mostly to do with the difference between full string and substring search. I didn't roll those back; the server will just have to change its semantics slightly after this is in, if it isn't doing the new semantics already.

Here's a changelog:

* Removed `Segment`s.
* Replaced `Segment`s with triples of a `Reference` name, start, and length.
* Removed `Sequence`s, and all methods thereof.
* Changed references to `Sequence` IDs to use `Reference` names instead.
* Removed `Side`s.
* Added `Strand` back to `Position`.
* Removed `sequenceIds` and `referenceNames` from `SearchReferencesRequest`.
* Removed `isPrimary` from `Reference`.
* Removed `ReferenceSet` inheritance and the `includedReferenceSets` field.
* Replaced `SearchVariantSetsRequest`'s `datasetId` with the old `datasetIds`.
* Replaced `SearchVariantsRequest`'s `variantSetId` with the old `variantSetIds`.
* Replaced `SearchVariantsRequest`'s `name` with the old `variantName`.
* Removed the unimplemented `/variants/<id>` method.
* Replaced `SearchCallSetsRequest`'s `variantSetId` with the old `variantSetIds`.
* Added `callSetName` back to `Call`.
* Fixed a couple arrays to default to empty.

### DIFF
--- a/src/main/resources/avro/variantmethods.avdl
+++ b/src/main/resources/avro/variantmethods.avdl
@@ -8,9 +8,10 @@ import idl "variants.avdl";
 /** This request maps to the body of `POST /variantsets/search` as JSON. */
 record SearchVariantSetsRequest {
   /**
-  The Dataset to search.
+  If specified, will restrict the query to variant sets within the
+  given datasets.
   */
-  string datasetId;
+  array<string> datasetIds = [];
 
   /**
   Specifies the maximum number of results to return in a single page.
@@ -64,16 +65,14 @@ org.ga4gh.models.VariantSet getVariantSet(
 /******************  /variants/search  *********************/
 /** This request maps to the body of `POST /variants/search` as JSON. */
 record SearchVariantsRequest {
-  /**
-  The VariantSet to search.
-  */
-  string variantSetId;
+  /** Required. The IDs of the variant sets to search over. */
+  array<string> variantSetIds = [];
 
   /**
   Only return variants which have exactly this name (case-sensitive, exact
   match).
   */
-  union { null, string } name = null;
+  union { null, string } variantName = null;
 
   /**
   Only return variant calls which belong to call sets with these IDs.
@@ -82,10 +81,23 @@ record SearchVariantsRequest {
   */
   union { null, array<string> } callSetIds = null;
 
+  /** Required. Only return variants on this reference. */
+  string referenceName;
+
   /**
-  Only return variants overlapping this `Segment`.
+  Required. The beginning of the window (0-based, inclusive) for
+  which overlapping variants should be returned.
+  Genomic positions are non-negative integers less than reference length.
+  Requests spanning the join of circular genomes are represented as
+  two requests one on each side of the join (position 0).
   */
-  org.ga4gh.models.Segment searchSegment;
+  long start;
+
+  /**
+  Required. The end of the window (0-based, exclusive) for which overlapping
+  variants should be returned.
+  */
+  long end;
 
   /**
   Specifies the maximum number of results to return in a single page.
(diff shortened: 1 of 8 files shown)


src/main/resources/avro/variantmethods.avdl
```diff
@@ -8,9 +8,10 @@ import idl "variants.avdl";
 /** This request maps to the body of `POST /variantsets/search` as JSON. */
 record SearchVariantSetsRequest {
   /**
-  The Dataset to search.
+  If specified, will restrict the query to variant sets within the
+  given datasets.
   */
-  string datasetId;
+  array<string> datasetIds = [];
```
jeromekelleher: Why are we rolling this back? We're already supporting exactly one dataset in the server, so the change required to move to `datasetId` from `datasetIds` is trivial. The changes required to properly support multiple datasets would be massive.

src/main/resources/avro/variantmethods.avdl
```diff
@@ -64,16 +65,14 @@ org.ga4gh.models.VariantSet getVariantSet(
 /******************  /variants/search  *********************/
 /** This request maps to the body of `POST /variants/search` as JSON. */
 record SearchVariantsRequest {
-  /**
-  The VariantSet to search.
-  */
-  string variantSetId;
+  /** Required. The IDs of the variant sets to search over. */
+  array<string> variantSetIds = [];
```
jeromekelleher: Same argument as variantSetIds above --- we only support one, so there's no point in reverting this.

src/main/resources/avro/variantmethods.avdl
```diff
@@ -131,24 +143,14 @@ SearchVariantsResponse searchVariants(
...
-
 /******************  /callsets/search  *********************/
 /** This request maps to the body of `POST /callsets/search` as JSON. */
 record SearchCallSetsRequest {
   /**
-  The VariantSet to search.
+  If specified, will restrict the query to call sets within the
+  given variant sets.
   */
-  string variantSetId;
```
jeromekelleher: Ditto here.

jeromekelleher:
Looks good to me @adamnovak. Other than the datasetIds/variantSetIds comments above (which I think are important) I'm +1.

src/main/resources/avro/variantmethods.avdl
adamnovak: Because the server expects to parse a one-element array, and not a single string. I'm trying to make the schemas look syntactically as much like the server expects as I can.

If you like, we could specify that the array has to be one element.

If you want to keep it as just a string, the server should parse just a string.

adamnovak:
@jeromekelleher If you want to switch the server to use single strings instead of arrays, I can leave the single strings in there and then we can merge this.

adamnovak:
How about this? I changed to the single string fields, and also restored the ability to get a variant by ID, which I think you asked for in your e-mail to me.

adamnovak:
Actually, Jerome is on vacation. Does someone else server-y want to review this? @dcolligan ? @macieksmuga ?

dcolligan:
+1

adamnovak:
Since there seems to be consensus from the server people, I'm merging this into #352 .

PR #371, opened 2015-07-30
Changed variantName back to name as in PR #239.

jeromekelleher:
This reverts a minor change made in #352. Changes to the reference server are trivial, and can be completed as part of a wider cleanup.

### DIFF
--- a/src/main/resources/avro/variantmethods.avdl
+++ b/src/main/resources/avro/variantmethods.avdl
@@ -73,7 +73,7 @@ record SearchVariantsRequest {
   Only return variants which have exactly this name (case-sensitive, exact
   match).
   */
-  union { null, string } variantName = null;
+  union { null, string } name = null;
 
   /**
   Only return variant calls which belong to call sets with these IDs.


dglazer:
Tentatively -1.

Looking back at #239, I don't see any discussion on why we changed the parameter name in the first place. (I understand and support the behavior changes in the comments.) From first principles I don't see a compelling reason to break backwards compatibility here. Am I missing something? @calbach , any thoughts (since I think you +1'ed #239)?

diekhans:
I think the really important thing is not backwards
compatibility; it's driving the whole schema to consistency.  We
are going to have to live with this for decades.

So the first step, what naming convention do we wan
to use:

   Variant.variantName
   Variant.variantId
or
   Variant.name
   Variant.id

While, personally, I tend to go for the less verbose, either way
is fine, as long as it is done consistently across all the record.

calbach:
I believe that when I had +1'd the original PR, I'd thought this was changing a field called `variant.variantName` to `variant.name`. I realize now that such a field [does not exist](https://github.com/ga4gh/schemas/blob/d2b3380992d920150c6d40f884c48cda0d50f321/src/main/resources/avro/variants.avdl#L183).

Looking again, this is actually changing the `SearchVariantsRequest`. This is less compelling because the semantics of this field are actually not consistent with our other `*Request.name` fields. Firstly, there are nested `Call`s within the `Variant` which could cause ambiguity (is this criteria matching calls or variants?), and secondly the actual field being matched is called `Variant.names`.  IIRC the field was named intentionally this way for the first reason. This inconsistency is a side-effect of the nested structure of variants+calls, which I think we're pretty committed to at this point and which I believe is cleaner in the graph branch of the API.

If this were a review of the initial commit of the field, I'd probably lean slightly towards `variantName` but could be convinced either way. Given that this is a breaking change to an existing field, I'm definitely -1. Apologies for my confusion on the previous PR.

-1

jeromekelleher:
OK, I tend to agree with @dglazer and @calbach.  I had a similar confusion to @calbach about the purpose of this name change originally. 

I'm going to go ahead and -1 my own PR, and close it unless @diekhans has some compelling contrary arguments.

PR #373, opened 2015-07-31
Baseline - latest core references/reads/variants schemas

macieksmuga:
The intention of this PR is to bring all three GA4GH projects, `schemas`, `server` and `compliance` in synch with each other by first adopting a common baseline, then re-adding the teams' contributions one-by-one to all three repositories simultaneously.


macieksmuga:
I can squash the individual commits that went into creating baseline, but I left them separate for now if somebody wants to examine how each individual team's code was separated (one team's contributions per commit).

I left `Experiment` and `Dataset` within the `metadata` module, as these are used in the baseline `reads`.  I also left `common` alone for now, with the idea that its constituent records could be refactored into other modules in another PR.

diekhans:
+1

skeenan:
+1

dglazer:
+1.

The comment on Dataset is now pretty out of date, but we can fix that separately. I wrote up suggested language to describe the original intent in https://github.com/ga4gh/schemas/pull/319#issuecomment-109291693 if you'd prefer to include it now.

jeromekelleher:
+1.

lh3:
+1

skeenan:
This has all the +1s we need. Merging.

PR #374, opened 2015-08-06
Remove the name argument from variant search.

jeromekelleher:
This PR removes the ``name`` parameter from the ``SearchVariants`` method. This is motivated by the ongoing attempt to sync the schema with what can be implemented effectively in the reference server. Since htslib does not provide this functionality, we cannot implement this query without filtering over potentially millions of variants, leading to very undesirable server behaviour.

During yesterdays refVar call it was decided that removing this functionality was a good option. The other option being to raise a HTTP 501 Not Implemented when a user attempts to search by name in the reference server.

### DIFF
--- a/src/main/resources/avro/variantmethods.avdl
+++ b/src/main/resources/avro/variantmethods.avdl
@@ -70,12 +70,6 @@ record SearchVariantsRequest {
   string variantSetId;
 
   /**
-  Only return variants which have exactly this name (case-sensitive, exact
-  match).
-  */
-  union { null, string } variantName = null;
-
-  /**
   Only return variant calls which belong to call sets with these IDs.
   If an empty array, returns variants without any call objects.
   If null, returns all variant calls.


sarahhunt:
+1
This option doesn't make much sense as it stands as the location still needs to be specified.

diekhans:
+1    It can always be restored if needed. Functionality should be motivated by use cases, which is not obvious here.

jeromekelleher:
@calbach, @dglazer, have you any thoughts on this?

dglazer:
No strong opinion from me -- we've already implemented this feature, so may
keep it, but that sort of optional superset functionality doesn't break
official compatibility, so isn't a big concern. (And as Mark said, we can
always add it back to the standard later.)

CH, any thoughts? The biggest issue would be if folks in the community are
already counting on search-by-name; I don't know of examples of that, but
can't be sure.

calbach:
+1. API semantics which requires secondary indexing over large collections like this should be given careful consideration and be motivated by use cases. There may be some use cases, but I doubt this is critical functionality for a minimal specification.

jeromekelleher:
Excellent; looks like we have agreement and enough +1s then. @diekhans, would you like to do the merging honours?

benedictpaten:
Merging

PR #376, opened 2015-08-11
Made a CallSet belong to one VariantSet.

jeromekelleher:
In line with recent  simplifications, this PR changes a CallSet such it refers back to exactly one variantSet. In the reference server, a CallSet is just some metadata about a column in some VCF files and so can only ever belong to one variant set.

### DIFF
--- a/src/main/resources/avro/variants.avdl
+++ b/src/main/resources/avro/variants.avdl
@@ -79,8 +79,8 @@ record CallSet {
   /** The sample this call set's data was generated from. */
   union { null, string } sampleId;
 
-  /** The IDs of the variant sets this call set has calls in. */
-  array<string> variantSetIds = [];
+  /** The ID of the variant set this call set has calls in. */
+  string variantSetId;
 
   /** The date this call set was created in milliseconds from the epoch. */
   union { null, long } created = null;


diekhans:
+1

sguthrie:
As #379 points out, it would be beneficial to migrate away from thinking about these schema in terms of VCF. My understanding is that a Callset is considered to be a called genome and a VariantSet is considered to be a group of Variants observed in one or multiple Callsets. If this is the case, we can easily conceive one might construct multiple VariantSets referring to the same Callset: let us say we make one VariantSet containing Callsets that were generated from females in a population and another VariantSet containing Callsets that were generated from cancer patients from the same population. Any Callsets corresponding to a female who has cancer will be in both VariantSets.

Though I have not been very active on this repository, I would vote -1. Apologies if voting is limited to those who make active pull requests only.

jeromekelleher:
Thanks for this @sguthrie, it's good to hear people's views on this. I guess my overriding question is this: what _is_ a CallSet? Is it simply a way of  representing some metadata about ``Call`` objects (which we don't want to copy into every ``Call`` because this would be wasteful), or does it represent something more sophisticated? If so, why does it need back-references to multiple variant sets? (But then, we don't seem to agree on what a variant set is either, so this is also hard.) 

Currently, the documentation says that a ``CallSet`` belongs to a ``VariantSet``, which would seem to contradict your interpretation above. I'm not saying either is correct, but we clearly have serious issues with these definitions right now.

lh3:
I think the correct data model should be:
```java
// record CallSet {} // drop this
record DatasetSample { // in common.avdl
  string id;
  string datasetId;
  string sampleName; // (datasetId,sampleName) unique across repo
}
record VariantSet {
//string datasetId; // drop this
  array<string> datasetSampleIds;
}
record Call {
//string callSetId, callSetName; // replace with the following:
  string datasetSampleId, sampleName;
}
record ReadGroup {
//string datasetId, sampleId; // replace with the following:
  string datasetSampleId;
}
```
This
* defines a global sample-related object and links reads and variants
* allows a VariantSet to include samples from multiple datasets (e.g. HRC and ExAC projects which both include 1000G as a subset)
* allows a (datasetId,sampleName) to be used in multiple VariantSets and keeps their track.

lh3:
PS: I have proposed this model twice before, but was largely ignored. Nonetheless, whenever we come back to samples and callSets due to ambiguity in their definitions, I think of this model again and believe it should solve most of our problems.

diekhans:
There needs to be the ability to support multiple sets of calls
on a single sample.  This would happen when running different
calling algorithms or different parameterizations of the same
algorithm.

Terry Speed found that the best approach to TCGA somatic variant
calling was to combine callers:
http://www.biomedcentral.com/1471-2105/15/154

jeromekelleher:
Thanks @lh3, this is very helpful. My initial impression of your proposal is that it's a big improvement over what we have now --- we explicitly link up variants and reads via a sample, which is what we need I think.

dglazer:
Per the discussion in https://github.com/ga4gh/schemas/pull/383#issuecomment-133586250, I agree with the original intent of this PR -- each `CallSet`, as currently defined and used, belongs to exactly one `VariantSet`. I'm temporarily -1 on merging this in, though, until the larger conversation here and in #383 resolves, since I don't want to make a small backwards-incompatible change until we know if we're going to make a larger change to the same area shortly.

jeromekelleher:
I agree with @dglazer. I'm going to close this PR, as I think the larger discussion in #383 is much more important.

PR #377, opened 2015-08-12
Simplified reference methods.

jeromekelleher:
This PR changes the reference and referenceSet search methods to accept single filter arguments rather than lists. The complex semantics of searching accessions is also reduced to simple string matching. This is part of a wider effort to simplify the schema so that it can be easily implemented and the semantics of each request is clear.

Performance arguments for allowing a client search for several references or referenceSets at once seem difficult to sustain. References are not something that a client needs to search for very often, so a few more server calls will make no difference in practice.

The semantics for searching over accessions depends on the server parsing accession numbers held within a string field. If this sort of functionality is required, then a formal model of the accession number object should be made, so that string parsing/construction is not required by either the client of the server.

### DIFF
--- a/src/main/resources/avro/referencemethods.avdl
+++ b/src/main/resources/avro/referencemethods.avdl
@@ -12,22 +12,20 @@ as JSON.
 */
 record SearchReferenceSetsRequest {
   /**
-  If nonempty, return the reference sets which match any of the given
-  `md5checksum`s. See `ReferenceSet::md5checksum` for details.
+  If not null, return the reference sets for which the
+  `md5checksum` matches this string (case-sensitive, exact match).
+  See `ReferenceSet::md5checksum` for details.
   */
-  array<string> md5checksums = [];
+  union { null, string } md5checksum = null;
 
   /**
-  If nonempty, return reference sets for which the accession
-  matches this string. Best to give a version number (e.g. `GCF_000001405.26`).
-  If only the main accession number is given then all records with
-  that main accession will be returned, whichever version.
-  Note that different versions will have different sequences.
+  If not null, return the reference sets for which the `accession`
+  matches this string (case-sensitive, exact match).
   */
-  array<string> accessions = [];
+  union { null, string } accession = null;
 
   /**
-  If not null, return reference sets for which the `assemblyId`
+  If not null, return the reference sets for which the `assemblyId`
   matches this string (case-sensitive, exact match).
   */
   union { null, string } assemblyId = null;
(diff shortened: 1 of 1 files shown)


diekhans:
+1  We really need to drive such optimization by compelling use cases.

src/main/resources/avro/referencemethods.avdl
```diff
@@ -12,19 +12,16 @@ as JSON.
...
+  union { null, string } md5checksum = null;
 
   /**
-  If nonempty, return reference sets for which the accession
-  matches this string. Best to give a version number (e.g. `GCF_000001405.26`).
-  If only the main accession number is given then all records with
-  that main accession will be returned, whichever version.
-  Note that different versions will have different sequences.
+  If not null, return reference sets for which the accession
+  matches this string.
```
dglazer: Does "matches this string" mean "is equal to this string" or "contains this substring" ? (I think I know the intent, but given that we're using the same phrase before and after this PR, I'm not sure.)

diekhans: This text implies that it is aware of accession version
numbering and knows how to parse it.  This is very reasonable.
Matching simple sub-strings, which was removed from the API, is
dangerous, as it could return spurious matches.

dglazer: @diekhans, are you describing the existing text, or the new text? I see a mention of version numbering in the existing text, but not in the new text. We should clarify the comment whether or not we end up deciding to change the behavior.

Two fyis:
- I was curious about the history of this part of the query interface -- it looks like it's unchanged since @richarddurbin first submitted it in #66 
- the current Google implementation does a "STARTS_WITH" for `accessions`. (And does an exact match for `md5checksums`.)

jeromekelleher: Good point @dglazer --- we're quite specific about this elsewhere, so we should make this string matching description uniform. I've pushed a commit addressing this (I'll squash the commits after if/when we've agreed they should go in).

dglazer:
Thanks for clarifying the comments @jeromekelleher. Adding back two fyi's that got lost because I put them on the (now outdated) diffs:
- I was curious about the history of this part of the query interface -- it looks like it's unchanged since @richarddurbin first submitted it in #66
- the current Google implementation does a "STARTS_WITH" for accessions, which is a simple way to support our understanding of the original desired functionality.

jeromekelleher:
After today's DWG call, I'd like to classify this as a bug (which means we want to resolve it before the next release). The consensus is we'd like to hear from @richarddurbin before anything else is done.

richarddurbin:
+1

jeromekelleher:
@dglazer, @calbach, any thoughts on this? I've updated the PR to squash the commits so it's ready to merge. This is a breaking change, but I hope one that will make the references API easier to use.

calbach:
+1, albeit reluctantly on account of the breaking changes. Agreed that it removes superfluous functionality.

dglazer:
+0

I'm fine with this simplification, since @richarddurbin is, but otherwise am not close enough to the use cases to assert that we're not removing important functionality.

jeromekelleher:
Looks like we have the requisite +1s --- anyone care to merge?

PR #378, opened 2015-08-12
Remove embedded references.

jeromekelleher:
Removed the embedded ReferenceId list from the ReferenceSet object. These can easily be obtained using the SearchReferences API call.

Embedding the IDs for child objects is not done anywhere else in the API (there are some where we embed the child _objects_, like `ReadGroupSet` and `Variant`).

### DIFF
--- a/src/main/resources/avro/references.avdl
+++ b/src/main/resources/avro/references.avdl
@@ -81,21 +81,12 @@ record ReferenceSet {
   string id;
 
   /**
-  The IDs of the `Reference` objects that are part of this set.
-
-  This field should be left null if the set contains a very large number of
-  `Reference`s. The `Reference`s in the set may also be obtained through a
-  `searchReferences()` API call.
-  */
-  union { null, array<string> } referenceIds = null;
-
-  /**
   Order-independent MD5 checksum which identifies this `ReferenceSet`.
 
   To compute this checksum, make a list of `Reference.md5checksum` for all
-  `Reference`s in this set, including any in any included set. Then sort that
-  list, and take the MD5 hash of all the strings concatenated together. Express
-  the hash as a lower-case hexadecimal string.
+  `Reference`s in this set. Then sort that list, and take the MD5 hash of
+  all the strings concatenated together. Express the hash as a lower-case
+  hexadecimal string.
   */
   string md5checksum;
 


src/main/resources/avro/references.avdl
```diff
@@ -81,21 +81,12 @@ record ReferenceSet {
   string id;
 
   /**
-  The IDs of the `Reference` objects that are part of this set.
-
-  This field should be left null if the set contains a very large number of
-  `Reference`s. The `Reference`s in the set may also be obtained through a
```
calbach: Perhaps this mention of SearchReferences could be moved to the record-level comment. It's not clean, since ReferenceSet doesn't really need to know about SearchReferences, but it could potentially save some time for clients. Alternatively, this could go into the protocol doc at the top.

calbach:
+1. This field is ugly and clients generally see it in our API and assume they should iterate and call GetReference(), which is quite slow. I believe we'd discussed (with Cassie) removal of this field over a year ago, but it either was forgotten or became obsolete in light of the graph changes.

src/main/resources/avro/references.avdl
jeromekelleher: I'm happy to move this comment if it makes things clearer. Is it really necessary though? We don't explain that we use ``SearchVariants`` to get a ``Variant`` in the header comments for ``VariantSet`` or anywhere else in the object hierarchy.

 I think end-users are generally going to be lost if they are trying to use these API docs to use the API. End users should be using libraries (with their own documentation). This documentation is really just for client-side library developers, I think.

calbach: On second thought you're right, it's not necessary. Reference Set -< Reference has a natural hierarchy.

Tangential to this PR:
'client' in my comment was referring to someone writing code against the API, in your view I guess this is a client library developer. If a client is going to be lost just looking at the docs that's probably a problem we should fix. I think depending on divergent sets of documentation in language-specific client libraries being better is not a good solution - in my opinion an ideal client library would be very thin and use as much autogenerated code from this schemas repo as possible (including docs).

diekhans:
+1

This is a good example of where we will have a simpler API if we do
these types of relations as queries rather than embedding it in the
data structures.

src/main/resources/avro/references.avdl
jeromekelleher: >  I think depending on divergent sets of documentation in language-specific client libraries being better is not a good solution - in my opinion an ideal client library would be very thin and use as much autogenerated code from this schemas repo as possible (including docs).

Yeah, you're right. I suppose what we're lacking is the 'big picture' documentation, but I think that's underway at UCSC.

I'd forgotten we'd talked about removing this ``referenceIds`` thing before: we should have made a PR on the spot!

diekhans: Jerome Kelleher <notifications@github.com> writes:
> Yeah, you're right. I suppose what we're lacking is the 'big picture'
> documentation, but I think that's underway at UCSC.

Underway, but delayed by
  (1) need to updated to match the new master
  (2) a European who seems to think August is for holidays

it will still be a work in progress

Issue #380, opened 2015-08-17
Formal definition of the data model needed


jeromekelleher:
Currently, the data model for container types (Datasets, VariantSets and ReadGroupSets) in the  GA4GH API is vague and ill defined. Where definitions are given, they refer back to the data file format on which they based. This is fine for structural purposes, but does little to clarify the semantic meaning of the collection. There are good arguments for allowing flexibility in the interpretations of data types, but there are also some major down sides:

1. **Interoperability** By leaving the interpretation of data types to server implementors, their methods of partitioning equivalent data among the various types will differ. Therefore, as a client side developer, if I write an analysis of (say) 1000G phase1 data on server implementation A, I may have to substantially rewrite this to run on implementation B.
2. **Content addressability** Unless the semantics of what a particular container holds is rigorously defined, it will not be possible to define a method for calculating a content hash (something that is recognised as being highly desirable for many reasons). It's clear how to do this for References, but it is an open problem for all the remaining types.

To start off this discussion I'd like to propose a straw-man model. I'm sure this is overly simplistic, but hopefully it'll be enough to get the discussion started, and we can iterate to something that is (a) flexible enough to be useful in a wide variety of scenarios; and (b) rigorous enough to give us a model we can reason with.

1. A ``ReferenceSet`` is a collection of References, and is defined as it currently is.

2. A ``Sample`` is an identifier associated with some biological observation. This might be an individual human tissue sample, or a bucket of sea water in a metagenomics study.

2. A ``DataSet`` is a collection of primary data for a defined set of ``Sample``s, and the inferences made using these data.

3. A ``DataSet`` contains a collection of ``ReadGroupSet``s. There may be many ``ReadGroupSet``s associated with a given sample (unmapped data, mapped to different references, different mappers/parameters).

4. A ``DataSet`` contains a collection of ``VariantSet``s. A ``VariantSet`` is a collection of ``Variant``s inferred from the data by some methodology.	

5. A ``Variant`` contains ``Calls``, which refer to a particular ``Sample``.

(I have omitted the ``CallSet`` here, as I've never managed to understand what it was, other than some metadata associated with a sample. )

The general idea is to try to encode the provenance of the inferences made on the primary data explicitly as part of the data model, and to explicitly link up the original samples to the variant calls. As I say, I'm sure this straw man is far too simplistic, but I hope it'll start some discussion from which we'll get a well-defined data model.

diekhans:
[...]

PR #383, opened 2015-08-19
removed CallSet; introduced NamedSample

lh3:
This PR removes the unnecessary CallSet object and introduces SampleGroup to keep track of cross-DataSet sample relationship. Note that ReadGroup.sampleName, VariantSet.sampleNames[] and DatasetSample.sampleName all refer to the same information. Different types of data in a Dataset can be related through sampleName. Data in different Datasets are related through SampleGroup.samples.

This PR may contain compiling errors and certainly needs to be polished. I will see the response of others first.

### DIFF
--- a/src/main/resources/avro/variants.avdl
+++ b/src/main/resources/avro/variants.avdl
@@ -41,7 +41,7 @@ record VariantSetMetadata {
 }
 
 /**
-`Variant` and `CallSet` both belong to a `VariantSet`.
+`Variant` belongs to a `VariantSet`.
 `VariantSet` belongs to a `Dataset`.
 The variant set is equivalent to a VCF file.
 */
@@ -58,43 +58,15 @@ record VariantSet {
   string referenceSetId;
 
   /**
+  List of samples that are involved in this variant set.
+  */
+  array<string> sampleNames = [];
+
+  /**
   The metadata associated with this variant set. This is equivalent to
   the VCF header information not already presented in first class fields.
   */
   array<VariantSetMetadata> metadata = [];
-}
-
-/**
-A `CallSet` is a collection of variant calls for a particular sample.
-It belongs to a `VariantSet`. This is equivalent to one column in VCF.
-*/
-record CallSet {
-
-  /** The call set ID. */
-  string id;
-
-  /** The call set name. */
-  union { null, string } name = null;
-
-  /** The sample this call set's data was generated from. */
-  union { null, string } sampleId;
-
-  /** The IDs of the variant sets this call set has calls in. */
-  array<string> variantSetIds = [];
-
-  /** The date this call set was created in milliseconds from the epoch. */
-  union { null, long } created = null;
-
-  /**
-  The time at which this call set was last updated in
-  milliseconds from the epoch.
-  */
-  union { null, long } updated = null;
-
-  /**
-  A map of additional call set information.
-  */
-  map<array<string>> info = {};
 }
 
 /**
@@ -110,21 +82,11 @@ record Call {
   /**
   The name of the call set this variant call belongs to.
   If this field is not present, the ordering of the call sets from a
-  `SearchCallSetsRequest` over this `VariantSet` is guaranteed to match
+  `SearchVariantsRequest` over this `VariantSet` is guaranteed to match
   the ordering of the calls on this `Variant`.
   The number of results will also be the same.
   */
-  union { null, string } callSetName = null;
-
-  /**
-  The ID of the call set this variant call belongs to.
-
-  If this field is not present, the ordering of the call sets from a
-  `SearchCallSetsRequest` over this `VariantSet` is guaranteed to match
-  the ordering of the calls on this `Variant`.
-  The number of results will also be the same.
-  */
-  union { null, string} callSetId = null;
+  union { null, string } sampleName = null;
 
   /**
   The genotype of this variant call.
--- a/src/main/resources/avro/variants.avdl
+++ b/src/main/resources/avro/variants.avdl
@@ -75,13 +75,13 @@ particular `Variant`.
 
 It may include associated information such as quality
 and phasing. For example, a call might assign a probability of 0.32 to
-the occurrence of a SNP named rs1234 in a call set with the name NA12345.
+the occurrence of a SNP named rs1234 in a sample with the name NA12345.
 */
 record Call {
 
   /**
-  The name of the call set this variant call belongs to.
-  If this field is not present, the ordering of the call sets from a
+  The name of the sample this variant call belongs to.
+  If this field is not present, the ordering of the names from a
   `SearchVariantsRequest` over this `VariantSet` is guaranteed to match
   the ordering of the calls on this `Variant`.
   The number of results will also be the same.
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -127,23 +127,15 @@ record Sample {
   string datasetId;
 
   /**
-  Sample name in the dataset. This name uniquely identifies the sample in the
-  dataset.
+  Submitted sample name in the dataset. This name uniquely identifies the
+  sample in the dataset.
   */
   string name;
 
   /**
-  The NCBI/EBI BioSample identifier of this sample. Different Sample objects
-  may have the same biosampleId.
+  Key-value pairs for additional sample information
   */
-  union { null, string } biosampleId = null;
-
-  /**
-  A globally unique ID, derived by some agreed upon scheme so that we can find
-  equivalent Sample's in different datasets and servers/implementations.
-  Typically, sequencing libraries and guids have a one-to-one relationship.
-  */
-  union { null, string } guid = null;
+  map<array<string>> info = {};
 }
 
 }
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -109,4 +109,44 @@ record Dataset {
 
 }
 
+/**
+In a Dataset, a sample is identified by its name that shall be unique in the
+Dataset. This name typically appears in a @RG line in a BAM file and in the
+sample header line of a VCF file. Unrelated samples in different Datasets may
+have the same name. DatasetSample is only meant to be embedded in SampleGroup,
+not referenced by other objects.
+*/
+record DatasetSample {
+
+  string datasetId;
+
+  /**
+  The sample name in the Dataset referred by datasetId
+  */
+  string sampleName;
+}
+
+/**
+A SampleGroup connects samples in different Datasets that have the same origin.
+A typical scenario is that samples from several Datasets are aggregated and
+reanalyzed, which result in a new Dataset.
+*/
+record SampleGroup {
+  /**
+  ID of this sample group
+  */
+  string id;
+
+  /**
+  The NCBI/EBI BioSample identifier of this sample group. Different samples
+  groups may have the same biosampleId.
+  */
+  string biosampleId;
+
+  /**
+  Samples in different Datasets which are considered the same
+  */
+  array<DatasetSample> samples;
+}
+
 }
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -110,12 +110,11 @@ record Dataset {
 }
 
 /**
-In a Dataset, a sample is identified by its name that shall be unique in the
-Dataset. This name typically appears in a @RG line in a BAM file and in the
-sample header line of a VCF file. Unrelated samples in different Datasets may
-have the same sample name.
+A sample uniquely named in a Dataset. The sample name typically appears in a
+@RG line in a BAM file and in the sample header line of a VCF file. Unrelated
+samples in different Datasets may have the same sample name.
 */
-record Sample {
+record NamedSample {
   /**
   Identifier of this sample
   */
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -113,40 +113,37 @@ record Dataset {
 In a Dataset, a sample is identified by its name that shall be unique in the
 Dataset. This name typically appears in a @RG line in a BAM file and in the
 sample header line of a VCF file. Unrelated samples in different Datasets may
-have the same name. DatasetSample is only meant to be embedded in SampleGroup,
-not referenced by other objects.
+have the same sample name.
 */
-record DatasetSample {
-
-  string datasetId;
-
+record Sample {
   /**
-  The sample name in the Dataset referred by datasetId
-  */
-  string sampleName;
-}
-
-/**
-A SampleGroup connects samples in different Datasets that have the same origin.
-A typical scenario is that samples from several Datasets are aggregated and
-reanalyzed, which result in a new Dataset.
-*/
-record SampleGroup {
-  /**
-  ID of this sample group
+  Identifier of this sample
   */
   string id;
 
   /**
-  The NCBI/EBI BioSample identifier of this sample group. Different samples
-  groups may have the same biosampleId.
+  The identifier of the dataset this sample belongs to
   */
-  string biosampleId;
+  string datasetId;
 
   /**
-  Samples in different Datasets which are considered the same
+  Sample name in the dataset. This name uniquely identifies the sample in the
+  dataset.
   */
-  array<DatasetSample> samples;
+  string name;
+
+  /**
+  The NCBI/EBI BioSample identifier of this sample. Different Sample objects
+  may have the same biosampleId.
+  */
+  union { null, string } biosampleId = null;
+
+  /**
+  A globally unique ID, derived by some agreed upon scheme so that we can find
+  equivalent Sample's in different datasets and servers/implementations.
+  Typically, sequencing libraries and guids have a one-to-one relationship.
+  */
+  union { null, string } guid = null;
 }
 
 }
--- a/src/main/resources/avro/reads.avdl
+++ b/src/main/resources/avro/reads.avdl
@@ -77,7 +77,7 @@ record ReadGroup {
   union { null, string } description = null;
 
   /** The sample this read group's data was generated from. */
-  union { null, string } sampleId;
+  union { null, string } sampleName;
 
   /** The experiment used to generate this read group. */
   union { null, Experiment } experiment;
--- a/src/main/resources/avro/variantmethods.avdl
+++ b/src/main/resources/avro/variantmethods.avdl
@@ -70,11 +70,11 @@ record SearchVariantsRequest {
   string variantSetId;
 
   /**
-  Only return variant calls which belong to call sets with these IDs.
+  Only return variant calls which belong to samples with these IDs.
   If an empty array, returns variants without any call objects.
   If null, returns all variant calls.
   */
-  union { null, array<string> } callSetIds = null;
+  union { null, array<string> } sampleNames = null;
 
   /** Required. Only return variants on this reference. */
   string referenceName;
@@ -112,8 +112,8 @@ record SearchVariantsRequest {
 record SearchVariantsResponse {
   /**
   The list of matching variants.
-  If the `callSetId` field on the returned calls is not present,
-  the ordering of the call sets from a `SearchCallSetsRequest`
+  If the `sampleName` field on the returned calls is not present,
+  the ordering of the samples from a `SearchVariantsRequest`
   over the parent `VariantSet` is guaranteed to match the ordering
   of the calls on each `Variant`. The number of results will also be
   the same.
@@ -148,68 +148,3 @@ org.ga4gh.models.Variant getVariant(
   The ID of the `Variant`.
   */
   string id) throws GAException;
-
-/******************  /callsets/search  *********************/
-/** This request maps to the body of `POST /callsets/search` as JSON. */
-record SearchCallSetsRequest {
-  /**
-  The VariantSet to search.
-  */
-  string variantSetId;
-
-  /**
-  Only return call sets with this name (case-sensitive, exact match).
-  */
-  union { null, string } name = null;
-
-  // TODO: Add more ways to search by other metadata
-
-  /**
-  Specifies the maximum number of results to return in a single page.
-  If unspecified, a system default will be used.
-  */
-  union { null, int } pageSize = null;
-
-  /**
-  The continuation token, which is used to page through large result sets.
-  To get the next page of results, set this parameter to the value of
-  `nextPageToken` from the previous response.
-  */
-  union { null, string } pageToken = null;
-}
-
-/** This is the response from `POST /callsets/search` expressed as JSON. */
-record SearchCallSetsResponse {
-  /** The list of matching call sets. */
-  array<org.ga4gh.models.CallSet> callSets = [];
-
-  /**
-  The continuation token, which is used to page through large result sets.
-  Provide this value in a subsequent request to return the next page of
-  results. This field will be empty if there aren't any additional results.
-  */
-  union { null, string } nextPageToken = null;
-}
-
-/**
-Gets a list of `CallSet` matching the search criteria.
-
-`POST /callsets/search` must accept a JSON version of `SearchCallSetsRequest`
-as the post body and will return a JSON version of `SearchCallSetsResponse`.
-*/
-SearchCallSetsResponse searchCallSets(
-  /** This request maps to the body of `POST /callsets/search` as JSON. */
-  SearchCallSetsRequest request) throws GAException;
-
-/****************  /callsets/{id}  *******************/
-/**
-Gets a `CallSet` by ID.
-`GET /callsets/{id}` will return a JSON version of `CallSet`.
-*/
-org.ga4gh.models.CallSet getCallSet(
-  /**
-  The ID of the `CallSet`.
-  */
-  string id) throws GAException;
-
-}


jjfarrell:
This revised schema better addresses  the typical use cases I see. 

There is one additional update required in a comment since CallSet was dropped:

record Call {

  /**
  The name of the call set this variant call belongs to.
  If this field is not present, the ordering of the call sets from a
  `SearchVariantsRequest` over this `VariantSet` is guaranteed to match
  the ordering of the calls on this `Variant`.
  The number of results will also be the same.
  */
union { null, string } sampleName = null;

I think should become:

record Call {

  /**
  The name of the sample this variant call belongs to.
  If this field is not present, the ordering of the calls from a
  `SearchVariantsRequest` over this `VariantSet` is guaranteed to match
  the ordering of the SampleNames in VariantSet 
  The number of Calls will also be the same.
  */
union { null, string } sampleName = null;

dglazer:
A few thoughts:
- We clearly need a way to talk about, and filter for, all the calls in a variantset that come from a particular analysis of a particular sample -- today, that's handled by the `Callset` object. If I read this PR right, it replaces that formal object with a `sampleName` string. If we're going to make changes, I'd rather go the other way, and add semantics/functionality to the object, not remove it.
- Among other things, I'm uncomfortable forcing users to manage their sample names as permanent unique IDs. (As opposed to today's model where the server can enforce that.)
- I agree that adding good "sample" support to the API, including a standard way to say "this callset is derived from this readgroupset", will be an improvement. The little I've heard about requirements, though, say that it's not as simple as just identifying the originating tissue sample, since there could (e.g.) be multiple different callsets all derived from one readgroupset, using different analysis tools and paramaters.

I suggest that we first agree on how we want to deal with samples in the API. (I know the metadata team has thought about that, but I don't remember if they converged on a proposal.) Once that's done, we can figure out how that affects today's schema.

P.S. I re-skimmed #49 (VCF-like variants proposal, May 2014) for the original thinking behind the current API. One interesting comment:
> Eventually I would hope there would be an actual GASample, which had an array of readgroupIds and an array of callsetIds. I expect as part of ga4gh we'll decide how a sample is defined and who is going to keep track of samples - but that's more like the metadata team.

Sounds like we may be getting close to "eventually" :)

lh3:
> I'm uncomfortable forcing users to manage their sample names as permanent unique IDs.

Sample names being unique in a dataset is a fundamental requirement of the projects I am aware of. This is how a project pins BAM, VCF and other sample information together. The role of sample name is simple and effective. CallSet complicates the simple role with an extra ID, a confusing name "call set", a non-existing sampleId and a dubious reference to variantSets. In the end, we are not clear how to get the reads corresponding to a sample in a VariantSet, which is rarely an issue in my daily works.

> I'd rather go the other way, and add semantics/functionality to the object, not remove it.

The current data model of CallSet is overly complicated. We need to simplify it rather than make it more complex.

> I suggest that we first agree on how we want to deal with samples in the API. (I know the metadata team has thought about that, but I don't remember if they converged on a proposal.)

The concept of "sample" goes two extremes. It can be very complex and can be very simple. On one hand, to precisely define a biological sample is very difficult in general. For example, do different tissues belong to the same sample? What about different cells? Different time points? This is an unsolved problem (you can look up NA12878 in the BioSample database to get a sense). My opinion is we should just reference BioSample instead of redefining BioSample in GA4GH.

On the other hand, within the scope of a project, a sample is very simple. It is just a unique string key in BAM, VCF or an Excel table. Two keys may correspond to two humans or two tissues of the same human. It is submitters' decision. The server doesn't care. This simple treatment solves the majority of our common problems and is how we work with sequencing data on daily bases.
<hr>
At the end of day, I would encourage people to think of practical use cases that can be achieved with CallSet but not without it. In addition, I could add an `id` field to `DatasetSample` if more prefer a server-maintained ID.

jeromekelleher:
I prefer  this to the current ``CallSet`` idea --- it just seems like a much more direct model of the data. I'm not quite sure of some of the details, but I like the basic idea of factoring out the (as far as I can tell) redundant ``CallSet`` concept, and moving to directly associating samples with VariantSets and ReadGroupSets.

On the details @lh3, why the split into the ``DataSetSample`` and ``SampleGroup`` objects? Perhaps something like this would be simpler?
```
record Sample {
    /* The ID of this sample. Unique within the repository. */
    string id; 
    /* The name of this sample. == Heng's DataSetSample.sampleName */
    string name;
    /* The dataset this sample belongs to */
    string dataSetId;
    /* == Heng's SampleGroup.bioSampleId */
   string bioSampleId = null; 
   /* A globally unique ID, derived by some agreed upon scheme so that we can 
      find equivalent Sample's in different datasets and servers/implementations
    */
   string guid;
    /* Plus, a bunch of other fields encoding metadata that we'll want to know about a sample */
}
```

The distinction between ``name`` and ``id`` here follows the growing consensus: ``id``s are server defined strings used to internally identify an object within a server instance, and names are user assigned identifiers, unique within a container. There is duplication of the metadata across different datasets I suppose, but if we have a GUID then perhaps this doesn't matter. A server is free to keep a single copy of each sample object, and keep references to this in the relevant data sets.

A ``VariantSet`` then contains calls for a particular sample ID or it does not. A ReadGroupSet contains reads for a particular sample ID or it does not. We can have as many ``VariantSet``s as we like within a DataSet, allowing for different combinations of samples, being called from different reads, using different algorithms, or what have you. This seems like a much simpler model than the current VariantSet/CallSet schema.

macieksmuga:
I'm not sure that this part of @dglazer's response was sufficiently addressed above:

> there could (e.g.) be multiple different callsets all derived from one readgroupset, 
> using different analysis tools and parameters.

This isn't about the sample or tissue that a set of calls came from, it's about the software processing provenance of those calls. For a single experiment this may not be an issue, but it's important in the case of complex genome-level projects, where the same sample data is continually re-processed against newer references and variant calling algorithms.

Lacking the notion of a `CallSet`, I assume the processing toolchain for a given call would be derived from the metadata of its containing `Variantset`?

lh3:
@jeromekelleher I changed to your model. There remain several unresolved issues:

1. I guess Sample::guid has a one-to-one relationship with BioSample if the latter is available. Then a guid corresponds to a sequencing library (right?). Ideally, we need to connect libraries constructed for the same individual. For example, CHM1 has been sequenced on both Illumina and PacBio machines. It is not obvious to me how we can know this from the BioSample database. Nonetheless, as we don't have a satisfactory solution yet, we may leave this issue to the experts in this field. (As I remember, BioSample has a tree-like hierarchy, but I don't see this hierarchy any more).

2. The scope of guid. I guess it is hard to achieve in practice. Another option is to make guid unique in a repo/server. We trust biosampleId to keep track of world-wide relationship.

3. Now Sample has an ID field, but I have not changed sampleName to sampleId in other places yet. I will explain what will happen if we do that. Endusers primarily interact with names rather than IDs. Then in `Call` and `ReadGroup`, we may need both sampleId and sampleName, which seems clumsy. This also adds extra constraints: sampleId must be consistent with sampleName and datasetId. Constraints lead to complications and potential inconsistencies. I still prefer to use sampleName. Nonetheless, if others are happier with sampleId, I can make changes. Using ID or name is not a deal breaker.

There are also a few less controversial issues. I need to kill compiling errors and add methods to search for samples. I can do that once we decide we want to merge this PR in principle.

@macieksmuga I believe @jeromekelleher has addressed that in the last paragraph, which also explains my intention. In a Dataset, you are allowed to have multiple VariantSets made from the same set/subset of BAMs (this should be discouraged for the best, but that is another topic). Their relationships can be trivially identified via sampleId or sampleName. I am not sure how `CallSet` works in case of multiple reanalyses. It is only making thing worse in my view.

macieksmuga:
@lh3 thanks for the additional clarification – @jeromekelleher I inadverently glazed over that part of your comment – that's pretty much the interpretation I was aiming for as well.

macieksmuga:
@jeromekelleher are we planning to pull this in before or after the schemas/server/compliance baseline task?

diekhans:
This really hints at the fundamental problem: we are designing
schemas for a data model that isn't defined.

I can find no documentation on what a sample even means to
GA4GH.  As @lh3 points out, samples are not sequenced, libraries
are.  They can be sequenced multiple times, with multiple
techniques.  A given sequencing run can be mapped to multiple
assemblies using multiple approaches.  Multiple variation calling
algorithms maybe run.  It's a tree at best.

While this sounds complex, it's reality.  If the data model
doesn't reflect reality, the use of the data model ends up being
even more complex.  One has to inconsistently hack around the
inadequate data model to use it.

BioSample doesn't solve the identification problem unless there
is a requirement that GA4GH APIs only manage data in the
BioSample databases.

Using UUIDs to assign GUIDs is very straight forward and can be done by
the data creator, not a central authority.  The probability of UUID's
colliding is far less that the data center being destroyed by a meteorite:

  https://en.wikipedia.org/wiki/Universally_unique_identifier#Random_UUID_probability_of_duplicates

I had always believed GA4GH `id' was a machine readable, unique
identifier and name being a more human readable name.  However,
this isn't defined by GA4GH, so my belief is a meaningless
fantasy.

We need to have a documented, formal definition of the data
model before defining schemas.  See:

         https://github.com/ga4gh/schemas/issues/380

jeromekelleher:
Thanks @lh3, this looks great! I think I'd favour using IDs as the linkage keys throughout the API, if only for consistency. I agree that users will primarily interact with names, as these are the more human friendly identifiers. To get the IDs for the names they want they could run a ``SearchSamplesRequest`` at the start of their program. I guess they'd have to do this anyway if they were using GUIDs or BioSamples rather than names, or were searching for samples that fitted some criteria (``population == "CEU"``, etc?).

I wouldn't argue strongly either way though, I don't think it's a big deal. I'm very interested to hear what others think of this schema now! There are some outstanding issues (GUIDs, mainly I guess), and it would be great to hear people's thoughts.

@macieksmuga --- it's too early to think pulling this change in yet in my view. It's a big change, and would have to be considered carefully from a compatibility perspective by the various stakeholders.

diekhans:
IMHO: no schema changes without a formal data model.

jeromekelleher:
I don't think we're that far from a data model if we can decide what a sample is @diekhans. What's your view of what a dataset should be @lh3? Is it a collection of reads and variants that are related to each other and some referenceSets(s) in a defined way (or definable, at least), or a bag of anonymous stuff? Or something in the middle?

mbaudis:
I second @diekhans ' opinions about user/machine assigned uuids ans their potential for object linkage. This has been discussed before and seemsnto be consensus; but as long as there is no formal definition of id/name/uuid/accession/versionid, it will come up again and again...

IMHO we should define a set of different id formats, with examples and proposed usage; e.g. 

*objectId - globally unique, collision-free UUID, e.g. v4
*localId - whatever locally unique id is suitable for legacy/readability reasons
*versionId - (content dependent/hashed) id of this object instance
*accession - URI
*name - whatever this object is called; mostly there to distinguish from machine-friendly ids

The "required/single/list" discussions can then be referred to the definition of the objects themselves.

diekhans:
You might be right that we are not far from a data model.  However,
if the data model isn't formally and unambiguously documented, it
might was well not exist.

"Looks like this in VCF" isn't a data model.  A useful analogy
for someone using it, but not a data model.

We need drawings, not Avro.

lh3:
First of all, this PR is only meant to show a sketch of the changes in my head such that we have something a bit concrete to discuss. @jeromekelleher is right that a lot need to be done to make it a real PR.

In my view, a dataset is a collection of data obtained from a fixed set of samples. What samples are included in a dataset is solely defined by its submitter with a primary constraint: samples shall have names unique in the dataset. We allow different types of sample-related data in a dataset (NB: this is different from an SRA Study), which are connected through the submitted sample names. For now, GA4GH only considers two types of data: reads and variants. We may have more in principle (e.g. expression). References are not part of a specific dataset.

I think within a dataset, our data model is clear and sound: it is the BAM and VCF model. If people are not happy, we just describe what a BAM/VCF model is. We may argue the BAM/VCF model is not the prefer model going forward (I for one want to replace VCF with a binary/unary allele model), but it is the model the majority of existing projects as well as our file-based reference server are built around. We need a major overhaul if we want to ditch the BAM/VCF model. Personally I don't mind either way, but realistically do we have the resources?

@jeromekelleher was asking how ReadGroup, ReadGroupSet and VariantSet are defined. It is naively simple in practice: they, like dataset, are subjectively defined by submitters (a ReadGroup is a @RG line; a ReadGroupSet is a full BAM and a VariantSet is a full VCF derived from full BAMs). I don't think it is possible impose objective universal definitions on these container types. We can only suggest conventions for common use cases.

> I can find no documentation on what a sample even means to GA4GH.

In the new schema, we can define `Sample` to have the same scope as a BioSample in NCBI/EBI. Any objections?

> Using UUIDs to assign GUIDs is very straight forward and can be done by the data creator, not a central authority.

I never oppose the idea of UUIDs, but how to compute it is worth a whole lot more discussions. I don't want these to distract this PR. It should be done elsewhere. This PR mainly aims to clearly connect reads, variants and samples within the scope of a dataset.

> However, if the data model isn't formally and unambiguously documented, it might was well not exist.

Could you show an example on how to "formally and unambiguously" document the data model in your mind? You are talking about drawings. Could you show such a drawing?

diekhans:
Our data model is not documented and thus not clear.  Being like
VCF is not good enough.  VCF can and does get used in a lot of
different ways.  If you want to define a interoperable API, you
have to have formal precision.

If we don't have the resource to define what we are doing, what
is the point?  We can already access BAM and VCF randomly over
the internet.

diekhans:
p.s. If there is formal documentation of the BAM/VCF data model,
then that is a starting point, but all I ever seen is file
format definitions.

dglazer:
@jeromekelleher -- I agree that
>  it's too early to think pulling this change in yet in my view. It's a big change, and would have to be considered carefully from a compatibility perspective by the various stakeholders

@lh3, thanks for driving this conversation, and for pushing us towards a better representation of samples in the API. I see several related but separate changes being discussed in this PR. First, things that the PR does **not** change:
* we want an object representing a 'column' in a variantset (i.e. all of the calls that belong together)
* the object should have a server-assigned id, and a user-assigned name
* the object should reference its container(s)
* the object should contain an array of other miscellaneous metadata

The proposed model-level changes I see are:
* make the mapping from object to sample 1:1, instead of letting multiple objects all reference the same `sampleId`
* add a `biosampleId`
* add a `guid`, details TBD
* remove the Get method to retrieve an object (although maybe I'm missing something, since I'm not sure how else we'd go from an object name to the object itself) 

The proposed syntax-level changes I see are:
* change the object name from `CallSet` to `Sample`
* remove the Search method for getting a list of the objects in a `VariantSet`; instead add an array of included object names to a `VariantSet`
* remove the `created` and `updated` fields
* change the container reference from `variantSetIds` to `datasetid`
 
If that summary is right, we can then talk about the model-level changes first, and see which of them we collectively like. Then, once we're in sync on the model we want, we can resolve the syntax-level changes.

P.S. For my reference, I put together a rough summary of the before&after object definitions.
The current schema defines a CallSet as:
```
record CallSet {		
  string id;  /* id of this object */	
  string name; /* name of this object */
  string sampleId; /* sample this call set's data was generated from */
  array<string> variantSetIds; /* IDs of the variant sets this object has calls in */	
  long created; /* when this object was created */
  long updated; /* when this object was last updated */
  map<array<string>> info; /* misc additional info */		
  }
```
The proposed schema defines a Sample as:
```
record Sample {
  string id; /* id of this object */
  string datasetId; /* dataset this object belongs to */
  string name; /* Sample name in the dataset (must be unique within the dataset) */
  string biosampleId; /* NCBI/EBI BioSample identifier of this sample. Different Sample objects may have the same biosampleId. */
  string guid; /* a globally unique ID; details TBD */
  /* Plus, a bunch of other fields encoding metadata that we'll want to know about a sample */
}
```

helenp:
Could someone from metadata join a future call and discuss sample/sample group please? Thanks Helen

diekhans:
I realize the problem here is that the thing being called a `sample' by
reads/variants isn't a sample.  It's something derived from a sample.

If the goal is to defined a data exchange API for the world's genomic
data, we have to use precise, defined terminology.

diekhans:
Hi @dglazer,

>   • we want an object representing a 'column' in a variantset (i.e. all of the
>     calls that belong together)

What is the definition of `belong together'?   Sorry, that is a rhetorical question ;-)
Here is all one has to go one:

    A `CallSet` is a collection of variant calls for a particular sample.
    It belongs to a `VariantSet`. This is equivalent to one column in VCF.

The problem is the semantics of the terminology and objects are
not rigorously defined.  The thing called a `sample' is not a
sample.  CallSet doesn't define what happens when you have
multiple collections on the same samples.  No where is what an
id or name defined.

The people creating the system don't even have a common
understand of what we are building.  How can we possibly produce
am interoperable API with interoperable data?

The reads group needs to reconvene and write rigorous
documentation.

We are going to do a PR later this month with a integrated
documentation system.  We need to fill in out with a normative
specification.

mlin:
If I may make an observation, I think some unnecessary confusion/distraction is being caused by this PR's inclusion of a new Sample data record with potentially controversial new metadata fields like GUID and biosampleId. The elimination of CallSet can be executed without atomically also defining this new Sample type (worthy though that would be). 

Here is a diagram I believe originally produced by @adamnovak and colleagues. 



A VariantSet is a two-dimensional tabular structure. A Call is a cell in the table. A Variant is a row of the table plus some critical info shared within the row (e.g. alleles). A CallSet is a column of the table, perhaps with some shared info, but not much. Variants and CallSets are both first-class objects with unique IDs and API methods to manipulate (or at least describe) them. Lastly, CallSets may encompass related columns of multiple tables (notwithstanding #376).

I think the proposal here is that there is no clear need for CallSet to be a first-class object with these bells and whistles. Instead we can just have a name for the column of the table. In the fullness of time it may be useful for this name to be a foreign key into other bits of the API where you can get more info or tie together related columns, but perhaps this isn't a must-have for the initial simplification. It is only critical that the user have control over the name while obeying workable uniqueness constraints.

Personally I'm +0 on killing CallSet now that I understand what it is (which occurred in the last few hours, to be clear). If killing it simplifies things and causes less confusion that would be good, though I think it could perhaps be rescued by a better name, and #376 in the absence of an immediate use case for spanning multiple VariantSets.

I'll also refer to the discussion in #380 which mentioned that CallSet might be more useful if the data were mutable via the API - but, there are some deeper problems with adding a new column to the table (the so-called "N+1 problem").

lh3:
> change the container reference from variantSetIds to datasetid

This is the single most important model change in this PR. Now, `Sample` is above reads and variants, not tied to variants. This connects different data types. In addition, it helps to make it explicit that sample names have a scope - within a dataset - and ought to be unique in this scope.

As to other model changes you mentioned, `sampleId` in the old schema doesn't reference anything. It is a dead link. `biosampleId` and `guid` are the same thing. They essentially take the place of the old `sampleId` as a cross-dataset identifier for samples. There will be a method to search for samples. It is an easy part once we reach a consensus on the model.

dglazer:
@diekhans and @helenp  -- yes, the heart of the issue here is the relationship between  "all of the calls that belong together in a variantset" and "sample".

Our API today takes a pragmatic approach -- it doesn't provide a semantic definition of "belong together", and instead provides a simple operational definition -- whoever loads data into the system gets to decide. As @lh3 said,
> It is naively simple in practice: [our groups of objects] ... are subjectively defined by submitters

I think we can, in parallel:

1. continue to improve our documentation, to capture a shared understanding of the current pragmatic approach
2. work with the Metadata TT to decide on how and when we want to introduce a more principled approach. (My *guess* is we'll make the principled approach optional, so data owners can continue to use whatever ad hoc approach they want, but will get benefits from adopting a standard model. We'll see.)

P.S. When I first started in this field, I thought it would be simple -- one vial of tissue == one sample == one readgroupset == one callset. I was naive.

dglazer:
@mlin -- yes, your picture matches my mental model of the data. (And the picture we use in http://ga4gh.org/#/documentation, although the copy there could use some small updates.) I also agree that much of the confusion today is coming from the two points you cite -- the name of the object, and the "one or many VariantSets" question.

@lh3 -- glad we're zeroing in on the central topic here -- the relationship between a "column of calls" and some object that spans (and therefore can join) multiple readgroupsets and variantsets.
* I think our current API is 90% pragmatic, but with a couple of small nods towards spanning functionality. In practice I believe the data is almost 100% pragmatic -- existing CallSets only belong to one VariantSet (which jibes with #376), and any data that uses the `sampleId` field uses it as a simple name, not as a link.
* I think doing more requires us to agree on:
  * what scope of joining do we want - per dataset? per server instance? global? 
  * what 1:1 vs. 1:many relationships do we want -- e.g. can one readgroupset be the basis for multiple callsets?

lh3:
No, we need an explicit global object for sample such that both reads and variants can reference. And once we have a global object, CallSet is made redundant.

@dglazer what's wrong with this PR? You and others have not really talked about problems. Yes, guid is a loose end, but the old schema is more obscure at this point.

lh3:
> any data that uses the sampleId field uses it as a simple name, not as a link.

Are you suggesting `sampleId` is a name, not an ID? Anyway, I will show why the lack of a global sample object is problematic.

As of now, sample information is tied to CallSet. What if a dataset only has reads but no variants? The users can't submit any sample information in this case. When we have both reads and variants, we always need to get sample information from `CallSet`, a variant object. This seems conceptually wrong. Now if we make `CallSet` reference a single `VariantSet`, how do we represent two variant sets called from the same BAMs (this is discouraged but may happen)? I guess we need to create two `CallSet` for the same sample and duplicate the sample information? Apparently, the new `Sample` is `CallSet` moved up a level, but this is a philosophical change that simples our model and solves all these problems.

dglazer:
@lh3 , in case I wasn't clear -- I like the core idea behind this PR, of introducing *"some object that spans (and therefore can join) multiple readgroupsets and variantsets"*. But I want to be sure we're all clear on the semantics of that object (as @diekhans would say, the data model), including the people from the Metadata team who have thought about Samples, so whatever we come up with sticks.

Here are some examples of specific features I don't see fully addressed in the PR as it stands. We don't have to support all of them, but that should be a conscious choice. (These mostly come from the list of differences I posted earlier.)
- can one readgroupset be the basis for multiple callsets? If so, where do we store the metadata associated with the callset (e.g. the toolchain was used to generate it)
- can one tissue sample be the basis for multiple readgroupsets? If so, do we want to represent that in the API?
- are we trying to add biological meaning to a dataset, in addition to today's administrative and organizational meaning? (I don't think we should, but that's a separate discussion).
- do we want to tie into a more global data model, either existing (e.g. BioSample) and/or new (e.g. some sort of guid)? Is that required for all data, optional for some data, or overkill? 
- given a sample name (or id), how do I retrieve the properties of that sample?
- do we want a sample to have an array of metadata info?
- do we want to have created/updated fields on a sample?
- do we want any way to search for samples, either in a callset, a dataset, or globally? if so, by which attributes?
- is "sample" the right concept? Or do we want something finer grain (e.g. "a sample, plus its prep, plus the toolchain used to analyze it")
- do we want any sort of hierarchy of samples (e.g. which ones come from a single subject), or is that way too messy to tackle?

diekhans:
Without something like CallSet, how does one represent running
multiple variant calling methods against a given `sample'?

Is it correct interpretation that a sample is really a misnomer
for an alignment of ReadGroupSet?

jjfarrell:
Having multiple variant sets called from the same BAMs occurs quite a lot
when calling structural variants.  VCF files  for multiple SV/indel
programs are often  typically generated to increase sensitivity/specificity
across a range of SV sizes and types (Pindel, Lumpy, Delly, CNVNator,
GenomeSTRiP, etc).

It also occurs in the GATK pipeline. There is a VariantSet of raw calls
generated which then undergoes variant quality score recalibration to
create another VariantSet.

I am also aware of a project that two VCFs were generated by two different
pipelines (Atlas and GATK) using the same bam files. These were then used
to generate yet another consensus VariantSet.

So having Sample records/objects at a higher level than the VariantSet
would be a more efficient model in these situations. This would also
simplify the situation when the Sample metadata needs to be updated due to
sample mixup in the lab (which is often detected when doing qc on the
genotypes).






On Fri, Aug 21, 2015 at 9:39 AM, Heng Li <notifications@github.com> wrote:

> any data that uses the sampleId field uses it as a simple name, not as a
> link.
>
> Are you suggesting sampleId is a name, not an ID? Anyway, I will show why
> the lack of a global sample object is problematic.
>
> As of now, sample information is tied to CallSet. What if a dataset only
> has reads but no variants? The users can't submit any sample information in
> this case. When we have both reads and variants, we always need to get
> sample information from CallSet, a variant object. This seems
> conceptually wrong. Now if we make CallSet reference a single VariantSet,
> how do we represent two variant sets called from the same BAMs (this is
> discouraged but may happen)? I guess we need to create two CallSet for
> the same sample and duplicate the sample information? Apparently, the new
> Sample is CallSet moved up a level, but this is a philosophical change
> that simples our model and solves all these problems.
>
> —
> <https://github.com/ga4gh/schemas/pull/383#issuecomment-133427174>.
>



-- 
John Farrell, Ph.D.
IT Manager
Biomedical Genetics-Evans 218
72 East Concord Street
Boston, MA

ph: 617-638-5491

lh3:
> can one readgroupset be the basis for multiple callsets? If so, where do we store the metadata associated with the callset (e.g. the toolchain was used to generate it)

The current practice is to generate a variant set in one batch. All the samples involved in the batch are associated with the same command line. This info is stored in the VCF header or equivalently, VariantSetMetadata in our API.

> can one tissue sample be the basis for multiple readgroupsets? If so, do we want to represent that in the API?

Yes, reads from one library can be spread across BAMs in principle, though this rarely happens in practice. I don't think this affect the current API either way, as the read API in master is not aware of samples at all (surprising!).

> are we trying to add biological meaning to a dataset, in addition to today's administrative and organizational meaning? (I don't think we should, but that's a separate discussion).

I have said a few times that Dataset is a crucial object type in that isolates the sample name space. Nonetheless, I am not going to touch Dataset with this PR. When defining sample, I just say that their names ought to be unique in a dataset.

> do we want to tie into a more global data model, either existing (e.g. BioSample) and/or new (e.g. some sort of guid)? Is that required for all data, optional for some data, or overkill?

I prefer to have a guid. We will certainly need that when we have both 1000G and aggregate variant sets such HRC/ExAC. Nonetheless, I don't mind if we drop guid/biosampleId for now. For the datasets hosted at google right now, we don't need guid.

> given a sample name (or id), how do I retrieve the properties of that sample?
> do we want any way to search for samples, either in a callset, a dataset, or globally? if so, by which attributes?

As I said, there will be a search method (in a new metadatasearch.avdl) once we reach a consensus on the harder problems. The search will be global for multiple data types. Methods are easy once the data model is clear.

> do we want a sample to have an array of metadata info?

Yes. I forgot to add that.

> do we want to have created/updated fields on a sample?

It doesn't matter either way. I prefer not, as typically samples come in one batch.

> is "sample" the right concept? Or do we want something finer grain (e.g. "a sample, plus its prep, plus the toolchain used to analyze it")

Sample has the same scope as BioSample and SRA Sample. It has nothing to do with the analysis software (when we call the same BAM with two callers, we mostly use the same sample name). However, the biological definition of Sample is irrelevant to this PR, except guid. If we don't want to have guid, we don't need to define samples biologically.

> do we want any sort of hierarchy of samples (e.g. which ones come from a single subject), or is that way too messy to tackle?

Certainly not in this PR.

> Without something like CallSet, how does one represent running multiple variant calling methods against a given `sample'?

@jeromekelleher has explained this earlier. If we have two calling methods, we create two VariantSets under the same Dataset. In my schema, `sampleNames` of the two VariantSets will be the same or have overlaps. The command lines are stored in VariantSet::metadata.

The harder problem is how to deal with those aggregate projects such as HRC which have 1000G as a subset. My recommendation is to create a new dataset e.g. "HRC". For samples used by both HRC and 1000G, they will have the same guid. This is how guid comes in.

diekhans:
Hi @dglazer
> @diekhans and @helenp -- yes, the heart of the issue here is the relationship
> between "all of the calls that belong together in a variantset" and "sample".

Yes, sample in metadata terminology means a biological sample.
I believe what sample in the variant design refers to a mapping
to a reference of a readgroupset.  The readgroupset is still a
several of levels from the actually sample.

> Our API today takes a pragmatic approach -- it doesn't provide a semantic
> definition of "belong together", and instead provides a simple operational
> definition -- whoever loads data into the system gets to decide. As @lh3 said,

So this goes to what is meant by `interoperable' in the DWG
goals.  Is it just at the format level or the semantic level?
In my experience, data interoperability is a far bigger (and
harder) problem in bioinformatics than the formats.

While we could have semantics is at a higher level, we will have
to have a rich enough data model to represent the actually data.
 
> P.S. When I first started in this field, I thought it would be simple -- one
> vial of tissue == one sample == one readgroupset == one callset. I was naive.

:-)  if only it was so easy ...

mbaudis:
@dglazer A single ```biosampleId``` wouldn't even cover all eventualities. In the current metadata description, we have a (not well defined) object ```experiment```, which was introduced to represent e.g. a specific library preparation. Now, if a library is prepared from pooled DNA, it may represent content from multiple ```biosampleId``` references.
A theoretically viable but rather abstract approach could be to use generic ```derivedFrom``` object id lists.

So agreement with @diekhans in that 
* the use of ```sample``` should be used equivalent to ```biosample```
* we (metadata) have to chart a data model

But it would also be helpful if people would look themselves at the metadata branch and comment/complain about what is and isn't there.

lh3:
There are sample in the informatics sense and sample in the biological sense. In the informatics sense, a sample is a unique key associated with a bunch of data. We can do most computational analyses without knowing the exact biological definition of sample. This PR attempts to abstract this "informatics sample" away from the extreme complexity of biological samples. It is not defining biological samples. Of course we need a link to biological sample. That is the primary role of "guid".

dglazer:
@lh3 , thanks for clarifying that when this PR uses "sample", it doesn't mean "biological sample" -- that wasn't clear to me. (It also wasn't clear that you meant `guid` to be a link to a biological sample -- I would have assumed that was the purpose of `biosampleId`.) I now agree with @diekhans that some sort of entity relationship diagram will be very helpful in making sure we're all talking about the same objects in the same way. I'll try to put together a strawman when I have a chance if nobody gets to it first.

lh3:
Ok, summary so far. I believe I have adequately addressed all @dglazer's questions, but my concerns with `CallSet` (e.g. no global sample info; disconnected reads/variants; clumsy multi-caller representation; wrong sample abstract model) are not resolved so far. @jeromekelleher and @jjfarrell understand and like the overall structure. @macieksmuga seems positive more or less. With +0, @mlin is not against these changes.

The most distracting fields are guid and biosampleId. I will drop them from this PR. Cross-dataset information is not supported with the current API anyway and is not much needed right now. Without guid/biosample, this PR will be entirely irrelevant to the biological definition of samples. We can always add id/type/etc for biological samples later, and add a field in Sample to reference the new biological sample type.

mlin:
I can support the concept of "informatics sample" as a join key. But, I'm not yet convinced to withdraw my proposal that jettisoning CallSet can be considered independently of creating the `record Sample`. That is, if one removes `record Sample` from this PR (moving around some comments about uniqueness), I think it remains an internally consistent proposal to reduce CallSet to "informatics sample."

Conversely, one could introduce `record Sample` and actually keep `CallSet` as some future place to squirrel away metadata about the (joint genotyping job, sample) cross. Not that I advocate this.

The reason why I think this is worth considering is that we seem to encounter controversy if we try to put anything interesting into `record Sample`. These records are then effectively just a junction table between some server ID and the user-specified sample name. I think this, like CallSet, is of marginal practical benefit.

diekhans:
I have never heard of a informatics sample.

Please come up with a name that is not confusing.

lh3:
@mlin 1000G has population and gender annotations. If we ever want to keep these on the server, we need a proper sample object. And for this purpose, the new dataset/project-specific `Sample` suits much better than the old variant-specific `CallSet`.

In our current world, once we have Sample, CallSet has no benefit other than confusing users. When we move to the N+1 variant calling method some day, we need to revamp the variants API ground up. It won't be late to add an independent CallSet at that time when we have the real need.

@diekhans Naming is irrelevant as "informatics sample" is only figurative and not used in the schema at all. Have a look at the latest PR and tell me what could go wrong specifically and how it is inferior to our current schema. Thanks.

jjfarrell:
While not in the master branch, it looks like the Metadata Task Team has
been working on Individual, Sample and Experiment schema records which
should be useful and  help address several of the issues raised in this
discussion.

https://github.com/ga4gh/schemas/blob/metadata/src/main/resources/avro/metadata.avdl


On Fri, Aug 21, 2015 at 2:08 PM, Mark Diekhans <notifications@github.com>
wrote:

>
> I have never heard of a informatics sample.
>
> Please come up with a name that is not confusing.
>
> Heng Li <notifications@github.com> writes:
> > There are sample in the informatics sense and sample in the biological
> sense.
> > In the informatics sense, a sample is a unique key associated with a
> bunch of
> > data. We can do most computational analyses without knowing the exact
> > biological definition of sample. This PR attempts to abstract this
> "informatics
> > sample" away from the extreme complexity of biological samples. It is not
> > defining biological samples. Of course we need a link to biological
> sample.
> > That is the primary role of "guid".
> >
> > —
> > Reply to this email directly or view it on GitHub.*
>
> >
>
> —
> <https://github.com/ga4gh/schemas/pull/383#issuecomment-133516156>.
>



-- 
John Farrell, Ph.D.
IT Manager
Biomedical Genetics-Evans 218
72 East Concord Street
Boston, MA

ph: 617-638-5491

diekhans:
@lh3, naming is very, very relevant.  Having clear, intuitive
names is vital to having a simple, accurate, and understandable
conceptually model.

Biological samples have existed long before genomics and or
computers.  SampleGroup is needs an accurate name.  If I understand
it correctly, `analysis target' is closer to the meaning of `sample'
in this context

mlin:
> @mlin 1000G has population and gender annotations. If we ever want to keep these on the server, we need a proper sample object. And for this purpose, the new dataset/project-specific Sample suits much better than the old variant-specific CallSet.

> In our current world, once we have Sample, CallSet has no benefit other than confusing users. When we move to the N+1 variant calling method some day, we need to revamp the variants API ground up. It won't be late to add an independent CallSet at that time when we have the real need.

Agree. I have not seen a robust defense of CallSet in this discussion. I was impressed that the group had the collective will to simplify the master branch recently, and favor further efforts to simplify and iterate toward concrete use cases. I was radically asking whether we can have _neither_ `CallSet` nor `record Sample` at this point in time, but now hereby concede.

dglazer:
@lh3 , thanks for the latest schema changes and comments. I think that now means the proposed set of changes is even smaller, which should make the discussion easier. Popping the discussion up a level -- I'm having a hard time evaluating the details of names/fields/methods without a better understanding of an "informatics sample" and the problems it's meant to solve. Objects I see in play are:

* a **'column' of calls** in a `VariantSet`, today called a `CallSet`
** it belongs to exactly one `VariantSet` (since I agree with @jeromekelleher in #376 that having that field be plural is a historical artifact, that doesn't reflect the way it's actually used)
** it takes the intentionally pragmatic approach of letting whoever loads data into the system decide how they want to group `CallSet`s into a `VariantSet` and `VariantSet`s into a `DataSet`
** it allows you to use metadata to suggest cross-linking with other `CallSet` and `ReadgroupSet` records, and with biological samples, but not in any formally defined way

* a **biological sample**, with the proposed name of `Sample` in the metadata branch
** it is an identifier for a physical tissue specimen
** it can, through various chains of wet and dry processing, give rise to one or more `ReadgroupSet` and `CallSet` objects
** it could be organized into various hierarchies (e.g. by subject, library, pipeline), and could be joined to various existing catalogs (e.g. BioSample); details being discussed in the metadata TT

* an **informatics sample**, with the proposed name of `Sample` in this PR
**  per @lh3, it is "*a unique key associated with a bunch of data. We can do most computational analyses without knowing the exact biological definition of sample.*"
** it belongs to exactly one `Dataset`
** I think it can be associated with zero or more `ReadgroupSet`s and zero or more columns of calls, as long as they're all in the same `Dataset`, but am not sure I'm understanding correctly.
** it allows you to use metadata to suggest cross-linking with biological samples, but not in any formally defined way

If that's right, the fundamental changes proposed in this PR are:
* add a formal way to connect one or more `ReadgroupSet` and 'column of calls' objects (*dG: +1 to the goal, once we come up with an approach we all like*)
* use a new name for the object (*dG: open to a better name; -1 on calling it `Sample`*)
* require all 'columns of calls' connected to a single `ReadgroupSet` to have been generated the same way (*dG: -1 on removing the flexibility to 'mix methods' in a single `VariantSet`*)
* make `Dataset` part of the semantic definition of the object hierarchy (*dG: -1*)

lh3:
> use a new name for the object (dG: open to a better name; -1 on calling it Sample)

Don't mind about naming. `DatasetSample` would be more precise but less appealing. Other suggestions are welcomed.

> require all 'columns of calls' connected to a single ReadgroupSet to have been generated the same way (dG: -1 on removing the flexibility to 'mix methods' in a single VariantSet)

Show me an existing data set where this flexibility is needed right now. Depending the actual use cases, I believe I can find better solutions than using CallSet.

> make Dataset part of the semantic definition of the object hierarchy (dG: -1)

In our current API, do we allow two distinct samples in a dataset to have the same name? If we allow, how can we distinguish samples? If each sample name uniquely identifies a sample, this will be the same as this PR. This PR just exposes the crucial hidden assumption that has been haunting GA4GH from the very beginning.
<hr>
Here are why the current API is problematic:

* -1 on `CallSet` name. Few in this field would think a call set is equivalent to a sample.
* -1 on being unable to connect variants and the corresponding reads due to the lack of a global sample object.
* -1 on the ambiguous `sampleId`in ReadGroup and CallSet et al. Is it a user submitted name or a repo-defined ID? You seemed to suggest the former, but then why there are both sampleId and name in CallSet?
* -1 on being unable to store sample information without variants. If `sampleId` is an ID, we even can't store sample names without variants.
* -1 on changing CallSet::variantSetIds[] to a single value. We wouldn't be able to represent multiple VariantSets called from the same samples. Multiple people have expressed their desire of this feature. -1 on using multiple variantSetIds[] values as well. This is more problematic. No way out with the current model.
* -1 on being unable to identify the same sample across datasets. The -1 is also applicable to the latest version of this PR as I have removed guid and biosampleId for clarity.

These are more serious concerns than your -1.

dglazer:
To be clear, I'm not saying the status quo is ideal -- we have a chance to clean up some historical cruftiness, and a chance to add the "represent connected reads and variants" feature we deferred the first time around. I just want to make sure that we make the best possible changes, since there's a cost any time we break backwards compatibility, so we don't want to do it often.

It sounds like my understanding of this PR (spelled out in https://github.com/ga4gh/schemas/pull/383#issuecomment-133586250), including the goals, the definition of an 'informatics sample', and the proposed changes, is correct -- good. Once I've digested a bit, I'll try to propose a modified way to achieve those goals, and (assuming I succeed) to articulate why I think it's a better fit for the long term.

dglazer:
I just wrote up #390 -- it's not yet an concrete alternate proposal, but instead attempts to provide a conceptual framework to help with this and other related discussions.

mlin:
I will at least reiterate my +0 as I know @lh3 is eager for input on this. I like the simplification in removing CallSet but am neutral at best on this new first-class `record Sample` containing almost nothing. 

However, I did then realize we have a similar "vessel for a unique ID and metadata" which some of our customers do actually use in a very similar manner as proposed here. We call it...the [Record](https://wiki.dnanexus.com/API-Specification-v1.0.0/Records#) :yum:. That is not a very common use or even one we'd thought of in advance though. Most of our users are fine with organizing related items in a more traditional folder structure, or by setting metadata on the analysis that generates them, or lastly the sample-level metadata lives in their own LIMS and they're not so interested in representing it in our system.

jeromekelleher:
I find @lh3's arguments pretty compelling here. It does seem that 'sample' is a tricky concept though, and so we should probably try to think of a different name to avoid confusion. In my naive view the concept corresponds to whatever 'NA12878' means in 1000G. Being able to tie together reads and variants in a systematic way within a dataset is essential, and this is by far the best proposal I've seen.

However, I agree that this is all quite complicated, and we should have at least an agreed on outline for how we attach all the required metadata before going forward with something like this.

deannachurch:
The 'sample' concept is tricky- but hugely important, especially in a cancer setting where an individual may provide may samples over the course of a treatment/project. In the clinical world, they seem to prefer the term 'specimen' over sample to indicate a particular thing (tissue, FFPE cure, vial of blood) to be sequenced. Additionally, I would suggest talking to the folks that developed BioSample- not to suggest they've solved this problem, but they have been thinking about it for years.

diekhans:
@deannachurch, the problem here is the think being called Sample
or DatasetSample is not a sample, it's something derived from
one of more samples.  It's slang from VCF.

deannachurch:
Thanks @diekhans - that is unfortunate. It would be a shame to see that propagated from VCF. Talking to the BioSamples folks though is probably a good idea.

diekhans:
+1 on dropping CallSet. CallSet makes no sense because it's not
related to anything in the data production and analysis model.
It's a column in a VCF.

-1 on the PR as it currently stands.

I actually think Heng's change probably make sense and are right,
however:
  - It's weakly documented and it's not clear how if fits in the
    rest of the GA4GH conceptual or data model.  This is true
    for most of GA4GH, so it's not a criticism of just this PR.

  - The terminology is unclear.  The thing being called a sample
    isn't a sample.  It's something derived from one or more
    samples.  It doesn't matter how much slang people use, that
    doesn't make a clear data model.  Calling it a DatasetSample
    makes it no clearer.  We now have a DatasetSample and a
    Sample metadata objects in a dataset.  What is it?  It an
    alignment of a read group?  A set of alignments of read
    groups?  Some arbitrary key linking things together?

  - This seems to change the definition of DataSet from google's
    definition, but it's not clear how.

  - There are no documented use cases.  How do we know this is
    right? How does one know how to map their data to this?

  - What are the immutable objects?  That is, what gets created
    once and what can be incrementally added?

CallSet has been around for a over a year and we are just
figuring out that it doesn't represent the data.  We need to
actually think things through in a documented way if we don't
want to repeat this.

diekhans:
@deannachurch, thanls, you have just demonstrated why we need a
clear data model.

The metadata group BioSamples needs to develop use cases from
BioSamples, it seems very well thought out.

src/main/resources/avro/variants.avdl
```diff
@@ -58,43 +58,15 @@ record VariantSet {
...
-  */
-  union { null, long } updated = null;
-
-  /**
-  A map of additional call set information.
+  The metadata associated with this variant set. This is equivalent to
+  the VCF header information not already presented in first class fields.
   */
-  map<array<string>> info = {};
+  array<VariantSetMetadata> metadata = [];
```
adamnovak: This is definitely a good idea to change. That's why we have this type.

adamnovak:
I am -1 on having references to `NamedSample` being by name. If it has an ID (which it does), we ought to reference it by ID.

I am +1 on getting rid of `CallSet` and just pointing `Call`s at some kind of metadata record for what they are trying to call on.

What are the advantages of this `NamedSample` over the metadata team's `Sample` object? Can we just port theirs over and point calls at it?

lh3:
Closing this issue as there are too many parallel threads on the same issue, although I am still in favor of this PR.

PR #384, opened 2015-08-20
Changing the type of the exists field in a beacon's response to boolean.

mcupak:
#335 rebased and submitted against the beacon branch.

### DIFF
--- a/src/main/resources/avro/beacon.avdl
+++ b/src/main/resources/avro/beacon.avdl
@@ -137,11 +137,8 @@ record BeaconInformationResource {
 The response to the Beacon query
 */
 record ResponseResource {
-  /** Whether the beacon has observed variants. True if an observation exactly matches request. Overlap if an
-  observation overlaps request, but not exactly, as in the case of indels or if the query used wildcard for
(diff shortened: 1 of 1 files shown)


src/main/resources/avro/beacon.avdl
```diff
@@ -137,11 +137,8 @@ record BeaconInformationResource {
 The response to the Beacon query
 */
 record ResponseResource {
-  /** Whether the beacon has observed variants. True if an observation exactly matches request. Overlap if an
-  observation overlaps request, but not exactly, as in the case of indels or if the query used wildcard for
-  allele. False if data are present at the requested position but no observations exactly match or overlap. Null
-  otherwise. */
-  string exists;
+  /** Indicator of whether the beacon has observed the variant. */
+  union{ null, boolean } exists;
```
mlin: Is it still valid to return null? If so when is that supposed to happen?

mcupak: We're taking null as "not specified" here, e.g. when you want to return an error (invalid query etc.). It's not meant to be a "valid" response. Now that you mention this, it probably makes sense to change the way errors are reported to avoid this confusion.

PR #388, opened 2015-08-20
Initializing beacon's frequency field to null as it's optional.

mcupak:
#339 rebased and submitted against the beacon branch.


mfiume:
+1

PR #392, opened 2015-08-26
Remove nullable referenceSetId in SearchReferences.

jeromekelleher:
This corrects an inconsistency in the API, and makes the SearchReferencesRequest symmetric with the other SearchXRequest objects. All other requests (SeachVariantSets, SearchReadGroupSets, etc) take the parent container as a mandatory argument.

This is a bugfix in my view, and should be included in the next release.


calbach:
+1. This seems reasonable to me.

macieksmuga:
+1 yes, fix!

benedictpaten:
+1

PR #395, opened 2015-08-30
Removing CallSet as a first-class object; VCF merging and the N+1 problem

mlin:
(Background: @dglazer recently wrote a [nice recap](https://github.com/ga4gh/schemas/pull/391#issuecomment-134821662) of the conceptual model underlying the Reads+Variants API. This PR grew out of discussion in #383 and #391, and involves conceptual issues that have been aired previously #49 #73 #168, but perhaps not fully litigated in this group.)

The concept on trial in this PR is the CallSet, which represents a "column of genotype calls" in a VariantSet, roughly equivalent to one column of a multicolumnar VCF file. CallSets are first-class objects in our API, with their own 'created' and 'updated' timestamps, and optionally other metadata. This supports a mode of operations in which existing VariantSets can be merged together, such that the merged VariantSet composes several CallSets generated at different times and perhaps using different methods. Modeling this situation seems to be the main reason for CallSet to be a first-class object rather than constitutively a part of a single VariantSet.

Because our Variants schema adheres to the VCF data model, it inherits some fundamental difficulties with merging VCFs. First, there's distinguishing homozygous reference from no-call at variant sites not present in all CallSets. Second, it's difficult to construct a "convenient" representation of the merged VariantSet when there are multiple different overlapping variants in the CallSets. We'll elaborate on "convenient" below, but the result of these difficulties is that the generation of multicolumnar VariantSets is usually more complicated than the composition of CallSets contemplated above. There are often algorithmic steps involved that transform the inputs or supplement them with other data, such that the new VariantSet really consists of all-new CallSets generated together. This "N+1 problem" — the need to reprocess everything in some way when adding a new [analysis] sample to a study — is an outstanding challenge in the abstract VCF data model.

Now that stated, simple compositional merging of existing VariantSets — e.g. joining VCF rows by exact (POS,REF,ALT) — can be a useful operation in practice. That type of procedure handles common cases like biallelic SNVs well, and VCF data can be prepared in particular ways that help to marginalize other problematic cases (e.g. reducing to allelic primitives, splitting multiallelic sites into overlapping biallelic sites, and including all homozygous reference genotypes). The residual issues/limitations with VCF merging may then be quite tolerable, depending on the use case.
### The proposal

This pull request removes CallSet as a first-class API object. A column of calls becomes constitutively a part of one VariantSet and shares its metadata. API operations refer to columns by some "handle" string, unique at least within the VariantSet. This "handle" concept is a placeholder as other in-flight pull requests aim to ascribe additional meaning to them, upon which a different term such as "analysis sample" will be substituted.
### The prosecution
- Let's jettison CallSet because we don't need to provide a first-class abstraction mainly useful to model one rather simplistic way of merging VariantSets.
- This doesn't _prevent_ anyone from merging VariantSets however they like, creating new VariantSets with distinct columns.
- It diminishes a potential illusion that the Global Alliance already has a sound general solution for VCF merging or the N+1 problem.
- The CallSet concept has proven confusing to many of us, let alone outsiders, and removing it simplifies things. 
- We can add a CallSet-like object later, if/when we adopt a new model of genetic variation with easier merge semantics (which will probably require an effective rewrite of the Variants API anyway).
### The defense
- Let's keep our first-class CallSets because they're helpful to model VariantSet merging, a useful and commonly-practiced operation.
- The corner cases where simple VariantSet merging causes problems are tolerable for many use cases, and it's much easier and more scalable than alternatives.
- CallSet objects are more generally a useful place to stick any metadata applicable specifically to the (VariantSet, analysis sample) cross, regardless of how/when they were generated.
- Just finding a better name for CallSet would suffice to reduce confusion.
- Having first-class CallSets now will architecturally prepare implementations for future variation models with easier merge semantics.

<hr/>
### Appendix: how should a multicolumnar VariantSet look?

By "convenient", above, we mean a multicolumnar VariantSet would ideally present a "squared-off" matrix of genotypes, with non-overlapping variants, mutually-exclusive alleles, and the fewest possible missing entries. This is desirable for downstream analysts, because it's then straightforward to calculate population-genetic statistics, association tests, genotype posteriors, and partition functions. Those are much harder to formulate based on "ragged" genotypes, because of the need to meticulously account for the entanglement of overlapping variants and alleles. Indeed, beyond mere convenience, some propose that VCFs "[SHOULD NOT](https://www.ietf.org/rfc/rfc2119.txt)" contain overlapping variants [[1](https://github.com/ga4gh/schemas/issues/168#issuecomment-60795294),[2](https://github.com/ga4gh/schemas/issues/168#issuecomment-60789648)].

Current formal methods for finding the squared-off genotype matrix are not really "VCF merging" _per se_, but rather probabilistic algorithms relying on lower-level forms of input data. The most challenging aspect of these algorithms is that the matrix is generated jointly, modeling all observations simultaneously in each row, and possibly across rows. Observing the N+1th sample sometimes causes revisions to all previous N samples!

Short of formal joint genotyping, there are also _ad hoc_ VCF merging procedures that work harder to achieve "squareness" than (POS,REF,ALT) joining, e.g. by unifying multiallelic sites and rewriting genotypes when possible, and using auxiliary read coverage information to adjudicate ref/no-call. Still, these are also generating _new_ data rather than merely composing existing data.

### DIFF
--- a/src/main/resources/avro/variantmethods.avdl
+++ b/src/main/resources/avro/variantmethods.avdl
@@ -70,11 +70,12 @@ record SearchVariantsRequest {
   string variantSetId;
 
   /**
-  Only return variant calls which belong to call sets with these IDs.
+  Only return variant calls belonging to these columns.
   If an empty array, returns variants without any call objects.
   If null, returns all variant calls.
+  TODO: rename
   */
-  union { null, array<string> } callSetIds = null;
+  union { null, array<string> } columnHandles = null;
 
   /** Required. Only return variants on this reference. */
   string referenceName;
@@ -112,11 +113,10 @@ record SearchVariantsRequest {
 record SearchVariantsResponse {
   /**
   The list of matching variants.
-  If the `callSetId` field on the returned calls is not present,
-  the ordering of the call sets from a `SearchCallSetsRequest`
-  over the parent `VariantSet` is guaranteed to match the ordering
-  of the calls on each `Variant`. The number of results will also be
-  the same.
+  If the `columnHandle` field on the returned calls is not present, the
+  ordering of the columns from a `SearchVariantsRequest` over the parent
+  `VariantSet` is guaranteed to match the ordering of the calls on each
+  `Variant`. The number of results will also be the same.
   */
   array<org.ga4gh.models.Variant> variants = [];
 
@@ -146,69 +146,6 @@ Gets a `Variant` by ID.
 org.ga4gh.models.Variant getVariant(
   /**
   The ID of the `Variant`.
-  */
-  string id) throws GAException;
-
-/******************  /callsets/search  *********************/
-/** This request maps to the body of `POST /callsets/search` as JSON. */
-record SearchCallSetsRequest {
-  /**
-  The VariantSet to search.
-  */
-  string variantSetId;
-
-  /**
-  Only return call sets with this name (case-sensitive, exact match).
-  */
-  union { null, string } name = null;
-
-  // TODO: Add more ways to search by other metadata
-
-  /**
-  Specifies the maximum number of results to return in a single page.
-  If unspecified, a system default will be used.
-  */
-  union { null, int } pageSize = null;
-
-  /**
-  The continuation token, which is used to page through large result sets.
-  To get the next page of results, set this parameter to the value of
-  `nextPageToken` from the previous response.
-  */
-  union { null, string } pageToken = null;
-}
-
-/** This is the response from `POST /callsets/search` expressed as JSON. */
-record SearchCallSetsResponse {
-  /** The list of matching call sets. */
-  array<org.ga4gh.models.CallSet> callSets = [];
-
-  /**
-  The continuation token, which is used to page through large result sets.
-  Provide this value in a subsequent request to return the next page of
-  results. This field will be empty if there aren't any additional results.
-  */
-  union { null, string } nextPageToken = null;
-}
-
-/**
-Gets a list of `CallSet` matching the search criteria.
-
-`POST /callsets/search` must accept a JSON version of `SearchCallSetsRequest`
-as the post body and will return a JSON version of `SearchCallSetsResponse`.
-*/
-SearchCallSetsResponse searchCallSets(
-  /** This request maps to the body of `POST /callsets/search` as JSON. */
-  SearchCallSetsRequest request) throws GAException;
-
-/****************  /callsets/{id}  *******************/
-/**
-Gets a `CallSet` by ID.
-`GET /callsets/{id}` will return a JSON version of `CallSet`.
-*/
-org.ga4gh.models.CallSet getCallSet(
-  /**
-  The ID of the `CallSet`.
   */
   string id) throws GAException;
 
--- a/src/main/resources/avro/variants.avdl
+++ b/src/main/resources/avro/variants.avdl
@@ -41,7 +41,7 @@ record VariantSetMetadata {
 }
 
 /**
-`Variant` and `CallSet` both belong to a `VariantSet`.
+`Variant` belongs to a `VariantSet`.
 `VariantSet` belongs to a `Dataset`.
 The variant set is equivalent to a VCF file.
 */
@@ -58,43 +58,17 @@ record VariantSet {
   string referenceSetId;
 
   /**
+  Handles to the "columns" of the VariantSet, each unique at least within the
+  VariantSet.
+  TODO: rename this pending the outcome of other PRs.
+  */
+  array<string> columnHandles = [];
+
+  /**
   The metadata associated with this variant set. This is equivalent to
   the VCF header information not already presented in first class fields.
   */
   array<VariantSetMetadata> metadata = [];
-}
-
-/**
-A `CallSet` is a collection of variant calls for a particular sample.
-It belongs to a `VariantSet`. This is equivalent to one column in VCF.
-*/
-record CallSet {
-
-  /** The call set ID. */
-  string id;
-
-  /** The call set name. */
-  union { null, string } name = null;
-
-  /** The sample this call set's data was generated from. */
-  union { null, string } sampleId;
-
-  /** The IDs of the variant sets this call set has calls in. */
-  array<string> variantSetIds = [];
-
-  /** The date this call set was created in milliseconds from the epoch. */
-  union { null, long } created = null;
-
-  /**
-  The time at which this call set was last updated in
-  milliseconds from the epoch.
-  */
-  union { null, long } updated = null;
-
-  /**
-  A map of additional call set information.
-  */
-  map<array<string>> info = {};
 }
 
 /**
@@ -108,23 +82,14 @@ the occurrence of a SNP named rs1234 in a call set with the name NA12345.
 record Call {
 
   /**
-  The name of the call set this variant call belongs to.
-  If this field is not present, the ordering of the call sets from a
-  `SearchCallSetsRequest` over this `VariantSet` is guaranteed to match
-  the ordering of the calls on this `Variant`.
+  The handle of the column this variant call belongs to.
+  If this field is not present, the ordering of the columns from a
+  `SearchVariantsRequest` over this `VariantSet` is guaranteed to match the
+  ordering of the calls on this `Variant`.
   The number of results will also be the same.
+  TODO: rename
   */
-  union { null, string } callSetName = null;
-
-  /**
-  The ID of the call set this variant call belongs to.
-
-  If this field is not present, the ordering of the call sets from a
-  `SearchCallSetsRequest` over this `VariantSet` is guaranteed to match
-  the ordering of the calls on this `Variant`.
-  The number of results will also be the same.
-  */
-  union { null, string} callSetId = null;
+  union { null, string } columnHandle = null;
 
   /**
   The genotype of this variant call.


mlin:
Since I opened this PR I guess I'm +1 but hopefully it's clear it's a close call to me, and I look forward to discussion. On the one hand I believe joint genotyping is a worthwhile best practice, and my group has a [side project](https://github.com/dnanexus-rnd/GLnexus/wiki) to scale it up. On the other hand we work with a number of customers with various *ad hoc* ways of merging VCFs and they're generally doing alright (the customer is always right!?)

lh3:
Apart from some minor differences, your treatment of `CallSet` is the same as my mind model on top of #383. You have my +1.

As an additional note, in contrast to #391 which attempts to model the variant calling process with `CallSet`, `ColumnHandler` hides away this process from the backbone of our data model. Variant calling can be very complex, but is ignored by most API users. Replacing first-class `CallSet` with an attribute-like object `ColumnHandler` will lead to a cleaner and simpler data/query model that is sufficient for most typical uses, while still retaining the ability to describe complex calling processes if we wish to.

lh3:
I should say that you have my +1 if we don't want to have an object equivalent to `AnalysisSample`, nonetheless, I still think it is clearer to have this object above reads and variants.

dglazer:
@mlin, nice writeup. (Or should I say "nice brief"? ;)

I think the essence of the schema change is well described by one of your points for the defense:
> CallSet objects are more generally a useful place to stick any metadata applicable specifically to the (VariantSet, analysis sample) cross, regardless of how/when they were generated.

I find that point compelling, and know of at least three use cases in which that metadata is very useful:
- variant set merging, as you describe (assuming the user has done something sane with all the caveats you mention)
- @haussler 's example of wanting to generate multiple call sets from a single collection of reads, to represent the clonal content of a tumor
- one of our customers, Autism Speaks, currently has a published variant set consisting of ~1300 call sets that came from Illumina data and ~400 that came from CGI data.

In all of those cases, the cross-associated metadata is essential, and today's CallSet provides a straightforward way to represent it. If we get rid of the CallSet object, I don't know how we'd handle any of these cases cleanly. (There are of course workarounds, but I think they'd all be messier than the status quo.)

I do agree with the prosecution that we haven't done a good enough job of documenting the current model, and am all for improvements on that front -- I'm eager to see the new framework from @diekhans , which should give us a good place to add those improvements.

lh3:
As I said before, metadata are attributes and are not qualified as first-class objects. You can just have:
```java
record CallingMethod { string analysisSampleId; /* then metadata on calling this sample */ }
record VariantSet { array<string> analysisSampleIds; array<CallingMethod> methods; }
```
or equivalently:
```java
record CallingMethod { string variantSetId, analysisSampleId; /* then metadata */ }
record VariantSet { array<string> analysisSampleIds; }
```
and move everything in your CallSet to CallingMethod to describe how each analysisSample is called in a VariantSet. This CallingMethod can be as complex/sophisticated as CallSet, but does not complicate first-class objects – most of time, API users simply ignore this CallingMethod. If we have CallSet, users would have to deal with an extra unfamiliar layer to retrieve data (when we work with variants, we rarely need this CallSet layer), and get confused with questions like: how is CallSet::name different from AnalysisSample::name?

diekhans:
Wonderful write-up @mlin.

It seems like a problem here is that we are trying to recode VCF
in JSON, as opposed to come up with a natural model that
represents the data.  Of course, this should suffer no data lose
over VCF and should be transformable into VCF.

It might even make sense for servers to be able to generate VCF
files from native GA4GH data model.  This would be preferable
over having a unnatural data model.

I think it would be a worth-while exercise to think through the
data model independent of VCF and see were things meet.

Here is a example of what the data create flow looks like
for two read groups with a single calling algorithm.

sample-A --> readgroup-A-D --> variant-calls-A-D-F --> variant-calls-A-DE-F
         \-> readgroup-A-E --> variant-calls-A-E-F -/

In most cases, the final analysis pretends like readgroup batch
effects don't exist and want to look at variant-calls-A-DE-F.
One wants to make it easy for the common case without losing the
underlying details.

If one actually cases about the case of comparing multiple
mutation calling methods, it looks more like:

sample-A --> readgroup-A-D --> variant-calls-A-D-F
         |                 \-> variant-calls-A-D-G
         \-> readgroup-A-E --> variant-calls-A-E-F
                           \-> variant-calls-A-E-G

A VCF file and a VariantSet are a combination of some set of
variant calls.  Having this static structure is a restriction of
a single flat file design.  For a dynamic API, a VariantSet
could be the result of a query, not a primary data structure.
There would no longer be a need for fixed merging of variant
calls.

I know this is simplistic to the point of being naive, but going
through a rethinking of how to actually represent the data as
opposed to encode VCF is necessary if we want to look towards
the future.

richarddurbin:
Very nice description and proposal.  Having read that and Heng's and your comments I am +1. 

Richard

Sent from my iPhone

> On 30 Aug 2015, at 8:25 pm, Mike Lin <notifications@github.com> wrote:
> 
> (Background: @dglazer recently wrote a nice recap of the conceptual model underlying the Reads+Variants API. This PR grew out of discussion in #383 and #391, and involves conceptual issues that have been aired previously #49 #73 #168, but perhaps not fully litigated in this group.)
> 
> The concept on trial in this PR is the CallSet, which represents a "column of genotype calls" in a VariantSet, roughly equivalent to one column of a multicolumnar VCF file. CallSets are first-class objects in our API, with their own 'created' and 'updated' timestamps, and optionally other metadata. This supports a mode of operations in which existing VariantSets can be merged together, such that the merged VariantSet composes several CallSets generated at different times and perhaps using different methods. Modeling this situation seems to be the main reason for CallSet to be a first-class object rather than constitutively a part of a single VariantSet.
> 
> Because our Variants schema adheres to the VCF data model, it inherits some fundamental difficulties with merging VCFs. First, there's distinguishing homozygous reference from no-call at variant sites not present in all CallSets. Second, it's difficult to construct a "convenient" representation of the merged VariantSet when there are multiple different overlapping variants in the CallSets. We'll elaborate on "convenient" below, but the result of these difficulties is that the generation of multicolumnar VariantSets is usually more complicated than the composition of CallSets contemplated above. There are often algorithmic steps involved that transform the inputs or supplement them with other data, such that the new VariantSet really consists of all-new CallSets generated together. This "N+1 problem" — the need to reprocess everything in some way when adding a new [analysis] sample to a study — is an outstanding challenge in the abstract VCF data model.
> 
> Now that stated, simple compositional merging of existing VariantSets — e.g. joining VCF rows by exact (POS,REF,ALT) — can be a useful operation in practice. That type of procedure handles common cases like biallelic SNVs well, and VCF data can be prepared in particular ways that help to marginalize other problematic cases (e.g. reducing to allelic primitives, splitting multiallelic sites into overlapping biallelic sites, and including all homozygous reference genotypes). The residual issues/limitations with VCF merging may then be quite tolerable, depending on the use case.
> 
> The proposal
> 
> This pull request removes CallSet as a first-class API object. A column of calls becomes constitutively a part of one VariantSet and shares its metadata. API operations refer to columns by some "handle" string, unique at least within the VariantSet. This "handle" concept is a placeholder as other in-flight pull requests aim to ascribe additional meaning to them, upon which a different term such as "analysis sample" will be substituted.
> 
> The prosecution
> 
> Let's jettison CallSet because we don't need to provide a first-class abstraction mainly useful to model one rather simplistic way of merging VariantSets.
> This doesn't prevent anyone from merging VariantSets however they like, creating new VariantSets with distinct columns.
> It diminishes a potential illusion that the Global Alliance already has a sound general solution for VCF merging or the N+1 problem.
> The CallSet concept has proven confusing to many of us, let alone outsiders, and removing it simplifies things.
> We can add a CallSet-like object later, if/when we adopt a new model of genetic variation with easier merge semantics (which will probably require an effective rewrite of the Variants API anyway).
> The defense
> 
> Let's keep our first-class CallSets because they're helpful to model VariantSet merging, a useful and commonly-practiced operation.
> The corner cases where simple VariantSet merging causes problems are tolerable for many use cases, and it's much easier and more scalable than alternatives.
> CallSet objects are more generally a useful place to stick any metadata applicable specifically to the (VariantSet, analysis sample) cross, regardless of how/when they were generated.
> Just finding a better name for CallSet would suffice to reduce confusion.
> Having first-class CallSets now will architecturally prepare implementations for future variation models with easier merge semantics.
> Appendix: how should a multicolumnar VariantSet look?
> 
> By "convenient", above, we mean a multicolumnar VariantSet would ideally present a "squared-off" matrix of genotypes, with non-overlapping variants, mutually-exclusive alleles, and the fewest possible missing entries. This is desirable for downstream analysts, because it's then straightforward to calculate population-genetic statistics, association tests, genotype posteriors, and partition functions. Those are much harder to formulate based on "ragged" genotypes, because of the need to meticulously account for the entanglement of overlapping variants and alleles. Indeed, beyond mere convenience, some propose that VCFs "SHOULD NOT" contain overlapping variants [1,2].
> 
> Current formal methods for finding the squared-off genotype matrix are not really "VCF merging" per se, but rather probabilistic algorithms relying on lower-level forms of input data. The most challenging aspect of these algorithms is that the matrix is generated jointly, modeling all observations simultaneously in each row, and possibly across rows. Observing the N+1th sample sometimes causes revisions to all previous N samples!
> 
> Short of formal joint genotyping, there are also ad hoc VCF merging procedures that work harder to achieve "squareness" than (POS,REF,ALT) joining, e.g. by unifying multiallelic sites and rewriting genotypes when possible, and using auxiliary read coverage information to adjudicate ref/no-call. Still, these are also generating new data rather than merely composing existing data.
> 
> 
>   https://github.com/ga4gh/schemas/pull/395
> 
> Commit Summary
> 
> Remove CallSet as a first-class entity; replace with a temporary 'column handle' concept (to be renamed later)
> File Changes
> 
> M src/main/resources/avro/variantmethods.avdl (77)
> M src/main/resources/avro/variants.avdl (63)
> Patch Links:
> 
> https://github.com/ga4gh/schemas/pull/395.patch
> https://github.com/ga4gh/schemas/pull/395.diff
> —
> 



-- 

dglazer:
@richarddurbin , how would you suggest we represent *"metadata applicable specifically to the (VariantSet, analysis sample) cross"* if we adopt this proposal? I"m worried about the real-world use cases (see https://github.com/ga4gh/schemas/pull/395#issuecomment-136246221) that we today support well, but would no longer be able to represent if we go down this path.

@lh3 , I agree that adding something like your `CallingMethod` object to this PR would provide an alternate home for that metadata, and if someone wants to flesh that out, we can certainly discuss it. Two small notes:
- we'd have to be sure that it supports multiple call columns in a variant set being derived from a single `analysisSampleId`, with different `CallingMethod`s
- I'm skeptical it will be worth breaking compatibility just to rearrange things, but it's hard to know until it's fully fleshed out.

lh3:
> if someone wants to flesh that out, we can certainly discuss it

See #396, which is modified from this PR

> we'd have to be sure that it supports multiple call columns in a variant set being derived from a single analysisSampleId, with different CallingMethods

A calling method is specific to a variant set. The same sample may have different calling methods in different variant sets.

> I'm skeptical it will be worth breaking compatibility just to rearrange things, but it's hard to know until it's fully fleshed out.

@adamnovak, @mlin, @richarddurbin, @jeromekelleher (I guess?) and myself all think CallSet is problematic. If CallSet is a wrong object, we should just drop it. Compatibility is not our top concern right now.

diekhans:
I would really like to see a diagram (vector drawing) of the design.  

Having a field columnHandle in a format without columns is very
confusing.  A new comer who is not experienced is VCF will be
lost.

I remain concerned that we have created an monolith data
structure that is based on the restrictions of a spreadsheet,
rather than one based on composition of the underlying data.

jjfarrell:
@mlin Nice job with the summary and PR. This PR combined with the proposed CallingMethod does an excellent job of modeling joint-genotyped VariantSets.  I am +1 with one suggestion regarding the Call Schema.

Since the variant.calls (array<Call> calls = [] in the variant record) is no longer ragged I believe,  the Call.SampleName field is no longer necessary and could be dropped if the array of ColumnHandles and variant.calls is in the same order and length. It should be unnecessary to store the SampleName 30 million times for each  sample and variant in a typical WGS variantset. 

If the array Calls is intended to still be ragged, a map<call> calls={} would be better as the variant call for a specific sample could quickly be accessed.

dglazer:
For the people who are supportive of this PR -- ho do you suggest we represent *"metadata applicable specifically to the (VariantSet, analysis sample) cross"* if we adopt this proposal?
- If your answer is "we don't; it's not important", then I'm worried about the real-world use cases (see https://github.com/ga4gh/schemas/pull/395#issuecomment-136246221) that we today support well, but would no longer be able to represent if we go down this path.
- On the other hand, if your answer is "we can, by introducing something like a CallingMethod", then we should really be discussing a PR that includes that, such as the one Heng wrote up.

richarddurbin:
Sorry.  I had not thought that through.  Change me to a 0 until there is consensus on this. 

Richard

Sent from my iPhone

> On 31 Aug 2015, at 3:15 pm, David Glazer <notifications@github.com> wrote:
> 
> @richarddurbin , how would you suggest we represent "metadata applicable specifically to the (VariantSet, analysis sample) cross" if we adopt this proposal? I"m worried about the real-world use cases (see #395 (comment)) that we today support well, but would no longer be able to represent if we go down this path.
> 
> @lh3 , I agree that adding something like your CallingMethod object to this PR would provide an alternate home for that metadata, and if someone wants to flesh that out, we can certainly discuss it. Two small notes:
> 
> we'd have to be sure that it supports multiple call columns in a variant set being derived from a single analysisSampleId, with different CallingMethods
> I'm skeptical it will be worth breaking compatibility just to rearrange things, but it's hard to know until it's fully fleshed out.
> —
> 



-- 

jeromekelleher:
Thanks for the analysis @mlin, this has really cleared things up a lot for me. I largely agree with what you're saying.  Fundamentally, it seems to me that the current `CallSet` concept is trying to model something we can't really do in practice (@mlin's N+1 problem), and is also trying to future-proof something that will almost certainly need drastic surgery (the editing-a-linear-reference  way of describing variation that underlies VCF).  I think this is the answer to @diekhans's point: we _do_ need to rethink the underlying data representation, and that's what the reference variation project is currently doing. This isn't something that's going to happen overnight though, so we should try to model the data that we have as simply and directly as we can, now.

This is possibly a source of confusion though: do people see the current variants API as something that will live on for a long time, or as something that we seek to replace and deprecate as we find better models for genetic variation? I'm in the 'replace and deprecate camp'.

@dglazer:
> how do you suggest we represent "metadata applicable specifically to the (VariantSet, analysis sample) cross" if we adopt this proposal?

I'm confused by all the different proposals going on now, but I would have thought that each VariantSet has at most one 'column of calls' that refers back to an `AnalysisSample`. Different calling methods result in different `VariantSets`, and the `VariantSet` knows about these calling methods. The `ReadGroup` class contains a `programs` attribute, which performs much the same task, so we could even reuse this framework.

diekhans:
The problem is a lack of documentation on the overall data model
for variants (and for VCF for that matter).  For those that
don't deal with VCF every day, it's not clear.

We can say we implement the VCF conceptual model, but without
documenting what that is, there will always be incorrect
assumptions.  Note VCF model != VCF file format.

Question: why maintain the matrix data model when we are not
using matrix data structures?  Matrices are an efficient
representation for a lot of cases but also more difficult to
compose.  Are there inherent restrictions on the sample-calls in
a row that make them dependent on each other (e.g. cross-sample
calling)?  If they are dependent, how can one know the
dependencies to do a VariantSet merge?

diekhans:
@jeromekelleher, nice analysis yourself.  I am in the camp that
if it's not documented, it doesn't exist.  Also happy to join
the `new models will be developed' camp if this is stated for
the users.

If a matrix is the best representation, then we need a better
way to find that per-sample-calls than VariantSet
map<array<string>> info = {}; with no documentation on the
contents.

I have been through the contents of the BAM header are up to the
inventor exercise.  It's a huge waste of time.

mlin:
Thanks all for kind remarks!

@lh3 I continue to NOT object to AnalysisSample :smile: but it was useful at least to me to separate my thinking about the issues with CallSet. @benedictpaten reminds us that schema PRs now need to be accompanied by branches of the reference server and compliance suite. Perhaps we can work on AnalysisSample and/or CallingMethod during the Oct 12 "hackathon" with @jeromekelleher. It seems like there's enough directional agreement to pursue those steps.

@jjfarrell Hmm, I don't think we intended to go so far as removing raggedness altogether right now. Currently it's kind of the server's discretion which columns to return calls for -- the SearchVariantsRequest can specify desired columns, but the server is not required to return calls for all of them, if I read the spec correctly. There might be solid use cases for that in expert hands; after all having to put near-row-fuls of 0/0 or ./. for rare variants is definitely a downside of VCF. But I think this is fairly separate from whether the first-class CallSet concept is needed - certainly interesting to consider though. Re: array vs map, the map would be really useful if we were talking about in-memory structures, but we're looking at API payloads which the client probably has to traverse fully to deserialize anyway :/

@dglazer I'd tend to agree with @lh3 that it's tough for metadata alone to justify the existence of CallSet in light of its cons. The per-column metadata should be rather trivial in size, so it's anyway probably more convenient to get it all in one shot from the VariantSetMetadata or the hypothetical CallingMethod, rather than a series of little API requests. To me CallSet hinges more on modeling (POS,REF,ALT) merging or similar, and whether we should or not is a very interesting philosophical question for the group I think. To the [examples](https://github.com/ga4gh/schemas/pull/395#issuecomment-136246221) you mentioned:

1. Merging is the central issue indeed, but if we're just talking about where to put the metadata arising from a merge, that can certainly just be piled into the VariantSetMetadata (analogous to the VCF header as @lh3 [mentioned](https://github.com/ga4gh/schemas/pull/391#issuecomment-134658670)), made a bit more structured if by CallingMethod if we think it necessary. 
1. In the tumors/normal/parents example, I would think the mixture proportions are estimated jointly, so it's also logical to put them on the VariantSet. Cramming more stuff into arrays and maps can be a recipe for trouble admittedly, but so can having multiple different justifiable places to put it.
1. The 3rd example seems to be a specific case of the 1st. The abstract merging difficulties may be compounded by the batch effects from different sequencing tech.

jeromekelleher:
> Perhaps we can work on AnalysisSample and/or CallingMethod during the Oct 12 "hackathon" with @jeromekelleher. It seems like there's enough directional agreement to pursue those steps.

Great idea @mlin!

dglazer:
Re per-column metadata -- @mlin, it sounds like we agree that there should be a way to represent it, if for no other reason than to support the merging use cases. That means we need to be able, for any column, to store and retrieve info on where it came from (e.g. a sample id and/or a readgroupset id) and how it was generated (e.g. what apps were run, when they were run). I certainly agree it's possible to represent that by *"cramming more stuff into arrays and maps"* on the VariantSet, but I think we'd end up with something isomorphic to today's implementation, albeit with different aesthetics, at which point I'm skeptical it will be worth breaking compatibility. Hard to say without seeing a fully fleshed-out proposal, though.

In other words, I'm -1 on any proposal that drops support for this kind of metadata; I'm open to proposals that preserve it, but am skeptical they'll make sense. (And I'm +1 on improving the documentation, so the behavior is clearer.)

Re raggedness -- I agree with @mlin that the SearchVariants spec allows the server to choose between dense and sparse formats for returning variant calls, based on which is more efficient and convenient. It can either return sparse arrays of calls that each include a column identifier, or dense arrays of calls that are ordered in a determined way. In all cases it has to return all the calls from the explicitly requested columns; it can also choose to 'pad' those with calls from other columns. (See the comment on `SearchVariantsResponse` for details -- we can flesh that out in the new doc branch if helpful.)

lh3:
@dglazer You have never showed what is wrong with `CallingMethod` in the other PR which is different from `CallSet`.

dglazer:
@lh3 -- you and I started that conversation in https://github.com/ga4gh/schemas/pull/396#issuecomment-136567264, but the PR was closed before the proposal was fully fleshed out.  I suspect that once it's fully fleshed out, and supports the same use cases as today, it won't be noticeably better than what we have today, and therefore isn't worth a disruptive change. I'm happy to discuss a fully-baked proposal; my personal prior is "it ain't broke" today, except for the lack of clear documentation, which we should definitely fix in the new doc branch.

lh3:
I closed #396 because you are not responding. If you want a full PR, I have just created #412. I won't repeat myself – I think the current schema is broken and causing unnecessary confusion and complexity.

mlin:
@dglazer 

> it sounds like we agree that there should be a way to represent [per-column metadata], if for no other reason than to support the merging use cases

Actually the question of how (or if!) to express the merging use cases is the interesting one worthy of debate to me, which is what this PR write-up was meant to explicate. Where to put this or that metadata is minor IMO, as we do seem to agree there are various practical solutions, whether on the VariantSetMetadata or on other objects. 

I'm somewhat puzzled by the assertion that "it ain't broke" after all the very interesting discussion around CallSet that has unfolded here lately. Like @lh3, not sure what else to say without just repeating.

dglazer:
@mlin, good points. Let me see if I can frame the conversation in a way that works for both of us. There are roughly three options being discussed:
  * a) keep today's semantics and syntax for per-column metadata, and improve the documentation
  * b) keep today's semantics, but update the syntax (e.g. with CallingMethod vs. CallSet)
  * c) change the semantics, by dropping support for per-column metadata

Your writeup, which as you point out is focused on the more fundamental question of semantics, helped clarify my thinking on why it's important that we preserve today's semantics. In summary, I'm +1 on (a), -0 on (b) modulo details, and -1 on (c).

---------

P.S. Re "it ain't broke" -- using the framework above, I believe the documentation is clearly broken; I think the semantics is a feature not a bug, but that's the fundamental question being discussed, and is my personal opinion.

mlin:
The demotion of CallSet to me is a change in semantics, not just syntax, *even if* we add CallingMethod as I understand @lh3's proposal. Right now CallSet is a first-class data object, which can even reside within multiple VariantSets. By eliminating CallSet we make each column of calls constitutively a part of one VariantSet -- this is the semantic change. CallingMethod, if we adopt that, is just some structured metadata, not a first-class data object. Adding it does not reverse the semantic change, and its placement within VariantSetMetadata or somewhere else is mainly aesthetics.

Following this semantic change, merging VariantSets manifests as the production of a new VariantSet with distinct columns. That's more representative of the spectrum of existing methods through which multicolumnar VariantSets may be generated, and is not incompatible with (POS,REF,ALT) joining. (This last point is how the prosecution earns my vote, in a photo finish!)

dglazer:
@mlin, good clarification -- would changing this line in today's definition of `CallSet` from plural to singular get the current schema to line up with your proposed semantics?
```
/** The IDs of the variant sets this call set has calls in. */
  array<string> variantSetIds = [];
```
 The defniition of `SearchCallSetsRequest` and `SearchVariantsRequest` already only accept a singular `variantSetId`. Or is it more pervasive?

lh3:
I will reiterate the issues with CallSet. 1) CallSet is a redundant object that complicates the use of API. We can achieve the vast majority of common queries without it in the way (see #412 and older #383). Yes, CallSet holds meta info about how a sample is called, but (a) we can keep the info in a meta object and (b) most of time, we don't care about the calling procedure. We just want to get calls given a list of samples shared across reads and variants. CallSet is not qualified as a first-class object, which is the consensus of almost everyone. 2) CallSet is a confusing object. It has sampleId and callSetName. @dglazer said google is using sampleId for user submitted sample names. Then what is the use of callSetName? Think about a typical submission where we have SAM and VCF/MAF/masterVar. There is only one sample naming system across a dataset. Where do we get both sampleId and callSetName? To add the confusion, @diekhans and @adamnovak both showed diagrams where sampleId is an ID linked to a `Sample` object, interpreting sampleId in a distinct way. 3) CallSet is incompetent. If we only have reads, there are no ways to keep sample information in the API; if we have multiple VariantSets for the same set of reads, we will have to duplicate sample information in CallSet? 4) At the very least, no experienced analysts in this field can deny that CallSet is a wrong name – can you?

All these issues can be resolved by simply demoting CallSet to a meta object.

dglazer:
I think we're mixing several discussions here:
- formalizing sample support vs. changing column-of-calls support
- semantics vs. syntax vs. documentation (see [above](https://github.com/ga4gh/schemas/pull/395#issuecomment-140795578)) for columns-of-calls

I believe everyone is in favor of formalizing sample support:
- #391 is an in-progress PR to do that *without* changing columns-of-calls
- #412 is an in-progress PR to do that *and* change columns-of-calls

Neither of those PRs are mature IMO, but they do directionally show the two approaches.

Around columns-of-calls, I still *think* that, if we were to make the [singular/plural change](https://github.com/ga4gh/schemas/pull/395#issuecomment-140856738) discussed above, we'd be in agreement on semantics. I'm not sure, though. Once we figure that out, we can talk about syntax.

lh3:
I don't know what "semantics" and "syntax" mean exactly in this context, but let's be specific –

In my comments, 3) can be resolved by a more general sample object, which #391 and #412 have. 1), 2) and 4) are applicable to both the current schema and #391. I will shorten them into three questions: 1) Is it correct that we can query genotypes without CallSet? 2) In a SAM/VCF submission (e.g. 1000g), how do you set both CallSet::sampleId and CallSet::name? 4) Could you show a paper that calls a sample in the 1000g VCF as a "callset"?

mlin:
Just changing CallSet's residence from plural to singular leaves wreckage behind. Why does it have its own unique identifier and created/updated timestamps? Why are there separate API methods to search them?

To diminish the privileged representation of VariantSet composition by keeping columns in one VariantSet, yet still retain the confusing+complicating CallSet object, seems like the worst of both worlds.

jjfarrell:
@dglazer <https://github.com/dglazer>

The CallSet in the present schema is broken as it supports an incorrect
approach to analyzing variants. It is not just poor semantics. The
VariantSet resulting from simply concatenating independently called
​CallSet ​
columns
​
is not analyzable
​​
.

​Under t
he present schema, each
​CallSet ​
column in
​such a
VariantSet
​would have something like 30 million missing calls if called independently
on 600 samples
. Those genotypes may be missing because there is no coverage in that
region. The genotypes may be missing because they are homozygous for the
reference.
​​
With this resulting sparse matrix of genotypes, t
here will
​NOT ​
be a probability assigned
​for each variant in each sample-
which is what is
​needed and ​
expected.
​ This is why vcf files containing more than one sample are joint genotyped
rather than simply merging independently called variants.

Whether  a variant is included in an analysis is based on an error model
using all calls in all samples.
​By looking across all samples and  variant calls, algorithms
can help distinguish machine artifacts from true positive variants. ​
This enables highly accurate variant sets filtered on a single estimate of
the accuracy of each variant.
​ Principle component analysis of these passed calls are then further used
to adjust for population substructure.  So a data model developed to
support a sparse matrix of genotype calls on multiple samples just makes no
sense.


On Wed, Sep 16, 2015 at 8:34 PM, Mike Lin <notifications@github.com> wrote:

> Just changing CallSet's residence from plural to singular leaves wreckage
> behind. Why does it have its own unique identifier and created/updated
> timestamps? Why are there separate API methods to search them?
>
> To diminish the privileged representation of VariantSet composition by
> keeping columns in one VariantSet, yet still retain the
> confusing+complicating CallSet object, seems like the worst of both worlds.
>
> —
> <https://github.com/ga4gh/schemas/pull/395#issuecomment-140933474>.
>



-- 
John Farrell, Ph.D.
IT Manager
Biomedical Genetics-Evans 218
72 East Concord Street
Boston, MA

ph: 617-638-5491

diekhans:
Folks, this discussion is having problems because a lack of
writing an design specification.  This is both for the current
variant model or the modification of it.  Trying to understand a
design from a schema or a pile of mail messages is bound to
creates lots of confusion.

If we want this project to be successful, we have to start
actually documenting designs and the rational behind them.


Mark

lh3:
Once we researchers get BAM/VCF, we seldom need to read additional documentations to retrieve alignments/variants. We understand each other because we are following a common practice.

We are having problems here because some of us are trying to come up with something new but unproven that needs explanations and documentations. As I reflect on recent discussions, I think it would be clearer to just map GA4GH objects to the common concepts in our daily works:

* reference genome <=> ReferenceSet
* chromosome <=> Reference
* project <=> Dataset
* *ad hoc* sample used in the project <=> AnalysisSample
* BAM <=> ReadGroupSet
* read group in the BAM <=> ReadGroup
* SAM line <=> ReadAlignment
* VCF <=> VariantSet
* VCF line <=> Variant
* genotype call <=> Call
* (no CallSet here)

I know the existing file-based models have problems, but building upon the existing practice would be a better start than trying to come up with something new. At least for researchers, such mapping will make them get hands on GA4GH APIs more quickly without going through lengthy documentations.

BTW, @dglazer, do you have answers to my three questions?

diekhans:
Some feel that GA4GH's goal is to define an interoperable API
that evolves existing practices.  That needs precise definition
of the conceptual mode, not just assumptions based on the
current, undocumented practices.  

This does need to clearly represent all the concepts in existing
data structures.  If we can't represent current data easily, it's
will not succeed.  

I am not argument for one representation or another, only
conceptual definition. Say something is a row in a VCF isn't
conceptual.

jeromekelleher:
I tend to agree with @lh3's position above in terms of the basic structure of the protocol. However, we shouldn't see this as a replacement for documentation. It shouldn't be necessary to have years of experience working with BAM/VCF etc. to use the protocol. Experienced people should immediately see a clear correspondence between the concepts they are familiar with and the API objects. Less experienced people should be able to read simple, clear and definitive documentation to get them up to speed quickly.

dglazer:
@lh3, re your three questions:

> 1) Is it correct that we can query genotypes without CallSet?

I think you're asking if there's a way to use SearchVariants to get back calls from a particular column, without passing in a CallSetId? If so, the answer is no in today's API, and yes if we introduced some other column identifier.

> 2) In a SAM/VCF submission (e.g. 1000g), how do you set both CallSet::sampleId and CallSet::name? 

The API doesn't currently specify data loading -- every server implementer can define their own way to load (and store) data. In the Google implementation, you call [UpdateCallSetRequest](https://cloud.google.com/genomics/reference/rpc/google.genomics.v1#google.genomics.v1.UpdateCallSetRequest).

> 4) Could you show a paper that calls a sample in the 1000g VCF as a "callset"?

I defer to you on legacy naming conventions. I'll be delighted when we agree on everything except the name. (I don't remember all of the conversations that led to using `CallSet` originally, but I do know that we called it `CallSample` at one point, and then agreed to change it.)

dglazer:
@diekhans @jeromekelleher -- yes, better documentation is necessary in all cases -- glad @diekhans is helping us get there. #413 is an attempt at improving the documentation on the current API behavior.

dglazer:
@mlin, re plural vs. singular CallSet's -- I'm still just trying to find common ground on the desired behavior (aka semantics) of the API; I'm not yet proposing any particular syntax. I *think* that one change would align today's API with the semantics that you propose -- if so, we can then discuss syntax, including what if any wreckage to remove.

To your specific question of created/updated timestamps -- even if we say a particular column can only belong to a single VariantSet, it's still possible to add and edit those columns within that VariantSet, so timestamps are still meaningful. If a particular server chose to only support immutable VariantSets, it would only have to store a single created/updated value for all the components of that VariantSet.

mlin:
@dglazer Good clarification of your semantics. However, the removal of CallSet intentionally discards the notion of a mutable VariantSet to which one can add more columns in a stepwise fashion. This is a feature of the PR and not a bug, as you might say, and the justifications for dropping that representation (in exchange for data model simplification) have been stated here by multiple participants. If you disagree strongly, then perhaps we can begin the real discussion! :wink:

lh3:
1) Of course CallSet is needed to retrieve Call right now as it is a first-class object. Nonetheless, we can in principle retrieve a Call by sampleId with today's schema or by analysisSampleId in #412 and #391. It is not necessary to go through an extra CallSet layer to retrieve Calls. Right? 2) It will be a nightmare if users can't expect how sample names in SAM/VCF are mapped to GA4GH objects. What is google using now? You said you map sampleId to SAM/VCF sample name. Then how do you set callSetName? The same? @jeromekelleher, what is the ref server strategy? 4) Let's put CallSet naming aside if you are open to changes.

@mlin in my view, removing CallSet is not directly related to mutable VariantSet. With CallingMethod in #412, we can still support mutable VariantSet without a first-class CallSet, if google single-handedly wishes to go against our common practice. To me, the culprit is this first-class CallSet which messes things up. It actually amazes me that I am even asking question 2) today when I have never had this issue since the first NGS project I participated.

lh3:
I am not against having more detailed documentations at all. Users at all levels, me included, find SAM/VCF specs *absolutely necessary*. I mainly wanted to say that this long thread is not caused by documentations. It is caused by the wrong data model. The mapping between GA4GH and our *ad hoc* concepts will help to reveal the problem and to attract more bioinformaticians who are familiar with the existing fields.

mlin:
@lh3 Fair enough. I was wondering why your CallingMethod had created and updated timestamps :P I have several other arguments against mutable datasets in a REST API that clients are repeatedly querying, but I concede they are not too relevant here.

lh3:
@mlin yes, CallingMethod having timestamps exactly means to support mutable VariantSets, even though I think it is not a good idea with our current multi-alleleic model.

diekhans:
Mutable VariantSets?  I think it's a huge mistake to go down the
path of treating web services as a mutable file system.  This
will produce a horribly complex API semantics and recreate the
data identification problems we have now.

Simple question: Are samples in the a VCF independent of each
other?

I think I know the answer, however Heng's assertions for a data
model seem to imply otherwise.

lh3:
No, I am not advocating mutable VariantSets at all. #412 supports mutable VariantSets mainly to make google happy.

> Simple question: Are samples in the a VCF independent of each other?

In our current practice, samples are correlated during the variant calling process. If you combine samples differently during calling, you may get different variant calls at a small chance.

diekhans:
@lh3 has provide an extremely important piece of information
that should have been in the variant design specification:

> In our current practice, samples are correlated during the variant calling
> process. If you combine samples differently during calling, you may get
> different variant calls at a small chance.


So, contrary to what some of us believed, samples are not called
independently.  So this supports the contention that CallSet
should not be independent of AnalysisSet and the VariantSet
merge is an invalid concept.

adamnovak:
I like @lh3's proposal that we officially spec out how a VCF should be
converted to the GA4GH data model. While things like "This is equivalent to
a VCF row" are not *sufficient* documentation, we should say something like
"In a VCF as produced by X tool 1.0, every row should become Y and every
column should become Z, and the probability estimates should be converted
over with such and such a formula". A formal description of our data model
will be greatly improved if we tie it back to other data models that people
may have more experience with.

This official spec might include a paragraph about how you might think that
you can just stick variant sets side by side to combine across samples, but
doing that actually invalidates all the likelihoods and loses you the
benefits of co-calling. We might also want to point to a "correct" VCF
merge algorithm, if one exists.

lh3:
@diekhans @adamnovak In my view, we should see VariantSet/VCF the following way, at least for now. Before submission, users can generate the VCF in whatever way they prefer – joint calling, gVCF merging, imputation or even naive merging – which we don't care. Once the users generate this VCF, they submit it as one immutable VariantSet object. Upon query, users are only allowed to query one VariantSet at a time. @dglazer mentioned that google is keeping multiple CompleteGenomics single-sample masterVars in one VariantSet. The right way to do it is to help users to merge these single-sample masterVars into one VCF/VariantSet before/during the database import. Once import is done, the VariantSet is locked forever and becomes immutable. We can only add a new sample by replacing the entire VariantSet.

In the longer term when we adopt a binary/unary allele model, we might be able to lift some of these restrictions. For example, querying across VariantSets on the fly will be made easier. We should leave some room for future if possible. Nonetheless, these are not our top concern right now. And so far as I can see, we can have enough room for future without complicating/confusing our common uses right now.

diekhans:
Thanks @lh3, this all makes sense. I am supportive as long as we
have a documented design, with rational, not just a schema.

mlin:
I strongly support the immutable mode of operations @lh3 [describes](https://github.com/ga4gh/schemas/pull/395#issuecomment-141523968). I'd also reiterate that, whether VariantSet is mutable or immutable, implementations could still provide their own API methods to merge existing VariantSets and VCF files using any algorithm.

What any implementation provider does with their users/customers is of course their own business, and indeed I also work with groups whose particular use cases are served adequately by *ad hoc* VCF merging. From the perspective of the Global Alliance, I believe we should permit such approaches but not endorse them, either explicitly or implicitly by twisting the data model for them. They are not conceptually sound for arbitrary or even typical VariantSets under the current representation.

dglazer:
@mlin, you say [above](https://github.com/ga4gh/schemas/pull/395#issuecomment-141486790) that:
> the removal of CallSet intentionally discards the notion of a mutable VariantSet to which one can add more columns in a stepwise fashion. This is a feature of the PR and not a bug ... If you disagree strongly, then perhaps we can begin the real discussion! :wink:

On to the real discussion :) We agree that there are *"groups whose particular use cases are served adequately by ad hoc VCF merging"*, and that the Global Alliance API *"should permit such approaches"*. I believe supporting those approaches requires (to quote the defense) having a
> place to stick any metadata applicable specifically to the (VariantSet, analysis sample) cross, regardless of how/when they were generated.

Which is why, as I said before
> I'm -1 on any proposal that drops support for this kind of metadata; I'm open to proposals that preserve it

To be clear on a few points:
* I'm **not** saying that variant sets need to be mutable. I think it's fine for us to stick to only standardizing the read-only aspects of the API, and letting each server decide how it wants to allow creating/updating data.
* I'm **not** saying that ad hoc VCF merging is without risk. I agree it has sharp edges, and data owners should use care when working with their data.
* I'm **not** saying that today's CallSet syntax is the only possible way to support this metadata, although it certainly does, and we have customers using it to do so today. I haven't seen a fully baked alternate proposal yet, although #412 is getting closer. 

And finally, re how much rope to provide our API users -- here's a loose analogy. The Google implementation used to reject invalid BAM syntax on import; we later decided that there was so much invalid syntax out there it made more sense to accept it when we were pretty sure we knew what was intended, and issue a warning instead. Other server developers could choose to be more or less strict than us, as long as, for all data we do support, we use the API in the same way. I feel similarly about this per-column metadata -- we at Google choose to support our customers who need it, and we need an API that allows us to keep doing so. Not every server developer has to make the same choice.

lh3:
> I haven't seen a fully baked alternate proposal yet, although #412 is getting closer.

I have added SearchAnalysisSamples() and SearchCallingMethod() to #412 (also syntax errors, thanks to @mlin). Any further suggestions?

haussler:
Great. I hope everybody can chip in and complete #412 to everyone's
satisfaction. I support the immutable VariantSet object, as it seems
everybody does at this point. We are close!!   -D

mlin:
Hmmm, I'd thought we'd dispatched the metadata question already. None of the outstanding PRs about CallSet "drop support" for column metadata. In this PR's minimalistic approach, one puts any desired details about the columns in the semi-structured VariantSetMetadata -- this matches current practice in VCF headers. Perhaps we'd like a more structured schema for the column info, like CallingMethod in #412. The remaining structural question there is totally minor, about where to dangle it from.

Nonetheless it's certainly true that neither this nor #412 are "fully baked" and probably won't be for some time, because this now wisely requires implementation. At the moment I think we're trying to gauge rough consensus on dropping CallSet.

In either approach this first-class object is removed from the critical path of fetching calls and slicing the VariantSet, while merging remains supportable. The expression of merging is more subtle which is an improvement in my view.

mlin:
> Great. I hope everybody can chip in and complete #412 to everyone's
satisfaction. I support the immutable VariantSet object, as it seems
everybody does at this point. We are close!!   -D

Yes, if the main remaining issue to work out is where to put metadata, then we are close indeed! (close to the point of readiness for the hackathon, at which point of course everything may change entirely :laughing:)

dglazer:
@mlin, I didn't realize that this PR intended to preserve support for per-column metadata. (Especially since the original proposal says *"A column of calls becomes constitutively a part of one VariantSet and shares its metadata."*) Sorry I misunderstood -- I'm glad we agree. Does that mean we can close this PR, and move the discussion to #412? Or do we want to flesh this one out also to add that support?

Now that I know we agree on per-column metadata, we're down to discussing API syntax and style. That means I don't know how to *"gauge rough consensus on dropping CallSet"* until we have a fully-baked alternative to compare it to, since with syntax and style, the devil is often in the details. I'm happy to take another look at #412, but it's clear it's going to take a while to converge. I'm going to prioritize getting the fully-consistent baseline, with schema+reference+compliance+doc, above that.

@haussler, re "immutable VariantSet object" -- to be clear, I'm saying the API should remain agnostic about questions of how data, including VariantSets, are created and updated. I don't think we should change the API to force server implementers to only support immutable VariantSets.

mlin:
I agree with tapping the brakes on detailed discussion of this until we try out some of the ideas in the reference server, at the upcoming hackathon or otherwise. The discussion has been really useful to prepare for that and clarify the existing relationship of CallSet to merging, metadata, mutability, and documentation (or lack thereof). It will remain helpful to  raise any other issues with removing the first-class CallSet concept that would put you "in the rough" -- so far I've heard the main question is where to put column metadata, and I'm confident we'll figure that out.

I now view this PR as an intermediate waypoint to #412. There are a couple different things going on in #412, which I'd guess is perfectly fine for @lh3 because he operates on a different astral plane of intellect and experience; I wanted a smaller bite :laughing:. I'd say let's get into the code a little bit and see if we're dog-tired by the intermediate waypoint, or else whizzing by. Another way we could slice it is to add CallingMethod here and keep #412 mainly for AnalysisSample. I'm also not worried about this, and fine with closing whenever this has outlived any intermediate usefulness.

Lastly one minor related clarification/reiteration: I think nothing further is strictly needed in this PR to represent the column metadata. You put it in the existing semi-structured VariantSetMetadata, just like one would get information about the inputs to `bcftools merge` in the output VCF header nowadays. CallingMethod is an improvement to this which I'm fine with for sure.

PR #396, opened 2015-08-31
Replaced first-class object CallSet with auxiliary object CallingMethod

lh3:
This PR is based on @mlin's PR #395 that removes the first-class object CallSet. Different from #395, this PR adds CallingMethod in response to @dglazer's request to record the variant calling process. More importantly, it gives @mlin's columnHandler a specific meaning – conceptual "sample" used in analysis – but without introducing an explicit AnalysisSample object. The sample relationship between objects are kept track by ReadGroup::sampleName, VariantSet::sampleNames[], Call::sampleName and CallingMethod::sampleName. In this commit, first-class objects are:
```java
// in metadata.avdl
record Dataset {}
// in reads.avdl
record ReadGroup { string datasetId; string sampleName; }
record ReadGroupSet { string datasetId; array<ReadGroup> readGroups; }
record ReadAlignment { string readGroupId; }
// in variants.avdl
record VariantSet { string datasetId; array<string> sampleNames; }
record Call {} // embedded in Variant only
record Variant { string variantSetId; array<Call> calls; }
```
A natural evolution of this commit is to add an AnalysisSample object as in #383 and #391, but without it, the schema still works as we are not keeping sample phenotypes in our schema at present. On a side note, it may be worth rethinking about the relationship between sampleName, ReadGroup and ReadGroupSet, but that is a separate PR.

### DIFF
--- a/src/main/resources/avro/reads.avdl
+++ b/src/main/resources/avro/reads.avdl
@@ -77,7 +77,7 @@ record ReadGroup {
   union { null, string } description = null;
 
   /** The sample this read group's data was generated from. */
-  union { null, string } sampleId;
+  union { null, string } sampleName;
 
   /** The experiment used to generate this read group. */
   union { null, Experiment } experiment;
--- a/src/main/resources/avro/variantmethods.avdl
+++ b/src/main/resources/avro/variantmethods.avdl
@@ -70,12 +70,12 @@ record SearchVariantsRequest {
   string variantSetId;
 
   /**
-  Only return variant calls belonging to these columns.
+  Only return variant calls belonging to these samples.
   If an empty array, returns variants without any call objects.
   If null, returns all variant calls.
   TODO: rename
   */
-  union { null, array<string> } columnHandles = null;
+  union { null, array<string> } sampleNames = null;
 
   /** Required. Only return variants on this reference. */
   string referenceName;
@@ -113,8 +113,8 @@ record SearchVariantsRequest {
 record SearchVariantsResponse {
   /**
   The list of matching variants.
-  If the `columnHandle` field on the returned calls is not present, the
-  ordering of the columns from a `SearchVariantsRequest` over the parent
+  If the `sampleName` field on the returned calls is not present, the
+  ordering of the samples from a `SearchVariantsRequest` over the parent
   `VariantSet` is guaranteed to match the ordering of the calls on each
   `Variant`. The number of results will also be the same.
   */
--- a/src/main/resources/avro/variantmethods.avdl
+++ b/src/main/resources/avro/variantmethods.avdl
@@ -70,11 +70,12 @@ record SearchVariantsRequest {
   string variantSetId;
 
   /**
-  Only return variant calls which belong to call sets with these IDs.
+  Only return variant calls belonging to these columns.
   If an empty array, returns variants without any call objects.
   If null, returns all variant calls.
+  TODO: rename
   */
-  union { null, array<string> } callSetIds = null;
+  union { null, array<string> } columnHandles = null;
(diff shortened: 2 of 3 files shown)


dglazer:
@lh3 , do you have a set of methods in mind for how the `CallingMethod` object would be used? As it stands it's an orphan object, since no other object references it, and no method returns it.

I *think* the logical methods to provide would be:
- Search, which returns a filtered list of the `CallingMethod`s for a particular `VariantSet`
- some way to get the `CallingMethod` associated with a particular column-of-calls in a `VariantSet`, which either requires adding an id to `CallingMethod` and storing that id in the column, or implementing one of the in-flight discussions of formalizing `AnalysisSample`

But if we add those we end up back to something almost identical to today's `CallSet`, so maybe you have some other pattern in mind?

lh3:
> Search, which returns a filtered list of the CallingMethods for a particular VariantSet

The methods will be:
```java
record SearchCallingMethodRequest { string variantSetId; array<string> sampleNames = []; }
record SearchCallingMethodResponse { array<CallingMethod> methods; }
```

> some way to get the CallingMethod associated with a particular column-of-calls in a VariantSet

A column is a sampleName. You can unambiguously identify a CallingMethod with a (variantSetId,sampleName) tuple. ID is not needed in principle.

> But if we add those we end up back to something almost identical to today's CallSet, so maybe you have some other pattern in mind?

Even if we have an ID field in CallingMethod, it is different from CallSet in that we can retrieve variant calls while completely ignoring CallingMethod.

richarddurbin:
I am looking at this while on the phone with Jerome.
So in Heng's pattern as I understand it the same calling method can be applied to many (often all) sampleNames in the variantSet.  It is not the same thing as a current CallSet, which contains all the call data for the sample.  I think this makes sense.
We do need a proper Sample object!

lh3:
@richarddurbin and @jeromekelleher 

> the same calling method can be applied to many (often all) sampleNames in the variantSet

Not quite. Anyway...

Given all the confusion on the subtly different models, my suggestion would be to first reach a consensus on the minimal model that is sufficient for [1000G](http://www.1000genomes.org), [HRC](http://www.haplotype-reference-consortium.org), [ExAC](http://exac.broadinstitute.org) and [SGDP](https://www.simonsfoundation.org/life-sciences/simons-genome-diversity-project-dataset/), and then to add new objects for metadata such as time stamps, calling methods and sample relationship. The minimal model is this PR minus the CallingMethod object, which will be similar to #395 except the interpretation of columnHandler.

jeromekelleher:
@lh3 --- Is the idea to change ``sampleName`` to ``analysisSampleId`` in a follow-up, once we've decided on what the metadata-containing ``AnalysisSample`` is? As it stands, I'm afraid I can't really see much benefit to this change: we might as well add the ``CallingMethod`` to the current ``CallSet`` object so we formalise the provenance of the calls. 

What I really liked about your earlier models was that we explicitly linked up the various parts of the data model with links to real objects, which could be fleshed out with more details as time goes by. This seems like a step backwards from that, as we're now back to matching sample names. I'm probably misunderstanding though...

lh3:
I still think #383 is the only right model (which has AnalysisSample but not CallSet), but with all the discussions moved to #391 (which has both CallSet and AnalysisSample) and #395 (which has neither CallSet nor AnalysisSample), I am confused. This PR is essentially #395 but retaining meta information on calling in response to @dglazer, but in the end, I admit it is impropriate to discuss meta information without a consensus on the primary model first.

I am frequently confused by parallel issues on the same topic, so I will close this PR to reduce my own confusions. Sorry for the noises.

Issue #167, opened 2014-10-23
Globally Unique Content Digest


delagoya:
### Globally Unique Content Digest

This issue is for broader discussion with the developer community regarding **Globally Unique Content Digests* for identification of data objects and data sets that are currently being defined by the various schema.

The overarching goal is to provide an API centric method to be able to identify a set of data across different data providers or repositories of genomic data sets, so that researchers are able to identify exactly a data set that may have different versions. For example, there are many sequence data sets derived from the NA12878, alignments to different reference genomes, and variant analyses from these alignments. Each of the reference genomes, source sequence sets from sequencing a library, the resulting alignment sets, and variant calls should have some data digest to unambiguously identify it across repositories.

### Tenents

Any version of any set of genomic data in the world can have a digital digest.

**digests are unique**:
: no copy of a data set (data + metadata) will ever have a different digest, and no two different data sets will ever “collide” by accidentally getting the same digest

**digests are privacy preserving**
: given only the digest you can’t determine anything significant about the genomic data set it identifies, unless you are also authorized to use that digest to retrieve the data itself

**digests are decentralized**
: they're not assigned by any central authority; each organization that handles genomic data sets can generate the digests themselves

**digests are unforgeable**
: requesting the data with an ID should be deterministic. If you are not able to recompute the same digest from the the received data set then something went wrong

**accessions and digests are not mutually exclusive**
: accession numbers are widely used in biological data repositories and represent stable identifiers for data sets. Accessions can be assigned to a data set that also has a digest, the difference being that digests are computable from the data, whereas accessions are not computable from the data, but rather rely on the source repository that assigned the accession to the data set.

### Issues effected

Other issues that may be effected by this one:

- #159 : Best practices for joining NGS-derived & clinical variation databases
- #150 : Synchronisation tools
- #143 : Content (Data) management lifecycle
- #142 : GA4GH APIs need to address scientific reproducibility (propose immutable datatypes)
- #112 : Standardize GAReference.name for the benefit of other APIs

----

The original thread from the GA4GH Data Working Group mailing list is located at [https://groups.google.com/forum/#!topic/ga4gh-dwg/Baz7hMjytQg](https://groups.google.com/forum/#!topic/ga4gh-dwg/Baz7hMjytQg).

awz:
@delagoya Any reason this is being closed?

delagoya:
I closed for two reasons: 

1. No comments or work done on the issue since Nov 20, 2014. 
[...]

PR #397, opened 2015-09-02
Biosample to metadata master

mbaudis:
This is a skeleton implementation of the ```Biosample``` object, to provide a hook for e.g. reads/variants objects.


lh3:
Is there any uniqueness requirement on `Biosample.name`?

mbaudis:
@lh3 Wouldn't think so. While the schema can't express this (?), this goes into the general "id definition problem". My take on this (expressed partially in various discussions) is:

* id: required; should be sane & at least locally unique identifier, usually something "readable" (GSM2492834, BRCA-2015-DCIS0012 ...)
* guid: uuidv4, computer generated at creation 9recommended)
* name: a not necessarily "clean", "descriptive id"
* description: whatever free, unstructured annotation the user wants to use; explicitly not containing attribute::value pairs

The distinction between "name" and "id" seems a bit arbitrary/duplicative, it reflects common usage (e.g. an id + a sample name), while the guid is the enhancement.

src/main/resources/avro/metadata.avdl
```diff
@@ -92,6 +92,60 @@ record Experiment {
...
+
+  /**
+  External identifiers for sample, e.g. EBI BioSamples accession.
+  */
+  array<string> externalIds;
+
+  /**
+  A description of the biosample.
+  */
+  union { null, string } description = null;
```
jeromekelleher: Looks like a duplicated ``description`` field @mbaudis.

mbaudis:
@jeromekelleher Oops; fixed. Late here ...

jeromekelleher:
Looks sensible to me @mbaudis. I'd make the same comments I did on #399 about IDs vs names. Having IDs be server generated and opaque is very useful implementation-wise, and we'd certainly struggle to replace that in the reference server. I think ``name`` is a generally useful thing for users. I want somewhere to put 'NA12878' or '1kg-phase3'. 

I  agree about GUIDs, but I think this is probably the sort of thing we should implement after we sync up the schemas and start the new development model. Implementation will make it a lot easier to see what works and what doesn't.

lh3:
I asked this because the BioSample database appears to require global naming uniqueness (based on the existence of many related but different NA12878 "sample name": PB_NA12878-sc-1646934, E-MTAB-1884:NA12878, NA12878_30x, Platinum3, NA12878-Human, NA12878-sc-2011-01-04-7605, ... which I found messy). It is good that you are not thinking of the same type of "name".

helenp:
@lh3 the samples you flag there are from a source database that we import to BioSamples that does require naming uniqueness - they are are from ArrayExpress - Biosamples itself does not, just needs unqiue ids. As BioSamples is an aggregator of some existing resources as well as getting direct submissions there are some data source specific artifacts. ArrayExpress will switch to using BioSamples ids to get around this problem in future. I agree with your desire to have a name as well as id.

mbaudis:
I think there is some consent about having an id (which is controlled system wide at the resource provider) and additionally a name. The id may be computer generated or curated, but itself would be required & suitable for referencing, at least in a local context.

On the other hand, objects should have a guid, created at object generation time, in a consistent, "anonymous", collision free format. My vote right now would be for UUIDv4. Though I would vote for this to be a required feature, I see at least 2 points which would have to be discussed:
* privacy implications, if the guid is transmitted within otherwise anonymous objects but may lead to de-anonymization when co-occurring in another place
* overhead; i.e. do you really want this to be attached to every reads/variants ... object

Another concept would be to make the ```id``` attributes being the guid itself, and the name then being the "human readable/local specific" id. Other legacy ids could then be dumped not an ```aliases``` | ```accessions``` list.

So this is something for experts like @helenp to opine on ...

jeromekelleher:
@mbaudis --- we should also consider the possibility of GUIDs being cryptographic content hashes which are attached to every object (reads, readgroups, variants, variant sets, datasets, etc). This is a lot more useful than just locally creating UUIDs in my opinion as it enables cross-site comparisons of data, real guarantees of data identity, versioning, and many other benefits. There has been a _lot_ of discussion about this --- see  #167.

mbaudis:
@jeromekelleher Yes, but this would be a content/version id. For tracking of ```wet``` objects, a static GUID is much more practical, since changing its information content wouldn't break the id. Has been part of metadata discussions before; and I've mentioned the additional versioning/hash option/requirement in previous comments. We also have ```recordUpdateTime``` for sorting objects with the same guid.

Also, I'm more for offering a larger selection of well-defined attributes with clear roles, instead of "users" (i.e. local implementation developers) designing their own workarounds.

mbaudis:
... ```recordUpdateTime``` being implemented in the metadata branch.

```
    The time at which this record was created.
    Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
    TODO: The format definition is temporary and will be updated to
    the upcoming AVRO time format (timestamp-millis) when this is supported.
    */
    string recordCreateTime;

    /**
    The time at which this record was last updated.
    Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
    */
    string recordUpdateTime;
```

jeromekelleher:
@mbaudis --- fair enough, I hadn't thought about wet objects. Perhaps these could have uuids and a content hash? It's still useful to compare them for content equality quickly I think.

mbaudis:
@jeromekelleher Yes, exactly my opinion (id + name + guid + hash).

jeromekelleher:
@mbaudis Ah, OK. Sorry I'm a bit slow on catching up on all the metadata work!

mbaudis:
@jeromekelleher @diekhans @dglazer @helenp Following the discussions about id | guid | name | accessions, I agree that we should set up the general id attribute procedure first before specifying something beyond ```id``` + ```name``` in master.
I've changed the Biosample PR to just include ```id```, ```name```, ```description```. Pls. vote, for moving this into master.

jeromekelleher:
I think this sort of change needs to be done under the new development model we've agreed on --- we need to implement Biosample in the reference server and compliance suite before it can be added to master in the schemas. Changes to master going on right now are just bug-fixes to get us to the point where we can sync up the three repositories, get the current schema into a more consistent state and tag a version we can work from. 

So, I'm -1 on adding a change like this to master right now. BioSample sounds like an ideal topic branch though...

diekhans:
@mbaudis, what is the motivation of doing this PR now?  I know the background
clarification related the Heng's proposed changes to variants.  If  there
isn't a more immediate need, then perhaps it should go into the topic branch
that implements that change.

mbaudis:
While driven by the Variants discussion, ```Biosample``` is a general object on a level outside of ref/var. It should not be implemented there.

The PR is, as the integration of ```Dataset``` has been, part of a "bottom up" piecemeal integration of ongoing branch developments. A PR against variants would lead to possible collisions with other branches (but anyway, just wouldn't be *clean*).

mbaudis:
I have to reiterate my push for getting ```Biosample``` into the master branch. It represents the core reference object for any analysis object.

diekhans:
We need to get it implemented in the reference server and
compliance suites.  This should probably go in as part of a
work that actually needs it.

mbaudis:
@diekhans So who will take this up?

mbaudis:
I've changed from ```Biosample``` to ```BioSample```, as this seems to be the preferred spelling of the record's name (e.g. https://github.com/ga4gh/schemas/pull/427); also goes with ```AnalysisSample``` (https://github.com/ga4gh/schemas/pull/391).

mbaudis:
I've added fixes in ```reads.avdl``` and ```variants.avdl``` to get the record names in sync (```bioSampleId``` instead of ```sampleId```).

mbaudis:
Closing this since work has been moved to https://github.com/mbaudis/schemas =>

PR #399, opened 2015-09-02
Dataset ids

mbaudis:
This adds ```name```, ```guid``` etc. to the ```Dataset``` object. Still just a skeleton implementation.

### DIFF
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -9,7 +9,7 @@ protocol Metadata {
 import idl "common.avdl";
 
 /**
-An experimental preparation of a `Sample`.
+An experimental preparation of a `Biosample`.
 */
 record Experiment {
   /** The experiment UUID. This is globally unique. */
@@ -46,7 +46,7 @@ record Experiment {
   union { null, string } molecule = null;
 
   /**
-  The experiment technique or strategy applied to the sample.
+  The experiment technique or strategy applied to the biosample.
   (e.g. whole genome sequencing, RNA-seq, RIP-seq)
   */
   union { null, string } strategy = null;
@@ -123,15 +123,9 @@ what's the distribution of that result in the data from our project?)
 */
 record Dataset {
   /**
-  The dataset's id. This is the local/legacy/generic id.
+  The dataset's id, (at least) locally unique.
   */
   string id;
-
-  /**
-  The dataset's guid. This is globally unique id (UUIDv4).
-  TODO: Required?!
-  */
-  union { null, string } guid = null;
 
   /**
   The name of the dataset.
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -122,10 +122,25 @@ For data accessors, they're a simple way to scope exploration and analysis
 what's the distribution of that result in the data from our project?)   
 */
 record Dataset {
-  /** The dataset UUID. This is globally unique. */
+  /**
+  The dataset's id. This is the local/legacy/generic id.
+  */
   string id;
 
-  /** A description of the dataset. */
+  /**
+  The dataset's guid. This is globally unique id (UUIDv4).
+  TODO: Required?!
+  */
+  union { null, string } guid = null;
+
+  /**
+  The name of the dataset.
+  */
+  union { null, string } name = null;
+
+  /**
+  Additional, human-readable information on the dataset.
+  */
   union { null, string } description = null;
 
 }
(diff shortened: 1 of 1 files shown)


src/main/resources/avro/metadata.avdl
```diff
@@ -122,10 +122,25 @@ For data accessors, they're a simple way to scope exploration and analysis
 what's the distribution of that result in the data from our project?)   
 */
 record Dataset {
-  /** The dataset UUID. This is globally unique. */
+  /**
+  The dataset's id. This is the local/legacy/generic id.
```
jeromekelleher: I would leave out the second sentence in this comment. We will (hopefully) have a high-level document explaining the distinction between an ``id`` (a server generated string that is unique and persistent within a particular implementation, but not portable across implementations), and a ``name`` (a user specified string which is intended to provide a more human friendly handle for objects).

src/main/resources/avro/metadata.avdl
```diff
@@ -122,10 +122,25 @@ For data accessors, they're a simple way to scope exploration and analysis
...
+  */
   string id;
 
-  /** A description of the dataset. */
+  /**
+  The dataset's guid. This is globally unique id (UUIDv4).
+  TODO: Required?!
+  */
```
jeromekelleher: I think an overall strategy for assigning guids to objects should be devised, which looks at _all_ the types before we add it here. I would favour removing this for now, and explicitly tackling GUIDs across the board. I don't really see any advantage to the data set having a GUID if ReadGroupSets and VariantSets don't.

jeromekelleher:
Looks great, thanks @mbaudis! I've made a few comments above --- basically, I think we need to clarify id/name/guid at the top level. @diekhans, I'm sure you have something to say on this...

mbaudis:
@jeromekelleher 

     basically, I think we need to clarify id/name/guid at the top level.

Exactly; see my comment on https://github.com/ga4gh/schemas/pull/397 and elsewhere ...

mbaudis:
@jeromekelleher Also, keep in mind that we're working on the metadata branch https://github.com/ga4gh/schemas/blob/metadata/src/main/resources/avro/metadata.avdl

src/main/resources/avro/metadata.avdl
dglazer: I agree -- I like the idea of an API-wide GUID strategy; let's not go piecemeal until we have one.

diekhans: agreed @dglazer, a GUID system needs to be consistent across all
GA4GH object in order to implement a provenance system.

mbaudis: @dglazer @diekhans Please see https://github.com/ga4gh/schemas/pull/397 (and the metadata branch).

dglazer: I've seen it, and I think there's good discussions/work going on there, that's not yet ready to merge into master. I think you'll quickly get consensus on cleaning up the `id` comment and adding a `name` field now. Then the full-on guid conversation can continue at its own pace.

diekhans: @mbaudis, we need to have some API-wide design work done before
we defining any new features like GUIDs.  GA4GH doesn't even
document the meaning of `id' and `name', which has lead to
a lot of confusion.

For now, pull request should be the absolute minimum required to
get the baseline functionality working.

mbaudis:
@jeromekelleher @diekhans @dglazer @helenp
I've changed the PR to just add ```name``` to ```Dataset``` (and change the ```id`` comment to *locally*. Pls. vote, for moving this into master.

helenp:
+1

src/main/resources/avro/metadata.avdl
```diff
@@ -9,7 +9,7 @@ protocol Metadata {
 import idl "common.avdl";
 
 /**
-An experimental preparation of a `Sample`.
+An experimental preparation of a `Biosample`.
 */
```
jeromekelleher: Are these Biosample comments intentional? We don't have this defined, so perhaps it's better to revert these to Sample (even though that's not defined either!).

dglazer:
I agree with @jeromekelleher re reverting the "biosample" comment changes.

Other than that, +1.

mbaudis:
@dglazer @jeromekelleher Reverted the comments, but not feeling that good about something logically wrong & just doing it for procedural reasons.

```Sample``` will be ```Biosample``` (is so already on the metadata branch, with PR and votes); and the comment reference ```Sample``` from the ```Experiment``` object is void (as of right now) which wouldn't be different when changing it to any other name. Just going with this would have saved some overhead, even if I didn't document the reasoning.

PR #401, opened 2015-09-03
removal of Experiment

mbaudis:
```Experiment``` is an ill-defined metadata object, which should not be included in the current skeleton master branch.


lh3:
Agree it is ill-defined, but if we simply remove it, we'll have to put the library name (the @RG-LB tag) in the generic info field. Perhaps it is better to add a ReadGroup::library string field?

mbaudis:
@lh3 @diekhans The only reference to ```Experiment``` is in ```reads.avdl``` (thanks @diekhans).

I (not alone) never liked its definition (way too specific for certain experimental type). So my suggestion is to replace it with something outside of metadata (remember, *Everything but the Sequence*?) and remove ```Experiment``` for now (and possibly reuse/replace it later with something more generic).

diekhans:
-1 for now.

If it doesn't prevent current functionality from working, lets
just leave it alone until we can do a coherent design across all
of the GA4GH modules.  The information in Experiment will be
need for not read data (micro array).

mbaudis:
@diekhans While I understand *your don't fix what isn't broken* argument, it also doesn't fix what structurally isn't sound. And the microarray argument isn't that good, since the I'm worried about microarray compatibility & the current ```Experiment``` just doesn't fit.

We should do better definitions of which objects are technique dependent (current ```Experiment``` as ```Libraryprep``` or such part of ```reads.avdl```) and move them out of the shared metadata space/document. And then work on the common objects & attributes, including all the ID types, used throughout the schema. In that way either confirming metadata as common higher level, or splitting int *metadata* and *common* (e.g. ID definitions, documentation).

Closing this, since it would need reconfiguration across files/domains. But we should move the discussion to some common area?!

diekhans:
@mbaudis, I think it is broken and needs fixed.  My point is
don't do it piecemeal fashion without an overall, documented
design.

PR #403, opened 2015-09-09
Adds missing name field to VariantSet and ReferenceSet

jeromekelleher:
This PR adds a ``name`` field to VariantSet and ReferenceSet. This makes them consistent with all the other container types (Dataset, ReadGroupSet and ReadGroup). The name field is a user-specified, human readable identifier.

This is the last proposed change before the baseline milestone. It should be uncontroversial since

1. it's implementing the growing consensus on the need for names;
2. it has no backwards compatibility implications as it's just adding a new field.

It would be good if we could either get a quick consensus on this and merge it, or definitively drop it as part of the baseline.

### DIFF
--- a/src/main/resources/avro/references.avdl
+++ b/src/main/resources/avro/references.avdl
@@ -80,6 +80,9 @@ record ReferenceSet {
   /** The reference set ID. Unique in the repository. */
   string id;
 
+  /** The reference set name. */
+  union { null, string } name = null;
+
   /**
   Order-independent MD5 checksum which identifies this `ReferenceSet`.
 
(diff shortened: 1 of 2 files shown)


dglazer:
+1 -- I agree with both of your points (consistently having names is good, a new default-null field is backwards compatible).

mbaudis:
+ 1 For name and nullable and general implementation of this.

jeromekelleher:
@diekhans, @sarahhunt, any opinions here? It would be good to get this merged soon so we can import the final version of the schemas into the reference server and compliance tests.

diekhans:
+1 it seems like an oversight that it wasn't included

Issue #402, opened 2015-09-04
Visualizing GA4GH schema


lh3:
Echoing @diekhans' comment, a colleague of mine asked whether there is a visualized schema for GA4GH. I don't know how to draw a diagram for Avro (after some google searches), but I know there are tools to visualize SQL schemas, so I made one to capture the first-class objects:

![ga4gh-sql-v1](https://cloud.githubusercontent.com/assets/480346/9690858/ecf89ba6-530c-11e5-9d97-7cb094cbef69.png)

Note that the SQL schema is not a loyal representation of our Avro schema, because:
1. It is difficult to convert some Avro relationships (e.g. embedded arrays) to SQL as is. 
2. I have changed some parts based on my preferences.

diekhans:
Adam Novak wrote some code to generate UML diagrams from the schemas:
     https://github.com/ga4gh/schemas/blob/master/doc/uml.svg

The intention is to include something based on this in some upcoming
documentation work.

haussler:
Yes.  Heng's image looks a lot more appealing than the UML diagrams I've
seen before.

jeromekelleher:
This is really useful @lh3, thanks! This schema make a lot of sense to me.

One minor point: why not ``referenceId`` in Variant instead of ``refName``?

lh3:
> why not referenceId in Variant instead of refName?

Just followed the current schema without much thoughts. Yes, that is a mistake. The image is updated, along with refName in ReadAlignment changed to referenceId as well.

diekhans:
This is a lot better than the automatically generated UML, although conversion to SQL is a bit of difficult to maintain process.  We can have one of our people look around for a good direct UML tool.  

@lh3, what tool did you use for this.

lh3:
I am using [MySQL Workbench](https://dev.mysql.com/downloads/workbench/). The editable file underlying this image is [here](http://lh3lh3.users.sourceforge.net/download/ga4gh-sql-v1.mwb). Manually created image helps to highlight the key relationships between primary objects. Nonetheless, @adamnovak's UML diagram still looks nice. It may be worth generating a new one for the much simpler schema at master. It will look cleaner now.

reece:
Andrew Jesaitis created these in Gliffy:

![ga4gh_variant_annotation_schema](https://cloud.githubusercontent.com/assets/109453/11026887/8e9607f2-8664-11e5-9811-96bf721e9e0a.png)

[...]

PR #412, opened 2015-09-16
Added AnalysisSample; replaced first-class CallSet with auxiliary CallingMethod

lh3:
This PR borrows AnalysisSample from #383 and CallingMethod from #396. The relationships between first-class objects are approximately described by #402, with some relatively minor differences.

### DIFF
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -127,23 +127,15 @@ record Sample {
   string datasetId;
 
   /**
-  Sample name in the dataset. This name uniquely identifies the sample in the
-  dataset.
+  Submitted sample name in the dataset. This name uniquely identifies the
+  sample in the dataset.
   */
   string name;
 
   /**
-  The NCBI/EBI BioSample identifier of this sample. Different Sample objects
-  may have the same biosampleId.
+  Key-value pairs for additional sample information
   */
-  union { null, string } biosampleId = null;
-
-  /**
-  A globally unique ID, derived by some agreed upon scheme so that we can find
-  equivalent Sample's in different datasets and servers/implementations.
-  Typically, sequencing libraries and guids have a one-to-one relationship.
-  */
-  union { null, string } guid = null;
+  map<array<string>> info = {};
 }
 
 }
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -109,32 +109,4 @@ record Dataset {
 
 }
 
-/**
-A sample uniquely named in a Dataset. The sample name typically appears in a
-@RG line in a BAM file and in the sample header line of a VCF file. Unrelated
-samples in different Datasets may have the same sample name.
-*/
-record NamedSample {
-  /**
-  Identifier of this sample
-  */
-  string id;
-
-  /**
-  The identifier of the dataset this sample belongs to
-  */
-  string datasetId;
-
-  /**
-  Submitted sample name in the dataset. This name uniquely identifies the
-  sample in the dataset.
-  */
-  string name;
-
-  /**
-  Key-value pairs for additional sample information
-  */
-  map<array<string>> info = {};
-}
-
 }
--- a/src/main/resources/avro/metadata.avdl
+++ b/src/main/resources/avro/metadata.avdl
@@ -109,4 +109,44 @@ record Dataset {
 
 }
 
+/**
+In a Dataset, a sample is identified by its name that shall be unique in the
+Dataset. This name typically appears in a @RG line in a BAM file and in the
+sample header line of a VCF file. Unrelated samples in different Datasets may
+have the same name. DatasetSample is only meant to be embedded in SampleGroup,
+not referenced by other objects.
+*/
+record DatasetSample {
+
+  string datasetId;
+
+  /**
+  The sample name in the Dataset referred by datasetId
+  */
+  string sampleName;
+}
+
+/**
+A SampleGroup connects samples in different Datasets that have the same origin.
+A typical scenario is that samples from several Datasets are aggregated and
+reanalyzed, which result in a new Dataset.
+*/
+record SampleGroup {
+  /**
+  ID of this sample group
+  */
+  string id;
+
+  /**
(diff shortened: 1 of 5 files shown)


src/main/resources/avro/metadata.avdl
```diff
@@ -109,4 +109,32 @@ record Dataset {
...
+  string id;
+
+  /**
+  The identifier of the dataset this sample belongs to
+  */
+  string datasetId;
+
+  /**
+  Submitted sample name in the dataset. This name uniquely identifies the
+  sample in the dataset.
```
dglazer: consistency point: the API-wide pattern is that names are human-provided and don't have to be unique, while ids are server-generated and must be unique.  It's certainly good practice to use unique descriptive names, but that's because they're human-readable, not technical reasons.

src/main/resources/avro/variantmethods.avdl
```diff
@@ -70,11 +70,11 @@ record SearchVariantsRequest {
   string variantSetId;
 
   /**
-  Only return variant calls which belong to call sets with these IDs.
+  Only return variant calls which belong to samples with these IDs.
   If an empty array, returns variants without any call objects.
   If null, returns all variant calls.
   */
-  union { null, array<string> } callSetIds = null;
+  union { null, array<string> } sampleNames = null;
```
dglazer: these should be sampleIds (per API-wide use of ids and names)

src/main/resources/avro/variantmethods.avdl
```diff
@@ -148,68 +148,3 @@ org.ga4gh.models.Variant getVariant(
   The ID of the `Variant`.
   */
   string id) throws GAException;
-
-/******************  /callsets/search  *********************/
```
dglazer: How do I get a list of all the columns in a variantset without some equivalent of this method?

src/main/resources/avro/variants.avdl
```diff
@@ -41,7 +41,30 @@ record VariantSetMetadata {
 }
 
 /**
-`Variant` and `CallSet` both belong to a `VariantSet`.
+Method used for variant calling
+*/
+record CallingMethod {
```
dglazer: I might be missing it, but this looks like an orphan object -- it's not referenced by any other objects or methods. Without that, I don't understand how it's meant to be used.

src/main/resources/avro/metadata.avdl
lh3: It is not just a good practice. It is a basic requirement. We will certainly have a query "give me the genotypes of NA12878". If "NA12878" is not an unambiguous sample label, this query will be impossible. Note that similarly in references.avdl, references in a referenceSet ought to have different names; otherwise it will be very tricky to use references.

mbaudis: @lh3 Since we develop for distributed data access, the only unambiguous identifiers will be GUIDs. And even those won't prevent duplicate objects on a "global" scale.

src/main/resources/avro/variantmethods.avdl
lh3: I won't explain the technical reasons – your example is very rare in practice; especially in the final data release, we almost never want to put two calling methods in one VCF/VariantSet. I don't mind if API doesn't work elegantly in this extreme case (note that it still works, just not elegantly). In contrast, the current API model is complicating the most common use cases. It actually doesn't really resolve your example if you simply take VCF and SAM submissions.

src/main/resources/avro/variants.avdl
lh3: I am not sure if it matters either way. It just describes meta information. If you insist, I could add `array<CallingMethod> callingMethods` to `VariantSet`.

src/main/resources/avro/metadata.avdl
lh3: Think about submissions and user queries. It will be a complete mess if we have two "chr1" in a referenceSet or two distinct samples labeled "NA12878" in a dataset.

mbaudis:
This is a very deep change to master, affecting metadata etc. It would actually be nice to address the (basically agreed upon, only not procedurally) ```Biosample``` integration https://github.com/ga4gh/schemas/pull/397 first; this may than lead to a better understanding of pro|con ```AnalysisSample```.

src/main/resources/avro/variantmethods.avdl
lh3: There will be the following in metadatamethods.avdl (which doesn't exist right now):
```java
record SearchAnalysisSamplesRequest {
  string datasetId; // not null
  array<string> sampleNames = [];
}
record SearchAnalysisSamplesResponse {
  array<string> analysisSampleIds;
}
```

lh3:
@mbaudis Within the scope of a dataset, this PR works and is self-consistent without BioSample. BioSample will be needed in future to support cross-dataset queries, but it is a separate issue. Also note that the current schema doesn't use BioSample (@dglazer said that sampleId is the user submitted name, not an ID). This PR is at least not worse than the current schema.

src/main/resources/avro/metadata.avdl
lh3: @mbaudis also note that I am not requiring global uniqueness. It is local uniqueness: reference names are unique within a ReferenceSet; analysisSample names are unique within a dataset.

src/main/resources/avro/variantmethods.avdl
diekhans: @lh3, your example query doesn't need to replace the callset
query, it can be an alternative.  Since I wasn't on the
conference calls, I have no idea what use cases were used to
define the queries.

My contentions is a clean conceptual data model that actually
represents the underlying data and build queries for particular
use cases will be far more satisfactory than trying to structure
the data around today's common queries.

mlin:
@lh3 a couple syntax fixups for this: https://github.com/dnanexus-rnd/schemas/commit/652febc4988b570e842242f9d834b077fac5b9e7

src/main/resources/avro/metadata.avdl
pcingola: I'm +1 for locally unique names since it enforces good practices. 
I minor nitpick: I think we should we enforce non-empty names as well.

Issue #408, opened 2015-09-14
Review and update variants documentation


diekhans:
In documentation branch: review, address FIXMEs, and improve documentation on vaiants.  In particular, add description of data model to variants introduction page.

Files are:
- src/main/resources/avro/variants.avdl
- doc/source/intro_variants.rst

diekhans:
see also:  https://github.com/ga4gh/schemas/issues/379

PR #417, opened 2015-09-19
first cut at API goals and design principles

dglazer:
@diekhans -- here's an initial set of goals and design principles for the overview doc.

### DIFF
--- a/doc/source/intro_apidesign.rst
+++ b/doc/source/intro_apidesign.rst
@@ -1,7 +1,34 @@
 .. _apidesign:
 
 ****************************
-GA4GH API Design (empty)
+GA4GH API Design
 ****************************
+=============
+Object Ids
+=============
 
+* Many objects need to be referenced by the API (e.g. a search method may return a list of objects) and by applications built outside the API (e.g. a visualization app may refer back to the objects being shown). The API has a standard mechanism for referring to such objects.
 
+* The standard way to programmatically reference an object is via an **id** field. The id is server-assigned, and must provide “id-like” semantics, including:
+
+  * ids are unique within the scope of the server instance
+  * ids are durable for the lifetime of the server, and persistent across server restarts -- once a user is given an ID for data stored on a server, the id remains valid for as long as the server is still storing that data
+
+* Many objects, including most ‘container’ objects, also have a **name** field. The name is user-defined, isn’t programmatically required to be unique in any given scope (although in practice data owners will often choose unique names), and is intended to be human readable.  This can be thought of as a display name.
+
+Cross-repository data federation will need a standard way to refer to a data object, regardless of which repository it’s in. There is no such standard currently, and the current API, including the id and name fields, isn’t sufficient.
+A future API may introduce standard cross-repository identifiers using some combination of **content hashes**, **GUIDs**, and central **accession** facilities.
+
+=====================
+Object Relationships
+=====================
+
+* Some objects are contained by other objects. These relationships can be
+
+  * many:1 (e.g. ReadGroupSets in a Dataset); aka single-include
+  * many:many (e.g. ReadGroups in a ReadGroupSet); aka multi-include
+
+* Some objects are derived from other objects. These relationships can be
+
+  * many:1 (e.g different aligned ReadGroupSet’s derived from an unaligned ReadGroupSet using different alignment algorithms and/or reference sequences)
+  * many:many (e.g. different VariantSets derived from a collection of ReadGroupSets using different joint variant calling algorithms)
--- a/doc/source/intro_apigoals.rst
+++ b/doc/source/intro_apigoals.rst
@@ -1,7 +1,29 @@
 .. _apigoals:
 
 ***************************
-GA4GH API Goals (empty)
+GA4GH API Goals
 ***************************
 
 
+* From the `GA4GH DWG <http://ga4gh.org/#/documentation>`_ site:
+
+    The Global Alliance for Genomics and Health (GA4GH) Genomics [API] will allow the interoperable exchange of genomic information across multiple organizations and on multiple platforms. This is a freely available open standard for interoperability, that uses common web protocols to support serving and sharing of data on DNA sequences and genomic variation. The API is implemented as a webservice to create a data source which may be integrated into visualization software, web-based genomics portals or processed as part of genomic analysis pipelines. It overcomes the barriers of incompatible infrastructure between organizations and institutions to enable DNA data providers and consumers to better share genomic data and work together on a global scale, advancing genome research and clinical application.
+
+* The API must allow flexibility in server implementation, including:
+
+  * choice of persistent backend (e.g. files, SQL, NoSQL)
+  * choice of implementation language (e.g. Java, Python, Go)
+  * choice of authorization model (e.g. all public, all private, fine-grain ACLs)
+  * choice of import mechanism (e.g. self-service vs. centrally managed)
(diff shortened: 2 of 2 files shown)


lh3:
This PR is too implementer-centric. From the view point of users, they would not care too much about flexibility in server implementation.

> The name is user-defined, isn’t programmatically required to be unique in any given scope

Do we really allow two "chr1" in a ReferenceSet? If a user asks "is there a SNP at chr1:1000?", how can he/she query? Some names are query-able, not just for display purposes. In that case, they act like UNIQUE keys in SQL. Even if you found a way to allow querying a ReferenceSet with two "chr1", it would lead to many pitfalls.

> The API must allow flexibility in server implementation

I would say something like "The API *should* allow flexibility in server implementation without impacting performance and user experiences". We could have designed SAM in XML to allow maximal flexibility, but few would have wanted to use that. There is a balance.

> The API must allow full-fidelity representation of data that was prepared using today’s common methods and stored using today’s common file formats.

I would change "must" to "should".

> whenever the semantics are clear, including when they're specified in valid data files, the API must allow preserving them.

Does "them" refer to all kinds of user submitted data? If so, I would change "must" to "should" at least. To play devil's advocate, I may even remove this note. The API aims to encourage data sharing. Keeping all data is a desired feature, but not a required one. Some data are typically not represented in the form of SAM and VCF (e.g. population of 1000g samples). If the API even can't keep such important data, why do we *require* it to keep other minor information?

> The API should allow adding more structure to data beyond today’s common practices

I would say "The API should match today's common practices, but allow adding more structure if common practices are inadequate". Matching common practices is the bases of extending them.

<hr>

In general, this PR is too implementer-centric. If I were to draft "GA4GH API Design", I would first say that the API largely (not precisely) matches our common practice, and briefly explain how bioinformaticians today can map their concepts to API objects. I would not emphasize too much on "flexibility in server implementation" in Design Goals. I would instead emphasize how users can get consistent results across implementations. I more like a user-centric view when writing such documentations.

diekhans:
Heng Li <notifications@github.com> writes:
> This PR is too implementer-centric. From the view point of users, they would
> not care too much about flexibility in server implementation.

The documentation is meant to comprehensive and will be broken
up into more user centric and server centric pieces.   The goals
are the goals for the whole API.

This is going to be an on going process to resurrect information
that was lost from the original discussions.

One we can actually read what was decided, then we can revisit
those decisions and *why*.  We don't even agree on the role of
Avro.  Some people think it's an IDL, some people think it's a
poor substitute for LaTeX.


>     The name is user-defined, isn’t programmatically required to be unique in
>     any given scope
> 
> Do we really allow two "chr1" in a ReferenceSet?

Yes.  but it would be horrible.  We need to capture the initial design
first.  This is, apparently, what that task team designed, but didn't
write down.

@lh3, since the VCF data model is not written down anywhere and you
are the first person I have ever heard say the was one, can you
take the first crack at writing it down:

      https://github.com/ga4gh/schemas/issues/408

Claim the ticket and fork the documentation branch.

Please include not only a detailed design, but rational.
Developers and users need to know why something is a certain
way.  This ends multi-month discussions about VCF columns being
independent We have to capture the why's as well as the how's.

Feel free to get others who understand the current VCF data
model involved.

haussler:
Great start!  Thanks David!  Keep it coming. -D

lh3:
> This [allowing duplicated reference names within a ReferenceSet] is, apparently, what that task team designed, but didn't write down.

The history of this reference naming uniqueness is a little interesting. When @richarddurbin first proposed Reference, he [said specifically][1]: "[Reference::name] must be unique within any ReferenceSequenceSet that contains this". He [moved Reference::name][2] to ReferenceSet::names a few days later. The rationale was explained [in this comment][4] with no intention to allow duplicated reference names. When we tried to tidy up Reference, we [added back Reference::name][3] without discussions (as I see it now, we overlooked why Richard deleted Reference::name; adding a name is actually changing the data model). Dropping the uniqueness requirement is not the task team design. It just slipped in, but we now interpret it as ground truth. In general, we should not be bond to the schema at the current stage. Just do the right thing. If allowing duplicated reference names is "horrible", then forbid it. We don't want a horrible API, do we?

BTW, reads and variants APIs are using refinance names. Allowing duplicated Reference::name will break everything. This strongly suggests that by design, we don't want duplicated reference names in a ReferenceSet.

> since the VCF data model is not written down anywhere and you are the first person I have ever heard say the was one

If we decide to follow the direction of #412 or #402, I can help the documentation (though I am yet to understand what you want to achieve exactly).

[1]: https://github.com/ga4gh/schemas/commit/935d1cb67a2fc8917a25e5ed2933e8c9dfde11e4#diff-54e8f45ccf984ee2f691ca4ceeb30c6eR7
[2]: https://github.com/ga4gh/schemas/commit/b5a3f86adf976835ceffa90dabff0cec7614afa5#diff-54e8f45ccf984ee2f691ca4ceeb30c6eL7
[3]: https://github.com/ga4gh/schemas/pull/105
[4]: https://github.com/ga4gh/schemas/pull/66#issuecomment-44501533

diekhans:
Heng Li <notifications@github.com> writes:
> The history of this reference naming uniqueness is a little interesting.

that it is

> When model). Dropping the uniqueness requirement is not the
> task team

They didn't document it, so uniqueness or non-uniqueness is undefined
and left to the implementor to guess.

> If we decide to follow the direction of #412 or #402, I can help the
> documentation (though I am yet to understand what you want to achieve exactly).

Document a design before implementing it so we don't have these
kind of misunderstanding.

The current variant schemas has been around for over a year and
no one noticed something was wrong.  A schema isn't a design.

jeromekelleher:
> BTW, reads and variants APIs are using reference names. Allowing duplicated Reference::name will break everything. This strongly suggests that by design, we don't want duplicated reference names in a ReferenceSet.

I don't have a strong opinion on whether we should *require* names to be unique or simply *encourage* it. Practically, I don't think it makes much difference, since all API linkage points should be through IDs and not names. The current API uses ``referenceId`` when doing a read search in any event. What is the argument for not enforcing unique names within a local collection @dglazer? 

I think the text in this PR is a good starting point, but I wonder about the 'shoulds' and 'musts'. Perhaps we should just state these as facts, rather than requirements? For example, we could just say

> The API allows flexibility in server implementation.

As it is, it reads a little oddly, like a mixture of a requirements specification and a description of the actual artifact. Other than that, I think this is a great start which we should aim to refine and update.

lh3:
@diekhans was asking what is the initial design. As of now, both reads and variants are using referenceName instead of referenceId. This clearly shows that by design, we have been always assuming referenceNames are unique within a ReferenceSet. Changing them to ID in future could work to some extent, but would be very confusing. Think about the workflow. Suppose we want to get SNPs on chr1. We retrieve the ReferenceSet and find it has two "chr1". What should we do next? For another example, suppose in a paper we see sample "S_French-1" is inbred. Now we look up this sample with the current SearchCallSet() and get two callSet IDs. What should we do? The key observation is that some names are used in query, not just in display. To achieve sensible queries, it is best for such names to have local uniqueness.

@dglazer on naming uniqueness, please respond this time. I have given such examples several times but never got followup.

Design goals are what we want to achieve, which may or may not have been achieved yet. They are not facts.

diekhans:
@lh3 I am asking for a design specification, not examples in email.

I use to think that id was as you describe name (required to be
unique) and name was a human display.  I wasn't in the
smoke-filled conference call that designed otherwise and it was
never documented.

Issue #87, opened 2014-07-02
Do we need the GA prefix on all objects?


cassiedoll:
All of our Avro files are namespaced under ga4gh, so a read alignment is currently `org.ga4gh.GAReadAlignment`

So the GA seems a little redundant from a packaging perspective, and is making all of the code/docs a bit harder to read (and type :) 

What does everyone think about removing GA from everything? 
We'd then have `org.ga4gh.ReadAlignment`

PR #419, opened 2015-09-21
Documented error conditions in SearchReadGroupSetsRequest.

hjellinek:
More to come.
See [compliance issue #87](https://github.com/ga4gh/compliance/issues/87).


diekhans:
Technically, the API doesn't throw.  It returns http error and with the GAException in the body.  We should come up with boiler plate text to fill in.

hjellinek:
Agree with you about boilerplate.  I figure after I or others have written a few of these, we'll come up with the right wording and formatting.

diekhans:
How about updating this to be clear there is no throw and it probably become the boilerplate.

Issue #418, opened 2015-09-21
The big name | ID | UUID | version ID discussion


mbaudis:
In many PRs we're now running into very differing views on when and how to use object identifiers. We, for all different areas/task teams, have to settle on one general way to use object identifiers (OID).

This discussion here does not address each object's need to implement a given type of OID, just the relationships between OID attribute name and value, and the general OID usage.

This discussion starts also based on the assumption that OID usage will not depend on a GA4GH provided arbitration system. However, this does not mean that such services won't be implemented.

Issues in which the proposed OID differ can be separated into

* uniqueness (global/local/none)
* content dependency (partial/full content hashing)
* generation (machine vs. readable/legacy)

This opens following attributes for discussion:

* id
  * required (exception to the "object's need for OID type not discussed here")
  * locally unique
  * usually something "readable" (GSM2492834, BRCA-2015-DCIS0012 ...)
  * main type for (local) references
* guid
  * UUIDv4
  * assumed to be globally unique without central arbitration, due to its structure
  * computer generated at creation
  * global references (though no guarantee that not duplication of another object)
* name
  * not necessarily "clean"; possibly descriptive ("patient #25, PB")
  * important for legacy data & "human parsing"
  * not recommended for object references
* versionID
  * (partial/complete) object hash
  * version management (e.g. combined with time stamps)
  * global references with absolut uniqueness

The main points which I predict tobe discussed here are the abstaining from using ```name``` for referencing (which is different from some PRs) and the guid/UUID types.

[
In principle, this could be collapsed by using ```guid``` type identifiers for ```id```. However, this would provide many headaches for both legacy data & privacy (e.g. you may want to pass on datasets with your locally unique identifiers, without allowing easy global identification of data references with respect to other resources).
]

jeromekelleher:
Thanks for kicking this off  @mbaudis, this is a key topic. I think you've summarised things very well here. 

Collapsing the GUID and ID would be problematic for implementers I think, as it is useful to have control of the ID structure. Certainly it would give us major headaches in the reference server. For example, we use variant IDs of the form ``base64encode(datasetName:variantSetName:variantName)``, where ``variantName`` is something like ``chr:pos:alt``. When we get a GET variant request for a given ID, we then use this structure to find the right dataset, variantSet and variant. If we didn't have this structure, we would have generate a GUID for every variant and read, and to maintain a global index of all GUID -> variant/read mappings, which would be rather large.

[...]

PR #420, opened 2015-09-22
Changed referenceName to referenceId.

jeromekelleher:
This PR follows on from recent discussions in #418, #417, #399, #403. There is a seemingly growing consensus that IDs should be used for all API linkage points, and names are more human-readable local identifiers which might be locally unique within a collection or not; see #417 for the discussion on this point.

Whether names are locally unique or not, I don't think they should be used as API object linkage. The Variants API currently uses `referenceName` to identify the Reference object within a given ReferenceSet that the variant is defined in terms of. This is inconsistent with other parts of the API.

The only argument that I can see in favour of using `referenceName` here rather than `referenceId` is that it's slightly more convenient for client side developers, since they can hard code the string  "1" (e.g.) instead of needing to lookup which reference object corresponds to this name. However, this only really applies to the case in which there's a well known ReferenceSet, with a small number of References. For most general applications, the client would have to look up the ReferenceSet to find the names of the contained references in any case, so it doesn't matter if we use `referenceName` or `referenceId`.

Using `referenceName` encourages poor programming practise, where we hard code the names of the chromosomes into code rather than looking them up. It is based on the idea of a small number of universally known, static references, which is something that will become less true over time, particularly when we think of graph references.

### DIFF
--- a/src/main/resources/avro/common.avdl
+++ b/src/main/resources/avro/common.avdl
@@ -22,9 +22,9 @@ represented by a `Reference` name, and a base number on that `Reference`
 */
 record Position {
   /**
-  The name of the `Reference` on which the `Position` is located.
+  The ID of the `Reference` on which the `Position` is located.
   */
-  string referenceName;
+  string referenceId;
 
   /**
   The 0-based offset from the start of the forward strand for that `Reference`.
--- a/src/main/resources/avro/variantmethods.avdl
+++ b/src/main/resources/avro/variantmethods.avdl
@@ -77,7 +77,7 @@ record SearchVariantsRequest {
   union { null, array<string> } callSetIds = null;
 
   /** Required. Only return variants on this reference. */
-  string referenceName;
+  string referenceId;
 
   /**
   Required. The beginning of the window (0-based, inclusive) for
--- a/src/main/resources/avro/variants.avdl
+++ b/src/main/resources/avro/variants.avdl
@@ -196,9 +196,8 @@ record Variant {
 
   /**
   The reference on which this variant occurs.
-  (e.g. `chr20` or `X`)
   */
-  string referenceName;
+  string referenceId;
 
   /**
   The start position at which this variant occurs (0-based).


mbaudis:
+1 for the general "if referencing use *xxxId*, not *xxxName*. No special opinion on the variants implement (out of my scope).

lh3:
Using id is the right choice on modeling, but it hurts user experiences in particular for the JSON interface. Think about the UCSC genome browser. It largely uses names only (note that for SQL, id and name are just a JOIN away, much more convenient than JSON). That said, if others unanimously think using ID is right approach, I can live with that, so -0 overall.

BTW, reads are using referenceNames. For consistency, that needs to be changed, too.

jeromekelleher:
I think reads already use ``referenceId`` @lh3 --- the second referenceName argument was removed as it was redundant (see #193) .

In my view, it's a mistake to think of end-users working with the JSON. To do any real work, a client library is needed to hide low-level network details of paging or streaming (just imagine how painful trying to do anything directly with the JSON in C would be, for example). The client library is going to take the JSON (or any other on-the-wire format) and parse this into language native structures. In C, we'd want structs that correspond directly to the protocol objects; in Java or Python these would be objects. In this world, we might imagine making the ``Reference`` object available directly within the returned ``Variant`` objects and not just the IDs. This would be very convenient for the client-side programmer.

Anyway, my point here is that I think we should regard this protocol as a low-level building block which we can make nice client libraries out of,  not as something we imagine users interacting with directly. As a low-level serialisation format, the use of IDs as the proper choice for linkage is clear I think.

lh3:
You can find referenceName in this chain: ReadAlignment->LinearAlignment->Position->referenceName. Reads are using referenceName.

Probably you remember our disagreement on the client library: I prefer very thin client libraries as my experience is that heavily relying on complex language-specific client libraries will slow the adoption. Anyway, that is another issue. Just to say I have a different opinion, but I won't block on this issue.

jeromekelleher:
Aha! Thanks @lh3, I missed ``Position.referenceName``. I've added this to the PR as we might as well tackle all of these at once.

I'm on the fence about whether we include this change (if it's accepted) in the upcoming API sync; I think it would be good to include it, but we also have to stop somewhere... @dglazer and @calbach, how painful is this for backwards compatibility?

calbach:
Thanks @jeromekelleher. This kind of change is not too bad, since we could technically choose to continue supporting `referenceName` in addition to `referenceId` while still implementing the specification (with the possible exception of `SearchVariants` since `referenceId` would be a required field).

I'm also on the fence about this change. I'm pretty firmly in support of thin, ideally autogenerated, client libraries (this is the general policy at Google as well). Every piece of functionality you intend to add on top of the API is something you will have to write and (more importantly) maintain in N different languages.

diekhans:
@calbach without higher level libraries, application programmers
are working at less expressive level and duplicating effort.
This leads to the temptation to add more to the API to make it
easier to use without client support.

However, this doesn't mean one group should be providing all the
client libraries either as part of a release..  Beyond the need
for the resources to do it, one group is most likely not going
to have the expertises in a variety of languages to do a good
job.

IMHO, it's a matter of encouraging client library development
not requiring it.

jeromekelleher:
@dglazer, it would be good to get your thoughts here. What do you think about changing all our current ``referenceName``s to ``referenceId``s?

(Apologies for the digression on client libraries; I think we're all on the same page here.)

david4096:
Just noticed this had been proposed previously. https://github.com/ga4gh/schemas/pull/616 is an updated version that includes the variant annotations and sequence annotations protocols. @diekhans

PR #422, opened 2015-09-22
Added descriptions of error conditions

hjellinek:
Added descriptions of error conditions to `getDataset`, `SearchReadsRequest`, `ListReferenceBasesRequest`, `SearchCallSetsRequest`, `getCallSet`.

The documented behavior had been discussed and agreed to in email and issue threads, but not mentioned in the schemas until now.


src/main/resources/avro/readmethods.avdl
```diff
@@ -12,6 +12,11 @@ This request maps to the body of `POST /reads/search` as JSON.
 If a reference is specified, all queried `ReadGroup`s must be aligned
 to `ReferenceSet`s containing that same `Reference`. If no reference is
 specified, all `ReadGroup`s must be aligned to the same `ReferenceSet`.
+
+_Error Conditions_
+
+- If `readGroupIds` is empty, this method will throw `GAException` and set the HTTP
```
jeromekelleher: I don't think this is a 501 condition. This is Bad Request, so should be a 400. Or maybe something more specific? I can't see any code that fits this condition better than 400 at http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

jeromekelleher: Also, I don't think `throw` is the right word here (implies exception handling). I think we should just say ``return a GAException``.

src/main/resources/avro/referencemethods.avdl
```diff
@@ -169,6 +169,11 @@ The query parameters for a request to `GET /references/{id}/bases`, for
 example:
 
 `GET /references/{id}/bases?start=100&end=200`
+
+_Error Conditions_
+
+- If `start` or `end` exceed the index of the last base of the `Reference`, this method will throw `GAException` and set the HTTP
```
jeromekelleher: Is line width > 80 chars here? (We really need to add some CI checks for this!)

jeromekelleher: I don't think this quite captures it. What about
> If ``start`` or ``end`` are negative, or greater than the length of the reference, this method ....

src/main/resources/avro/variantmethods.avdl
```diff
@@ -150,7 +150,13 @@ org.ga4gh.models.Variant getVariant(
   string id) throws GAException;
 
 /******************  /callsets/search  *********************/
-/** This request maps to the body of `POST /callsets/search` as JSON. */
+/** This request maps to the body of `POST /callsets/search` as JSON.
+
+_Error Conditions_
+
+- If `variantSetId` does not refer to an existing `VariantSet`, this method will throw `GAException` and set the HTTP
```
jeromekelleher: Line width again.

src/main/resources/avro/referencemethods.avdl
diekhans: how about `out of range of the reference?

src/main/resources/avro/readmethods.avdl
diekhans: Could we start a page in the documentation branch that generally describes
what HTTP errors should be uses in what conditions?

src/main/resources/avro/referencemethods.avdl
jeromekelleher: Yeah, that's good.

src/main/resources/avro/readmethods.avdl
calbach: Definitely a 400. I agree this documentation should live at a more general level, we have many places in the schemas where we say a parameter is "required". If these parameters are not provided, the server should return a 400.

I can take a stab at adding a section about this somewhere, but I'm not sure where it should live in the documentation branch.

PR #423, opened 2015-09-23
Create intro_commonattributeformats.rst

mbaudis:
This is a proposal to define common attribute::value usage inside the documentation system. This first document takes up the ID discussion from https://github.com/ga4gh/schemas/issues/418

### DIFF
--- a/doc/source/intro_commonattributeformats.rst
+++ b/doc/source/intro_commonattributeformats.rst
@@ -5,32 +5,39 @@ Formats for object identifiers (OID)
 Object identifiers have properties based on varying requirements, 
 depending on their usage. Main features to be considered are:
 
-* uniqueness (global/local/none)
-* content dependency (partial/full content hashing)
-* generation (machine vs. readable/legacy)
+- uniqueness (global/local/none)
+- content dependency (partial/full content hashing)
+- generation (machine vs. readable/legacy)
+
+//FIXME: the GUID question
 
 Following attributes are used as record identifiers and descriptors:
 
-* id
-  * required
-  * locally unique
-  * systematic and/or "readable" in local context (GSM2492834, BRCA-2015-DCIS0012; partial hash ...)
-  * main type for (local) references
-* guid
-  * UUIDv4
-  * assumed to be globally unique without central arbitration, due to its structure
-  * computer generated at creation
-  * global references (though no guarantee that not duplication of another object)
-* name
-  * not necessarily "clean"; possibly descriptive ("patient_25, PB")
-  * important for legacy data & "human parsing"
-  * not recommended for object references
-* versionId
-  * (partial/complete) object hash
-  * version management (e.g. combined with time stamps)
-  * global references with absolut uniqueness
-* accessions
-  * list of one or more IDs or URIs for public exposure of the object
-* description
-  * free text description of the object's scope
+id
+--
+- required
+- (at least) locally unique
+- systematic and/or machine generated
+- possibly, but not necessarily "human readable" (GSM2492834, BRCA-2015-DCIS0012 ...)
+- main type for (local) references
+
+name
+----
+- not necessarily "clean"; possibly descriptive ("patient_25, PB")
+- important for legacy data & "human parsing"
+- not recommended for object references
+  
+versionId
+---------
+- (partial/complete) object hash
+- version management (e.g. combined with time stamps)
+- global references with absolut uniqueness
+
+accessions
+----------
+- list of one or more IDs or URIs for public exposure of the object
+
+description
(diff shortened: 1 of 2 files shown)


diekhans:
thanks @mbaudis .  A lot of this is a proposal (guid and versioning), not part of the current master API.  This should be in the appropriate topic branch (metadata in this case).  We need to capture the current design first.

This document needs to be filled out:
https://github.com/ga4gh/schemas/blob/documentation/doc/source/intro_apidesign.rst

mbaudis:
There is no current master API, is there? And if not, this just could be part of it, shouldn't it?

mbaudis:
This document needs to be filled out:
    https://github.com/ga4gh/schemas/blob/documentation/doc/source/intro_apidesign.rst

So, who will do this? This isn't just metadata.

mbaudis:
O.k.; everybody... (answered in the metadata call).

diekhans:
Overall design documentation should go here:

   https://github.com/ga4gh/schemas/blob/documentation/doc/source/intro_apidesign.rst

this will need contribution from lots of people.

lh3:
It would be good to explain the stability. Accessions are stable. For example, we expect to see ENSG00000139618 (BRCA2) across different Ensembl builds. NCBI accessions are like this, too. In contrast, [Ensembl MySQL][1] uses many internal integer primary keys (which is common in SQL design). I believe these keys are at times changed when Ensembl makes a new gene build. Would you call these primary keys as IDs (we do in SQL design)? I raised this because @dglazer thinks

> ids are durable for the lifetime of the server, and persistent across server restarts -- once a user is given an ID for data stored on a server, the id remains valid for as long as the server is still storing that data

ID by this definition is stable. Then those internal SQL primary keys are not IDs. Are ids stable or not?

[1]: http://useast.ensembl.org/info/docs/api/variation/variation-database-schema.pdf

diekhans:
Hi @lh3,  The term `id' gets used and abused to represent uniqueness in different scopes.
While I was quite surprised that `id' was server-local, the analogy to foreign
keys is a good way to think about it.

The term `id' is often used for primary key columns, for instance Ensembl's
convention is the primary key is table_id.  So the primary key column for
homo_sapiens_core_81_38.transcript is transcript_id (see below)

While intuitive names are important, rigorously defining their
meaning is required for an unambiguous API.

+--------------------------+------------------------------------------------------------------------------------------+------+-----+---------------------+----------------+
| Field                    | Type                                                                                     | Null | Key | Default             | Extra          |
+--------------------------+------------------------------------------------------------------------------------------+------+-----+---------------------+----------------+
| transcript_id            | int(10) unsigned                                                                         | NO   | PRI | NULL                | auto_increment |
| gene_id                  | int(10) unsigned                                                                         | YES  | MUL | NULL                |                |
| analysis_id              | smallint(5) unsigned                                                                     | NO   | MUL | NULL                |                |
| seq_region_id            | int(10) unsigned                                                                         | NO   | MUL | NULL                |                |
| seq_region_start         | int(10) unsigned                                                                         | NO   |     | NULL                |                |
| seq_region_end           | int(10) unsigned                                                                         | NO   |     | NULL                |                |
| seq_region_strand        | tinyint(2)                                                                               | NO   |     | NULL                |                |
| display_xref_id          | int(10) unsigned                                                                         | YES  | MUL | NULL                |                |
| source                   | varchar(40)                                                                              | NO   |     | ensembl             |                |
| biotype                  | varchar(40)                                                                              | NO   |     | NULL                |                |
| status                   | enum('KNOWN','NOVEL','PUTATIVE','PREDICTED','KNOWN_BY_PROJECTION','UNKNOWN','ANNOTATED') | YES  |     | NULL                |                |
| description              | text                                                                                     | YES  |     | NULL                |                |
| is_current               | tinyint(1)                                                                               | NO   |     | 1                   |                |
| canonical_translation_id | int(10) unsigned                                                                         | YES  | UNI | NULL                |                |
| stable_id                | varchar(128)                                                                             | YES  | MUL | NULL                |                |
| version                  | smallint(5) unsigned                                                                     | NO   |     | 1                   |                |
| created_date             | datetime                                                                                 | NO   |     | 0000-00-00 00:00:00 |                |
| modified_date            | datetime                                                                                 | NO   |     | 0000-00-00 00:00:00 |                |
+--------------------------+------------------------------------------------------------------------------------------+------+-----+---------------------+----------------+

ENSG00000139618.10

mbaudis:
I've removed the GUID attribute and reformatted fro Sphinx.

mbaudis:
I am closing this now since documentation is going other places & GUID etc. still not being tackled.

PR #424, opened 2015-09-24
documentation_stub

mbaudis:
@diekhans Does this seem like a start, or what would be your expectation?


diekhans:
Part of this seems like more of an introduction to the goals.  Also needs to be more specific.

mbaudis:
Sure & definitely. More of a test if this is where things should go ...

PR #427, opened 2015-09-25
replace CallSet id and name with label

richarddurbin:
This replaces the id (server-owned) and name (user-owned) for a CallSet by a label (user-defined, subject to uniqueness and syntax constraints).  This allows user-meaningful labels for the equivalent of VCF columns, without semantic problems of clashes.

Also it moves the info map for a call set to a new CallMethod object, with a methodId in the call set, and a default one in the variant set.
This allows sharing of meta-information about the calling process and other aspects of call sets.

Other changes are that I made call-sets only belong to one variant set (why could they belong to many?) and removed the updated timestamp because currently we have no way to update.

The new concept is Label, which is a string of printable characters, unique within the scope of the variant set, possibly with restricted syntax e.g. [a-zA-Z][a-zA-Z0-9_]\* like programming language identifiers.  The aim is that this should be a reasonable thing for users to look at and use, e.g. to display as column headings, or use in queries.
This replaces both columnId, which is server-unique and not under user control, and name, which is user-defined and can’t be relied on to be unique.
We are OK to require the labels to be unique, because the scope is limited to the VariantSet, so long range clashes are not a problem.  The semantics of merging VariantSets can be specified as part of the merge operation - both options of maintaining uniqueness by adding suffixes if necessary or forcing merge when the labels coincide have merit in different circumstances.
Because the labels must be unique, they can be used as a key for retrieval of CallSet columns in the VariantSet.
Having them separate from the sampleId means that you can have two CallSet columns for the same sample in one VariantSet, which is nice to be able to compare call sets made in different ways from the same sample (e.g. Illumina versus CG, or different callers).
Having a methodId that references an external Method object, which contains an info map for metadata, allows columns to share the same information about how they were called.  We also add an optional methodId to the VariantSet itself, which would then act as a default for all its callsets, covering the standard case where all callsets are made the same way.

It is good for both sampleId and methodId to be optional, so that lightweight VariantSets can be made by importing VCF files or the equivalent, without having to create lots of other empty objects first.  We should not underestimate the importance of lightweight use of the object representation and API - a large amount of sequence data handling is done in small labs with LIMS that will autopopulate Sample objects, or in exploratory analysis, and we would like users not to have to manage unnecessary appendages.

I need to add changes to variantMethods to this pull request, but don't know how to just now in this web interface.

### DIFF
--- a/src/main/resources/avro/variants.avdl
+++ b/src/main/resources/avro/variants.avdl
@@ -65,6 +65,11 @@ record VariantSet {
   the VCF header information not already presented in first class fields.
   */
   array<VariantSetMetadata> metadata = [];
+  
+  /**
+  The default CallMethod id for CallSets in this VariantSet.
+  */
+  union { null, string } methodId = null;
 }
 
 /**
@@ -73,31 +78,23 @@ It belongs to a `VariantSet`. This is equivalent to one column in VCF.
 */
 record CallSet {
 
-  /** The call set ID. */
-  string id;
-
-  /** The call set name. */
-  union { null, string } name = null;
+  /** 
+  The label for this CallSet.  User-defined, unique within its VariantSet.  
+  Required to satisfy identifier syntax [A-Za-z][A-Za-z0-9_]* 
+  */
+  string label;
 
   /** The sample this call set's data was generated from. */
-  union { null, string } sampleId;
+  union { null, string } sampleId = null;
+  
+  /** An id for a CallMethod with information about the calling process. */
+  union { null, string } methodId = null;
 
-  /** The IDs of the variant sets this call set has calls in. */
-  array<string> variantSetIds = [];
+  /** The ID of the variant set this call set has calls in. */
+  string variantSetId;
 
-  /** The date this call set was created in milliseconds from the epoch. */
+  /** The time this call set was created in milliseconds from the epoch. */
   union { null, long } created = null;
-
-  /**
-  The time at which this call set was last updated in
-  milliseconds from the epoch.
-  */
-  union { null, long } updated = null;
-
-  /**
-  A map of additional call set information.
-  */
-  map<array<string>> info = {};
 }
 
 /**
@@ -111,23 +108,13 @@ the occurrence of a SNP named rs1234 in a call set with the name NA12345.
 record Call {
 
   /**
-  The name of the call set this variant call belongs to.
+  The label of the call set this variant call belongs to.
   If this field is not present, the ordering of the call sets from a
   `SearchCallSetsRequest` over this `VariantSet` is guaranteed to match
   the ordering of the calls on this `Variant`.
   The number of results will also be the same.
   */
-  union { null, string } callSetName = null;
-
-  /**
-  The ID of the call set this variant call belongs to.
-
-  If this field is not present, the ordering of the call sets from a
-  `SearchCallSetsRequest` over this `VariantSet` is guaranteed to match
-  the ordering of the calls on this `Variant`.
-  The number of results will also be the same.
-  */
-  union { null, string} callSetId = null;
+  union { null, string } callSetLabel = null;
 
   /**
   The genotype of this variant call.
@@ -215,7 +202,6 @@ record Variant {
   */
   long end;
 
-
   /**
   The reference bases for this variant. They start at the given start position.
   */
@@ -240,4 +226,11 @@ record Variant {
   array<Call> calls = [];
 }
 
+record CallMethod {
+  /** The id for this call method */
+  string id;
+  
+  /** A map of information about the calling method */
+  map<array<string>> info = {};
+}
 }


lh3:
1. At present, Both CallSet and ReadGroup have a `sampleId` field. Some say this is the user submitted name at @RG-SM in SAM and at the sample line in VCF; others say this is an ID. I don't know what sampleId is. It is not well defined either way. My suggestion is to get rid of of this sampleId completely before we have proper BioSample.

2. Sample label is not a VCF-level concept. It is a project-level concept. It glues BAM and VCF in practice. In addition, we also need a place to hold sample information (e.g. population and gender in 1000g) independent of VCF. Such information is also applied to the whole project, not only to VCF. We need a project-level sample object. That is AnalysisSample in #412 and #402. AnalysisSample is not biological sample. It is your sample label moved above both VariantSet and ReadGroupSet.

3. You seem to suggest calling query-able name as label. Another option, as @jjfarrell suggested, is to call such a name as a tag.

diekhans:
@mbaudis has a basic BioSample ready to go.  We can incorporate this
into any pull request.

mlin:
Thanks @richarddurbin @dglazer et al.! Adding to "label" and "tag", I'd tried "handle"- I'm indifferent to these terms. A couple simplifications I'd like to suggest based on the prior discussions:

1. Adding `array<string> VariantSet::columnLabels` (tentative field name). This gets the metadata objects out of the way when just fetching calls, unless the application specifically needs them. In light of this simplification I find it reasonable to assume that the array will not become problematically large in the service lifetime of our VCF-based data model.
2. In light of the availability of `VariantSet::method`, the remaining case served by sharing CallMethod between columns by ID seems really obscure to me. I would put `union { null, CallMethod } VariantSet::method` and similarly on CallSet.
3. Delete CallSet and `VariantSet::updated`. Just kidding!

@lh3 re your point 2 - I don't think this proposal intends for "label" to reference an analysis sample or other object, for example when Richard discusses possibly adding suffixes to the label during a merge operation, and "you can have two CallSet columns for the same sample in one VariantSet" while the label is locally unique. Those suggest label is just some arbitrary string the user gets to control, but technically the application still has to dig into the metadata to cross-reference the analysis sample. (My columnHandle also avoided claiming to designate a sample although I'd assumed it would become that.)

richarddurbin:
sampleId is clearly not the name in the SAM file or VCF file.  As an id it has to be able to be server generated and not under used control.

In my view we should adopt the BioSample object Mark Diekhans and Michael Baudis refer to, and link to it with sampleId, both from ReadGroup and CallSet.  I recognise that there can be different ReadGroup libraries from the same individual person and different callsets on them, but there can also be different callsets from the same readgroup.  BioSample allows (I presume) definition of distinct DNA preps from the same individual as different samples - that is certainly in the EBI/NCBI BioSample which has proven utility (it does really work) and was I think the basis for the GA4GH object design.  Certainly all the gender and population information we had in AnalysisSample as a holding object should be in BioSample.

Whether the "name" in ReadGroup should be the same as the label in CallSet is up for discussion.  I guess it often will be, but probably we should not force this because you can have two call sets derived from the same ReadGroup.  You might ask whether "name" in ReadGroup should be a "label".  The problem with that from my point of view is whether the scope can be adequately defined to ensure that the user can reasonably guarantee uniqueness within scope.

That said, in most cases of course the CallSet label will be the BioSample.name, as we typically use in VCF objects.  This is all fine, but there are certainly good variantsets where one does not want that, and also naive "bare" variantsets coming from other sources without associated reads etc.

I am happy in principle to use some other word than "label" for the concept - I just wanted to get away from the tug of war about "id" (which I think should not be a string but an opaque type ID) and "name".  For me "tag" is slightly less good because I use that word to denote the key for a key:value pair in a map, but maybe I am idiosyncratic about that.

mbaudis:
@richarddurbin The PR for a skeleton ```BioSample``` object is in https://github.com/ga4gh/schemas/pull/397. However, this does not yet include e.g. ```sex```, as this is one of the attributes defined through an ```OntologyTerm``` object, which so far is only defined in the metadata branch. No problem to extend the ```BioSample``` PR with it, though.

See  https://github.com/ga4gh/schemas/blob/metadata/src/main/resources/avro/metadata.avdl

jjfarrell:
Another merge situation that probably should be addressed is when
concatenating SNVs/small Indels from one Caller and structural variants
from another Caller. Each variant row has been joint genotyped but by
different programs optimized for a particular type or size of the variant.

In this situation, each joint genotyped row would need to reference one
MethodId (eg. HaplotypeCaller, GenomeSTRiP, Lumpy, Pindel, etc).​

On Thu, Sep 24, 2015 at 8:05 PM, Richard Durbin <notifications@github.com>
wrote:

> This replaces the id (server-owned) and name (user-owned) for a CallSet by
> a label (user-defined, subject to uniqueness and syntax constraints). This
> allows user-meaningful labels for the equivalent of VCF columns, without
> semantic problems of clashes.
>
> Also it moves the info map for a call set to a new CallMethod object, with
> a methodId in the call set, and a default one in the variant set.
> This allows sharing of meta-information about the calling process and
> other aspects of call sets.
>
> Other changes are that I made call-sets only belong to one variant set
> (why could they belong to many?) and removed the updated timestamp because
> currently we have no way to update.
>
> The new concept is Label, which is a string of printable characters,
> unique within the scope of the variant set, possibly with restricted syntax
> e.g. [a-zA-Z][a-zA-Z0-9_]* like programming language identifiers. The aim
> is that this should be a reasonable thing for users to look at and use,
> e.g. to display as column headings, or use in queries.
> This replaces both columnId, which is server-unique and not under user
> control, and name, which is user-defined and can’t be relied on to be
> unique.
> We are OK to require the labels to be unique, because the scope is limited
> to the VariantSet, so long range clashes are not a problem. The semantics
> of merging VariantSets can be specified as part of the merge operation -
> both options of maintaining uniqueness by adding suffixes if necessary or
> forcing merge when the labels coincide have merit in different
> circumstances.
> Because the labels must be unique, they can be used as a key for retrieval
> of CallSet columns in the VariantSet.
> Having them separate from the sampleId means that you can have two CallSet
> columns for the same sample in one VariantSet, which is nice to be able to
> compare call sets made in different ways from the same sample (e.g.
> Illumina versus CG, or different callers).
> Having a methodId that references an external Method object, which
> contains an info map for metadata, allows columns to share the same
> information about how they were called. We also add an optional methodId to
> the VariantSet itself, which would then act as a default for all its
> callsets, covering the standard case where all callsets are made the same
> way.
>
> It is good for both sampleId and methodId to be optional, so that
> lightweight VariantSets can be made by importing VCF files or the
> equivalent, without having to create lots of other empty objects first. We
> should not underestimate the importance of lightweight use of the object
> representation and API - a large amount of sequence data handling is done
> in small labs with LIMS that will autopopulate Sample objects, or in
> exploratory analysis, and we would like users not to have to manage
> unnecessary appendages.
>
> I need to add changes to variantMethods to this pull request, but don't
> know how to just now in this web interface.
> ------------------------------
>
>   https://github.com/ga4gh/schemas/pull/427
> Commit Summary
>
>    - replace CallSet id and name with label
>
> File Changes
>
>    - *M* src/main/resources/avro/variants.avdl
>    <https://github.com/ga4gh/schemas/pull/427/files#diff-0> (59)
>
> Patch Links:
>
>    - https://github.com/ga4gh/schemas/pull/427.patch
>    - https://github.com/ga4gh/schemas/pull/427.diff
>
> —
> <https://github.com/ga4gh/schemas/pull/427>.
>



-- 
John Farrell, Ph.D.
IT Manager
Biomedical Genetics-Evans 218
72 East Concord Street
Boston, MA

ph: 617-638-5491

lh3:
Another terminology for "label" is "symbol". In the compiler theory, a symbol usually refers to the name of a variable or a function. Within a certain scope (e.g. between a closed curly brackets in C), the same symbol identifies the same variable. LISP/Schema has a concept of symbol, too, which plays the same role.

In the context of GA4GH, we may have both "name" and "symbol". We put no restrictions on naming clash, but require symbols to be query-able and unique within a certain scope, such that objects can be identified by their symbols. Currently we have two types of symbols: sample symbols and reference sequence symbols.

pcingola:
I agree that using "symbol/label" (unique within the scope) and 
"name" (no restriction) could solve some of the confusion. 

  Saying "chromosomeSymbol / chromosomeLabel" instead 
of "chromosomeName" could be puzzling at first glance, but I'm 
willing to live with that ("geneSymbol" instead of "geneName" is 
probably OK for most people).

lh3:
@pcingola thanks for mentioning gene name/symbol. HGNC (the official gene naming committee) also uses "symbol" explicitly. For the [TP53](http://www.genenames.org/cgi-bin/gene_symbol_report?hgnc_id=HGNC:11998) gene, its official symbol is "TP53" while its name is "tumor protein p53".

richarddurbin:
Thanks John.  Yes, this clearly also happens that we may want to attach a MethodId to a Variant.
I am now not sure whether this creates combinatorial effects at the single call level that we should worry about.  I would rather not.

Richard

> On 25 Sep 2015, at 20:22, jjfarrell <notifications@github.com> wrote:
> 
> Another merge situation that probably should be addressed is when
> concatenating SNVs/small Indels from one Caller and structural variants
> from another Caller. Each variant row has been joint genotyped but by
> different programs optimized for a particular type or size of the variant.
> 
> In this situation, each joint genotyped row would need to reference one
> MethodId (eg. HaplotypeCaller, GenomeSTRiP, Lumpy, Pindel, etc).​
> 



-- 

PR #432, opened 2015-10-02
genotype to phenotype

bwalsh:
Re-add the g2p schemas in.  This PR is paired with https://github.com/ga4gh/server/pull/915


calbach:
Hi bwalsh@, I noticed that `sequenceAnnotations.avdl` is included in this PR. Is this part of the proposal, or is this more of a placeholder to indicate that this PR depends on a full set of sequence annotations PRs being merged first?

diekhans:
sequenceAnnotations is also been developed in the RNA topic branch.  Please coordinate
with Sean Upchurch <sau@caltech.edu> and myself.

src/main/resources/avro/genotypephenotype.avdl
```diff
@@ -0,0 +1,176 @@
...
+Disease Ontology (DO): http://purl.obolibrary.org/obo/DOID_4
+Sequence Ontology (SO): http://www.sequenceontology.org/browser/
+Evidence Code Ontology (ECO): http://www.ontobee.org/browser/index.php?o=ECO
+Phenotypic Qualities (PATO): http://www.ontobee.org/browser/index.php?o=PATO
+*/
+protocol GenotypePhenotype {
+
+import idl "common.avdl";
+import idl "metadata.avdl";
+import idl "sequenceAnnotations.avdl";
```
bwalsh: Dependency between G2P & sequenceAnnotations

lh3:
For someone unfamiliar with g2p like me, could you show a query example using g2p? Specifically, what are the user inputs (e.g. "c.117_118delTG on BRCA2"?) and what are the server outputs (e.g. "breast cancer"?)? If so, could you write down the exact user query and server response in JSON? If not, could you show an example what g2p is designed for? The schema is mostly talking about various ontologies. I am a bit lost. Thanks.

bwalsh:
@lh3 

Using curl:

```
curl --data '{"feature": {"ids": [{"identifier": "CGD:d8c2d551UniProtKB:P10721#P10721-1Region", "version": "*", "database": "http://www.monarchinitiative.org/_"}]}}' --header 'Content-Type: application/json' http://127.0.0.1:8000/genotypephenotype/search
```

Using the cli:
```
ga4gh_client genotypephenotype-search --feature '{"ids": [{"identifier": "CGD:d8c2d551UniProtKB:P10721#P10721-1Region", "version": "*", "database": "http://www.monarchinitiative.org/_"}]}’ --outputFormat ‘json’ http://127.0.0.1:8000/
```


Response
```
{
  "features": [
    {
      "featureType": {
        "ontologySource": "http://ohsu.edu/cgd/",
        "id": "27d2169c",
        "name": "http://ohsu.edu/cgd/27d2169c"
      },
      "end": null,
      "parentIds": [],
      "start": null,
      "featureSetId": "",
      "referenceName": null,
      "attributes": {
        "vals": {}
      },
      "id": "http://ohsu.edu/cgd/4657f28c",
      "strand": null
    }
  ],
  "evidence": [
    {
      "evidenceType": {
        "ontologySource": "http://www.ncbi.nlm.nih.gov/pubmed/",
        "id": "18955458",
        "name": ""
      },
      "description": null
    }
  ],
  "environmentalContexts": [],
  "phenotype": {
    "id": null,
    "type": {
      "ontologySource": "http://ohsu.edu/cgd/",
      "id": "1b6442aa",
      "name": "GIST with decreased sensitivity to therapy"
    },
    "ageOfOnset": null,
    "qualifier": null,
    "description": null
  },
  "id": "http://ohsu.edu/cgd/4657f28c",
  "description": null
}

```

lh3:
Thanks. But where could a user get "CGD:d8c2d551UniProtKB:P10721#P10721-1Region" and "http://www.monarchinitiative.org" in the first place? In addition, ohsu links such as "thttp://ohsu.edu/cgd/27d2169c" are not available. Is registration required to access this database?

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,159 @@
...
+
+
+/**
+The feature collection to search for.  One or more features (variants,
+genes, etc) can be specified.  The idea here is that if a
+query is for a gene, then any alleles to that gene for which
+there are annotation records would be returned.
+*/
+record GenomicFeatureQuery {
+  array<org.ga4gh.models.Feature> features;
```
sarahhunt: This specifies an array of GA4GH Feature records is the required input, but the implementations I have seen and compliance tests appear to use Feature ids. Which is intended?
The same query applies to the other input types.

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,159 @@
...
+  Provide this value in a subsequent request to return the next page of
+  results. This field will be empty if there aren't any additional results.
+  */
+  union { null, string } nextPageToken = null;
+}
+
+/**
+Gets a list of evidence accessible through the API.
+
+`POST /genotypepheotype/search` must accept a JSON version of
```
sarahhunt: genotypepheotype => genotypephenotype

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,159 @@
...
+`SearchGenotypePhenotypeResponse` as the post body and will return a JSON version
+of `SearchGenotypePhenotypeRequest`.
```
sarahhunt: These are the wrong way round!

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,159 @@
...
+*/
+SearchGenotypePhenotypeResponse searchGenotypePhenotype(
+  /**
+  This request maps to the body of `POST /genotypepheotype/search` as JSON.
```
sarahhunt: genotypepheotype =>genotypephenotype

src/main/resources/avro/genotypephenotype.avdl
```diff
@@ -0,0 +1,176 @@
...
+    The set of features of the organism that bears the phenotype.
+    This could be as complete as a full complement of variants,
+    or as minimal as the confirmed variants that are known causation
+    for the annotated phenotype.  
+    Examples of features could be variations at the nucleotide level, 
+    large rearrangements at the chromosome level, or relevant epigenetic
+    markers.  Relevant genomic feature types are suggested to be 
+    those typed in the Sequence Ontology (SO).
+
+    The feature set can have only one item, and must not be null.
```
sarahhunt: This comment is not clear - can a feature set have only one type of item?

kshefchek:
In regards to [comment](https://github.com/ga4gh/schemas/pull/432#issuecomment-147425286)

        Thanks. But where could a user get "CGD:d8c2d551UniProtKB:P10721#P10721-1Region"
        and "http://www.monarchinitiative.org" in the first place? In addition,
        ohsu links such as "http://ohsu.edu/cgd/27d2169c" are not available.
        Is registration required to access this database?

This is being built from an rdf serialization of a database that is not web accessible, so in many cases we created IRIs as placeholders that do not resolve.  More information about the dataset can be found here: https://github.com/monarch-initiative/mckb

macieksmuga:
@bwalsh I _love_ the new documentation, including the use cases provided. Are the diagrams included in the document also checked into github/part of this PR? (if so, I missed that).

doc/source/schemas/genotypephenotype.rst
```diff
@@ -0,0 +1,172 @@
...
+3) As a genetic counselor, I may be wondering if a mutation in one of my
+   clients' genes has ever been associated with a disease. I could then
+   do a query based on the gene name as the feature and
+   `disease <http://purl.obolibrary.org/obo/DOID_4>`__ as the
+   phenotype.
+
+For specifics of the json representations, please see the
+`server <https://github.com/ga4gh/server>`__ and
+`compliance <https://github.com/ga4gh/compliance>`__ repositories.
+
```
diekhans: Very useful doc!  It would be good to tie it into the other parts of the API. So how do you go from variants or variant annotations to G2p.

bwalsh:
Thanks for the encouragement.

The diagrams are links to uploads to githubusercontent.com.   What would be
a better way to maintain the diagrams? (store locally in the  directory?)

-Brian Walsh

On Wed, Feb 3, 2016 at 10:08 AM, Maciek Smuga-Otto <notifications@github.com

macieksmuga:
One question: Sequence Annotations also has `Feature`, with the attendant `searchFeaturesRequest` method. Are we headed for a collision, or is it resolved via some namespace magic?

bwalsh:
@benedictpaten @calbach @diekhans
The server and compliance PRs are "ready to merge" - how can I help in getting this PR merged in?
Thank You,
-b

AAMargolin:
+1. Hopefully ready to merge.

rnpandya:
+1 let's move this forward!

shadowtramp:
+1
Good start!

jacquisb:
Well if you need one extra, here it is +1
Good luck

diekhans:
the source to diagrams should be checked into the docs directory.  This way, documentation can be maintained along with the source

diekhans:
This is not quite ready to merge.  We need to begin converging with where sequence annotations and metadata are at in the other branches to not have to tear G2P up radically when they come in.  Probably not a huge deal.

diekhans:
I don't understand how collections of G2P mappings are managed in G2P.

Reads and Variants have ReadSets and VariantSets.  These are members of datasets.  This schema
seems global to an entire server instance.

diekhans:
After meeting with @bwalsh  and @david4096, we realize that G2P doesn't provide any type of container object to group and manage sets of associates.  They are also not part of any dataset.  This is all addressed by the patch provided by @sarahhunt .  We believe it will be fast to implement this and it needs to be done now rather than merge into master something that is fundamentally incompatible with the rest of GA4GH and hence not releasable.

diekhans:
The semantics of  SearchGenotypePhenotypeRequest are very unclear.  I would really have no idea how to construct a query.  There is an ExternalIdentifierQuery for all records, but no ExternalIdentifier in genotypephenotype.avdl.  It might match something in Feature attributes, but it is unclear what attributes it would match against.  evidence can be queries by string and  ExternalIdentifier but only has a description and ontology term.  What does these match.

Maybe this just needs documentation, but it needs to be clear how to use it before it gets merged, ever if the queries are very minimal.

@AAMargolin, can the G2P TT make the query definition more rigorous so this can move forward?

diekhans:
Since there are no response from the G2P TT on the discussion above regarding query, I am
giving this pull request a -1.

The behavior of the queries are ill defined and not implementable as specified.  The engineers are needed to guess at the correct behavior.

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,159 @@
...
+
+  union {null, string, ExternalIdentifierQuery, OntologyTermQuery,
+         GenomicFeatureQuery} feature = null;
+
+  union {null, string, ExternalIdentifierQuery, OntologyTermQuery,
+         PhenotypeQuery} phenotype = null;
+
+  union {null, string, ExternalIdentifierQuery, OntologyTermQuery,
+         EvidenceQuery} evidence = null;
+
```
diekhans: Instead of having the same union repeated three times for the query arguments create a record for the union.  This will ultimate result in generated objects that will be useful in writing common code.

macieksmuga: Each of these unions is different. Are you proposing a "meta-union" over them, along the lines of:
```
          GenomicFeatureQuery, PhenotypeQuery, EvidenceQuery}
```
?

diekhans: Oh, I see, I didn't notice the second line differenced.

No, not a meta query.

never mind

doc/source/schemas/genotypephenotype.rst
reece: Yeah, what Mark said! User stories like this are gold. We should hold this up as an example for others. 

We might also consider how to pull these out of code and structure them, perhaps as a github wiki item or as an element of documentation. What if major code feature merges must include a user story in documentation?

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,159 @@
...
+BRCA2 would mean that in addition to any direct associations
+to the BRCA2, all associations to sequence variants of BRCA2 would also
+be returned.  Similarly, queries with OntologyTerms should perform
+the subclass closure.
+
+Each query can be made against a string, an array of external
+identifers (such as for gene or SNP ids), ontology term ids, or
+full feature/phenotype/evidence objects.
+*/
+record SearchGenotypePhenotypeRequest {
```
reece: If I understand this correctly, I think we should be concerned about clashing of unscoped identifiers. For example, I read this as supporting something like ``{ 'phenotype': 'FH' }``, in which I think it's unclear whether that's FH the gene (via "ExternalIdentifierQuery") or Familial Hypercholesterolemia (via "PhenotypeQuery"). Is that (or something like it) a valid concern here?

[Larger conversation: We (GA4GH) are using unions as a way to avoid important conversations regarding precision of meaning. This is directly analogous to strong v. weak typing tradeoffs in programming languages. Anyone else have similar concerns?]

diekhans: as usual, @reece says it far better than I can.

The subject of the queries is not defined.  What does string
match in one of these queries?  What happens when the objects
evolve?

The lack of precision in the APIs is a concern.  GA4GH took on a
very big challenge that requires rigorous definition: defining
an interoperable API.  It's much easier for Facebook to define
an API, because their is one implementation.

bwalsh:
closed (replaced with https://github.com/ga4gh/schemas/pull/619)

PR #436, opened 2015-10-08
Update intro_metadata.rst

mbaudis:
This PR adds time formats as part of the metadata documentation. Formats have been discussed previously and are (partially) used in the metadata branch already.

The documentation here serves as reference for related attribute definitions throughout the schema.

### DIFF
--- a/doc/source/intro_metadata.rst
+++ b/doc/source/intro_metadata.rst
@@ -6,11 +6,90 @@ Metadata API (stub)
 
 For the Metadata schema definitions, see the `Metadata schema <schemas/metadata.html>`_
 
-
 ------------------
 Metadata
 ------------------
 
-..todo::
-   Add documentation
+`FIXME: Stub here for time format submission; to be extended...`
 
+* standardized use of common attributes/values
+* a schema of how objects relate to each other
+
+
+-----------------------------------
+Date and Time Format Specifications
+-----------------------------------
+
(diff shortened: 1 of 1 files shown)


dglazer:
How much of this documentation applies to the current API in master, and how much is only relevant to proposed future enhancements?

For example, are there any durations or intervals in the current schema? (I'm not remembering any offhand, but I haven't looked exhaustively.)

mbaudis:
Not sure if there is any time related implementation in the current version => check?

However, we know that we will need both time points and intervals/durations, at least in metadata, but probably also for object housekeeping. For metadata, the need of the different types given through the examples is pretty clear & has been discussed based on our developing data test implementations.

I've made a note about 

```FIXME: Add examples for time attributes as when such are implemented in the schema.```

diekhans:
Given the slow progress on doc, I am for collecting as much
stuff as possible in this branch and then cherry picking a merge
into master.

dglazer:
Thanks Michael for confirming that this is mostly forward-looking.

Mark, your call on process -- it just means the merge into master will be
trickier.

diekhans:
Doc that seems solid and short term future looking I will take
in and deal with the merge.  Longer term, most speculative
stuff, such as GUID, I have ask to be delayed.

Will probably take suggestion of breaking in to multiple pull
requests.

Issue #441, opened 2015-10-20
Variant Annotations: AlleleLocation's overlapEnd coordinate should be optional


pcingola:
In the current schema (variant_annotations branch), AlleleLocation's overlapEnd seems 
to be a mandatory which is probably excessive. For instance, current ANN's format 
used for storing annotations in VCF files does not provide end coordinates for overlap 
within cDNA, CDS or Protein. So following this requirement would complicate server 
implementations that their back data in VCF files.

I propose to simply change the definition by making it an optional field, so the new 
record would be like this:

    record AlleleLocation {
      /** Relative start position of the allele in this coordinate system */
      int overlapStart ;
    
      /** Optional relative end position of the allele in this coordinate system  */
      union { null, int } overlapEnd ;
    
      /** Reference sequence in feature (this should be the codon at CDS level) */
      union { null, string } referenceSequence = null;
    
      /** Alternate sequence in feature (this should be the codon at CDS level) */
      union { null, string } alternateSequence = null;
    }

AngieHinrichs:
It can be a pain to calculate missing end coordinates - but isn't the end coordinate rather important?  It seems like this would simply move the pain downstream, and would there be enough information to recalculate it?

Are you concerned that the end coordinates would be discarded in VCF?  Why not keep them in the INFO column?

pcingola:
In today's call @sarahhunt mentioned that the field was indeed intended as optional (at 
least that's what I understood), so we wouldn't be changing the schema so much, just 
formalizing the intention.

The reason relates to the ANN field used for annotations (here is a link to the current 
ANN standard: http://snpeff.sourceforge.net/VCFannotationformat_v1.0.pdf). Using 
this ANN field, we don't have all the data required to create cDNA/CDS/protein AlleleLocation 
records, the best we can do is to make the whole entry 'null'.

I agree that the best solutions would be to: a) update the ANN standard; and/or b) not implement 
servers based on VCF files. But in the mean time, we need a schema that allows us to 
implement servers and adding non-standard INFO fields would be painful in the long run.

AngieHinrichs:
OK, I understand.  I didn't catch the reference to "ANN" in your initial comment.

sarahhunt:
[...]

Issue #439, opened 2015-10-20
Variant annotations: AnnotationSet should point to VariantSet instead of DataSet


pcingola:
This issue is for the variant annotations branch.

Currently the structure makes it difficult to retrieve the variantSet since we must go all the way up to the DataSet and match (possibly more than one) the VariantSet:

    record AnnotationSet {
      /** The ID of the annotation set record */
      string id;
      /** The ID of the dataset to which this annotation set belongs */
      string datasetId;

A simple change to a variantSet handle instead of a dataSet handle would improve the situation:
 
    record AnnotationSet {
      /** The ID of the annotation set record */
      string id;
      /** The ID of the variantSet to which this annotation set belongs */
      string variantSetId;

calbach:
I definitely disagree with **changing** it, though a variant set ID field could certainly be added to the annotation set provenance. It would not make sense in general for an annotation set to belong to a variant set.

pcingola:
Could you elaborate on why "it would not make sense"?
Everyone in today's call seemed to agree that it would be best to change it.

calbach:
I assumed this was referring to the [annotations branch](https://github.com/ga4gh/schemas/tree/annotations) which models arbitrary sequence annotations, e.g. Genes (which clearly won't necessarily 'belong to' a variant set). Could you link to a specific line to clarify?

pcingola:
Actually this refers to the "variant annotations" branch.
Apologies for the confusion, I should have clarified it in the title.

pcingola:
Changed the issue's title, I hope it is less confusing now.

diekhans:
in general, `AnnotationSet' is a confusing name.  I would encourage naming this VariantAnnotationSet.  Annotation is a very general term that means different things in different genomics contexts.

pcingola:
I'm OK with changing the name, but this should probably 
be discussed as a separate issue...

sarahhunt:
Hi Pablo,

I've had a look at this, and querying for annotation across many variant 
sets was the original intention.

The proposed change would make the use case of 'I'm interested in this 
gene - what variants are in the region?' more tedious for the client. 
You would need to issue multiple queries for each 1KG, ClinVar, ESP, 
ExAC, etc VariantSet and filter through them all though the results from 
each query would overlap heavily.

That said, the change gives the user a clearer idea of provenance/data 
quality and fits better with the concept of a dataset as a container for 
reads which can be submitted to different calling proceedures to make 
different variant sets.

[...]

PR #443, opened 2015-10-22
VariantAnnotationSet change & Allele location update

sarahhunt:
Updates following VariantAnnotation call on 20/10/2015 and further discussion in issues #439 and #441. 

Graph Allele support is also removed awaiting stability in the underlying Allele records.


pcingola:
+1

PR #449, opened 2015-11-03
Draft release notes for version 0.6.

hjellinek:
Candidate release notes for version 0.6.  Comments welcome.

### DIFF
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -1,0 +1,97 @@
+# Changes to `ga4gh/schemas` `master` branch since version `v0.5.1` (Oct 2, 2014)
+
+### Pervasive changes
+
+Renamed all protocols and records to remove the "`GA`" prefix from
+their names.  (Hence `GASearchVariantSetsRequest` is now
+`SearchVariantSetsRequest`, etc.)  The exception is `GAException`,
+which is unchanged.
+
+There is no longer one unitary namespace.  Objects now reside in
+"`org.ga4gh.models`", methods in "`org.ga4gh.methods`".
+
+Updated the version number to `0.6.0-SNAPSHOT`.
+
+### Added
+
+Datasets are now used throughout the API.
+
+#### Methods:
+
+* `getDataset`
+* `searchDatasets`
+* `getReadGroupSet`
+* `getReadGroup`
+* `getVariant`
+* `getVariantSet`
+* `getCallSet`
+
+#### Records:
+
+* `ExternalIdentifier`
+* `Experiment`
+* `Dataset`
+* `ReadStats`
+* `Fragment` _(apparently unused)_
+
+#### Enum:
+
+* `Strand`
+
+### Removed
+
+* `src/main/resources/avro/beacon.avdl`
+* Files designated "Work In Progress" ("`wip`"):
+    * `src/main/resources/avro/wip/metadata.avdl`
+    * `src/main/resources/avro/wip/metadatamethods.avdl`
+    * `src/main/resources/avro/wip/variationReference.avdl`
+
+### Changed
+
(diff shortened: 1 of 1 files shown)


benedictpaten:
This looks useful to me, 1+. In defining a formal release process we should ensure we keep adding to this.

doc/release-notes.md
```diff
@@ -0,0 +1,97 @@
...
+* `getReadGroup`
+* `getVariant`
+* `getVariantSet`
+* `getCallSet`
+
+#### Records:
+
+* `ExternalIdentifier`
+* `Experiment`
+* `Dataset`
```
dglazer: Dataset already existed in 0.5.1  (it was defined in reads.avdl)

doc/release-notes.md
```diff
@@ -0,0 +1,97 @@
...
+* `getReadGroupSet`
+* `getReadGroup`
+* `getVariant`
+* `getVariantSet`
+* `getCallSet`
+
+#### Records:
+
+* `ExternalIdentifier`
+* `Experiment`
```
dglazer: Experiment already existed in 0.5.1

doc/release-notes.md
```diff
@@ -0,0 +1,97 @@
+# Changes to `ga4gh/schemas` `master` branch since version `v0.5.1` (Oct 2, 2014)
+
+### Pervasive changes
+
+Renamed all protocols and records to remove the "`GA`" prefix from
```
jeromekelleher: These are really Avro changes, since they don't actually affect the wire format. I think it would be helpful to break the changes into two sections, one for the wire format, that client side developers can use to see what they have to change, and another for the Avro biased summary that you have here.

doc/release-notes.md
```diff
@@ -0,0 +1,97 @@
...
+* `getVariantSet`
+* `getCallSet`
+
+#### Records:
+
+* `ExternalIdentifier`
+* `Experiment`
+* `Dataset`
+* `ReadStats`
+* `Fragment` _(apparently unused)_
```
jeromekelleher: The `Fragment` should really be removed before we make the release. It currently doesn't do anything, so we should just remove it from the schema entirely. It'll be easy enough to add back in later when we figure out what we need it for.

jeromekelleher:
This is a good start @hjellinek and captures the high-level Avro based stuff well. What we're missing though is an overview of what's happened to the wire-format and how this has changed. There's important stuff like the change in the semantics of the ``callSetIds`` field in the ``SearchVariantsRequest``, the change from ``dataSetIds`` to ``dataSetId`` etc through lots of the requests. These are the things we need to capture: what does a client side developer *need to know* if they are changing from v0.5.1 to v0.6.0?

adamnovak:
I think we should fix up the couple of errors pointed out above about things having existed before, merge this in, and then elaborate on changes that aren't mentioned here in additional PRs on top.

+1

doc/release-notes.md
calbach: Sent https://github.com/ga4gh/schemas/pull/499 to address this.

jeromekelleher:
What's the status of this PR? There's still quite a bit of work to be done, but I think @hjellinek has moved on to greener pastures. Is someone else taking it up?

macieksmuga:
+1 on Adam's suggestion. Since I can't access Herb's fork of this repo, I'll close this PR and open up a new one with the above minor suggestions implemented. Then we can walk through the changes that Jerome is suggesting.

dglazer:
+1 to these changes; they clearly make it better.

I still like @jeromekelleher 's suggestion from earlier: _I think it would be helpful to break the changes into two sections, one for the wire format, that client side developers can use to see what they have to change, and another for the Avro biased summary that you have here._

PR #453, opened 2015-11-06
Setup includes for glossary

macieksmuga:
Moved short_reads section into glossary includes, also included in reads API description.


diekhans:
I think this is too large for a glossary entry, however we can edit it later, so I will merge.

PR #458, opened 2015-11-06
Revise Avro and JSON documentation

adamnovak:
In GA4GH world, we use only the AVDL part of Avro, and we serialize to
JSON in our own special simple way. This previously was not documented.
Now it is.

I've also put some information (to be expanded) on how we define an API
by defining a bunch of objects.

### DIFF
--- a/doc/source/appendix/avro_intro.rst
+++ b/doc/source/appendix/avro_intro.rst
@@ -4,27 +4,69 @@
 Apache Avro
 *******************
 
-Copied from the `Avro documentation <http://avro.apache.org/docs/1.7.7/>`_
+Apache Avro is a data serialization ecosystem, comparable to Google's Protocol Buffers.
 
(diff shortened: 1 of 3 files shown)


diekhans:
This is not exactly accurate.   The reference server optional validates using Avro IDL components.  The complainant suite use Avro IDL components to parse the JSON.  We don't use the Avro wire protocol and we misuse the serialization by going in at a lower level.

So it's not accurate to say we only use the avdl files;

adamnovak:
If we're using the Avro library code to read and write the JSON, are we convincing it to not read and write Avro-style JSON with the tagged unions? Or are we actually using tagged-union JSON and I'm just wrong?

adamnovak:
OK I've updated this to say that we also use the JSON libraries, and to document our self-imposed restrictions on union types.

PR #463, opened 2015-11-06
Fixed broken image and anchor.

mbookman:


diekhans:
should this be ../_static/ rather than /_static/ ???  or is the /_static translated by sphinx to a relative URL??

mbookman:
This source:

    ./doc/source/api/variants.rst:.. image:: /_static/variant_objects.svg

Turns into:

    ./target/generated-docs/merged/html/api/variants.html:

I'm not clear where in the build the path transformation is occurring.

If "../_static" works, then this becomes really a question of convention. The downside to relative paths is that if the .rst file is moved (as it was in this case), then the image include may break. No strong opinion, here.

PR #468, opened 2015-11-07
Updated documentation: reads.rst edits

macieksmuga:


diekhans:
the source for the image needs to be included

PR #472, opened 2015-11-07
Improve specificity of the Reads API documentation.

calbach:
Add missing docs, increase specificity in several cases, and make some minor style tweaks.


calbach:
I think Mark thought this was a PR onto the `documentation` branch. In actuality this was against `master`. Oops...

How do we want to proceed? I can send a PR to revert this, and then resend the original PR. Mark, if you agree, please merge my upcoming revert PR.

PR #474, opened 2015-11-07
Revert 'Read API documentation' which was accidentally merged too quickly

calbach:
Reverts ga4gh/schemas#472


calbach:
Thanks, I'll revert the revert in the next few days.

PR #476, opened 2015-11-07
Streamline doc build process

reece:
The previous doc build process rebuilt all avpr and rst for every invocation, regardless of whether the source avdl files were updated.

This PR implements 1) a make-based incremental build, 2) a top-level 'make docs' target to coordinate the process. Rebuilding docs is much faster. See doc/README.rst for details.

This will also make it possible to easier to build rst from avdl files and commit them to the repo. Doing so will enable read the docs builds.

generate_sphinx_docs.sh wraps new method for backward compatibility. Note that docs now end up in build/html (which is more standard for python sphinx builds).


dglazer:
@reece -- this is great, thanks for getting it in. Are you also planning to submit a PR against master to merge the tools in there, or is there more work needed first?

PR #482, opened 2015-11-09
closes #481: create initial doc structure in master

### LINKED ISSUE #481: Create initial documentation structure in master
During the 2015-11-06 docathon at Google, significant progress was made to create a documentation structure and tooling.  Although the documentation is not complete (i.e., not ready for release), the group believed in desirable [1] to port the doc skeleton  and build tools into master. That is the goal of this issue.

When this issue is completed, we can undertake automating the build process at either Read the Docs or on a public server elsewhere. (The choice will depend on whether the avdl-to-rst process can be cajoled into working on RTD.)

[1] https://docs.google.com/document/d/1MEEkv3V37StEV7ps8jtVnzCQxxIRz101timpOf-ZI6s/edit?ts=563d4f39

reece:
This PR adds the skeleton outline from the docathon to the master doc tree. It also streamlines the doc build process ('make docs' in the schemas root dir). Read the Docs automatically builds docs on commit. See ga4gh-schemas.readthedocs.org/. 

Note that this PR includes only the top-level index.rst and intro.rst. Content from the docathon may be merged when it is finalized (perhaps selectively, by top-level doc area).

After this PR is merged:
* Someone with github admin rights needs to update the RTD webhook. 
* I will selectively merge relevant changes from master into documentation.

@dglazer: Please review the above website to ensure that this fits with your expectation about changes to master and comment on this issue with feedback.


doc/README.rst
```diff
@@ -0,0 +1,55 @@
...
+
+.. note:: The ``tools/sphinx/generate_sphinx_docs.sh`` script is
+          obsolete.  The functionality is entirely subsumed by ``make
+          docs``, which also enables incremental updates.
+
+A `daily build
+<http://hgwdev.cse.ucsc.edu/~markd/ga4gh/documentation-pr/>`_ is also
+available at UCSC. 
+
+Changes are in progress to enable building the documentation at Read
```
dglazer: Checking -- do you want to leave this as "in progress" until all the i's are dotted for auto-building?

dglazer:
+1 -- this is a clear improvement over the (almost non-existent) doc infrastructure in master, and will make it much easier to incrementally merge in changes from the documentation branch.

A few small comments -- completely up to you which ones you want to address now, vs, in subsequent PRs:
* doc/source/intro.rst isn't baked yet. I have a slight preference for just submitting a placeholder for now (like you do with the other sections), but I'm also okay with leaving it as is, since it's not bad, and we can iterate
* the RTD website you link to (http://ga4gh-schemas.readthedocs.org/en/latest/api/index.html) doesn't seem to quite match this PR -- for example, I don't see a Schemas entry in the top-level ToC, and I don't see any sub-headings under the Appendix entry.
* related -- I expected that RTD website to not only have a Schemas entry, but to have it auto-populated based on the .avdl in master. That wouldn't be great doc yet, but it would be current.

reece:
@diekhans: Please hold on merge. I'll address @dglazer's comments with an updated PR.

reece:
Explanation of updated PR:

* Removed intro content. Can be merged anytime from documentation branch.

* In sphinx, headings come from the target doc, not the index. If the target doc is missing, sphinx warns and omits it from the toc. This is why some toc elements were missing. (Others were missing because I had inadvertently omitted them.)

* I added schema docs back as requested. NOTE: this means that the commit now includes the schema rst files, which in turn means that we are now on the hook for updating these files when the .avdls change. A note to this effect is now in doc/README.rst.

dglazer:
Thanks Reece -- most of these changes look like what I expected.

I'm confused about the schema docs, though, and might have sent us down a blind alley.
* a) the first version of this PR had no schema docs
* b) I pictured a version that would work like the documentation branch, with schema docs dynamically built from the .avdl (manually for now and automatically down the road). There wouldn't be any checked-in .rst files corresponding to the .avdl files, so doc/source/schema would be almost empty (as is https://github.com/ga4gh/schemas/tree/documentation/doc/source/schemas)
* c) the current version of this PR has checked-in copies of snapshotted .rst files, presumably built from the current .avdl files in master

If (b) is doable, I still think it's best, since it puts us on a path to gradually improve the docs as we merge in improved comments on the .avdl. But if not, then I'd prefer (a), which is a subset of where we want to be, not (c), which creates a bunch of files we'll shortly want to remove.

diekhans:
I just created this PR against the documentation branch that removes the
pandoc dependency:
       https://github.com/ga4gh/schemas/pull/483

I haven't merged it yet.  Dose it make sense to merge this into master 
with #481 as well?

Mark

dglazer:
That depends - does it work with the existing .avdl files in master?

diekhans:
dropping pandoc works with existing .avdl, although there might some
formatting ugliness.

Although I though the goal was to get the tools merged, not all
the documentation working.

reece:
On Mon, Nov 9, 2015 at 5:31 AM, David Glazer <notifications@github.com>
wrote:

> If (b) is doable, I still think it's best, since it puts us on a path to
> gradually improve the docs as we merge in improved comments on the .avdl.
> But if not, then I'd prefer (a), which is a subset of where we want to be,
> not (c), which creates a bunch of files we'll shortly want to remove.


This explanation may help:
https://github.com/reece/schemas/blob/master/doc/README.rst.

RTD is simple: they run their installation of sphinx-build against a repo's
doc tree with the repo's conf.py. That's it. We don't get any other
official entry points. This implies that all files necessary for
documentation must be present in the commit.

I have confirmed that java is not available on RTD, which means that
on-the-fly conversion is not possible.


I have one strong and one weak opinion about our path forward.

The strong opinion is that RTD is a big enough win that we should make it
work and that there are acceptable options to substantially mitigate the
avdl-rst consistency issues (see below). If we don't make RTD work, we will
end up owning the entire doc build process, and that would be unfortunate.

If we agree that RTD is our target automated doc platform, then my weak
opinion is that we should proceed with the PR as-is. In the spirit of
constant, imperfect incremental progress, I'd rather have some schema
content with blemishes rather than none. (Actually, I have this feeling
about the rest of the pages too.)

I'm first in line for disliking checking in intermediates. This as a
stop-gap measure that moves the ball forward toward automated builds; it's
not a long term plan.

Here are 3 options to wean us off of the committing the intermediate doc
files:

   1. Best option: UCSC's contractor is planning to build a sphinx domain
   for protobuf. They could almost certainly arrange to build docs directly
   from avdls, akin to the sphinx autodoc module that transparently builds
   docs from .py files. This would be an elegant solution to the entire issue.
   2. Another option is to write a test to check for avdl-rst consistency.
   That would treat avdl-rst consistency kinda like unit tests and code style
   compliance checks, and block code merges if they fail.
   3. Mark floated the idea of building rsts on travis and pushing to a new
   doc repo that would be picked up by RTD.

I personally don't like 3 for the same reason I think we shouldn't own the
doc build machinery.

-Reece

diekhans:
I completely agree with moving ahead and refining.  Also, RTD is
a great place to publish and the reference server is there.

Running a doc build as a temporary solution doesn't really worry
me all that much.  However, RTD is motivating for a common place
to find documentation (for Python people).

To eliminate Java in the Avro sphinx domain would require
writing a parser for avdl files.  I am loathed to spend
significant abouts of money on something that will be thrown
away quickly.  I don't think it's a three-hour job.  Plus it
will slow down the work on protocol buffers, since they are
resource limited.

I am only weakly against checking in the intermediates.  It's
more cruft and will show up in the code review process.  I guess
if they were automatically build every time a rst or avdl is changed,
they will end up in people's pushes and then not get stale.

My concern about travis is just that it might be annoying to get
it working, but it would work fully withing github/RTD and be
triggered by the original commit, so it seems different that
running out own build process.

By all means, charge ahead.   Any doc is an improvement.

reece:
On Mon, Nov 9, 2015 at 10:49 AM, Mark Diekhans <notifications@github.com>
wrote:

> To eliminate Java in the Avro sphinx domain would require
> writing a parser for avdl files. I am loathed to spend
> significant abouts of money on something that will be thrown
> away quickly. I don't think it's a three-hour job. Plus it
> will slow down the work on protocol buffers, since they are
> resource limited.
>

Completely agree. In my "best option" suggestion re: the contractor, I was
referring to protobuf but inadvertently referred to avdls (I intended
.proto).

I also would not spend another moment even thinking about avdls if we're
committed to conversion (which I think we are).

I was trying to suggest that you (Mark) tack on to the existing contracting
engagement to build docs from *protobuf* definitions. This should be doable
and is probably even easy (but I don't know how!).

-Reece

diekhans:
Yes, protobuf domain will not be dependent on non-python code and it's
already part of the contract.

Mark

dglazer:
All sounds good -- we know where we're heading (with auto-built RTD docs
from .rst and .proto), and we have a plan for getting there (with the
contractor wiring up .proto -> RTD).

I'll defer to Reece and Mark on which intermediate steps move us there best
-- if checking in the generated .rst files for now is better than the
alternatives, fine with me.

reece:
I'd prefer committing the rsts, but not strongly. Mark, are you cool with keeping the rsts in now so that the docs are not completely barren? We can remove them when proto-to-rst is in place or earlier if you take on the travis build, whichever happens first.

diekhans:
Just bouncing ideas; I am totally cool with it!

it's certainly the approach that is the least work.

Another approach would be to keep the parse schema in json format
if that make it easier.

reece:
On Mon, Nov 9, 2015 at 1:11 PM, Mark Diekhans <notifications@github.com>
wrote:

> Another approach would be to keep the parse schema in json format
> if that make it easier.
>

I don't think so. If we're going to store an intermediate, it might as well
be a fully baked intermediate if you know what I mean.

reece:
@diekhans -- Planning on merging?

diekhans:
+1 this is great

Jeltje:
+1, progress, hurray!

PR #484, opened 2015-11-09
(Unreverting) Improve specificity of the Reads API documentation.

calbach:
Add missing docs, increase specificity in several cases, and make some minor style tweaks.

Trying this again after merging too-soon on the original attempt, and subsequently reverting that PR. Reverts ga4gh/schemas#474

### DIFF
--- a/src/main/resources/avro/readmethods.avdl
+++ b/src/main/resources/avro/readmethods.avdl
@@ -11,11 +11,11 @@ This request maps to the body of `POST /reads/search` as JSON.
 
 If a reference is specified, all queried `ReadGroup`s must be aligned
 to `ReferenceSet`s containing that same `Reference`. If no reference is
-specified, all `ReadGroup`s must be aligned to the same `ReferenceSet`.
+specified, all queried `ReadGroup`s must be aligned to the same `ReferenceSet`.
 */
 record SearchReadsRequest {
   /**
-  The ReadGroups to search. At least one readGroupId must be specified.
+  The ReadGroups to search. At least one id must be specified.
   */
   array<string> readGroupIds;
 
@@ -72,7 +72,23 @@ record SearchReadsResponse {
 }
 
 /**
-Gets a list of `ReadAlignment` matching the search criteria.
+Gets a list of `ReadAlignment`s for one or more `ReadGroup`s.
+
+`searchReads` operates over a genomic coordinate space of reference sequence
+and position defined by the `Reference`s to which the requested `ReadGroup`s are
+aligned.
+
+If a target positional range is specified, search returns all reads whose
+alignment to the reference genome *overlap* the range. A query which specifies
+only read group IDs yields all reads in those read groups, including unmapped
+reads.
+
+All reads returned (including reads on subsequent pages) are ordered by genomic
+coordinate (by reference sequence, then position). Reads with equivalent genomic
+coordinates are returned in an unspecified order. This order must be consistent
+for a given repository, such that two queries for the same content (regardless
+of page size) yield reads in the same order across their respective streams of
+paginated responses.
 
 `POST /reads/search` must accept a JSON version of `SearchReadsRequest` as
 the post body and will return a JSON version of `SearchReadsResponse`.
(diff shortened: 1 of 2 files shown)


dglazer:
+1 -- only affects comments, and all the edits are improvements over where we started.

(Full disclosure -- CH and I worked on these files together during the docathon, so I'm not impartial.)

src/main/resources/avro/readmethods.avdl
```diff
@@ -72,7 +72,20 @@ record SearchReadsResponse {
...
+aligned.
+
+If a target positional range is specified, search returns all reads whose
+alignment to the reference genome *overlap* the range. A query which specifies
+only read group IDs yields all reads in those read groups, including unmapped
+reads.
+
+All reads returned (including reads on subsequent pages) are ordered by genomic
+coordinate (by reference sequence, then position). Reads with equivalent genomic
+coordinates are returned in an unspecified but deterministic order.
```
jeromekelleher: Why "deterministic" order here? This doesn't really add anything, and is ambiguous. I think "... are returned in no specific order" would be better. I think it's better to put it this way as it discourages clients from relying on server-specific orders, which is bad for interop.

src/main/resources/avro/reads.avdl
```diff
@@ -114,6 +128,11 @@ record ReadGroup {
   map<array<string>> info = {};
 }
 
+/**
+A ReadGroupSet is a logical collection of ReadGroups. Typically one ReadGroupSet
+represents all the reads from one experimental
```
jeromekelleher: Join line?

jeromekelleher:
This is certainly an improvement. I'm +1 modulo the niggle about ordering above.

src/main/resources/avro/reads.avdl
calbach: Done.

src/main/resources/avro/readmethods.avdl
calbach: What I was trying to express here was: if I send the same request twice to the same repository over data which hasn't changed, I expect to get the same ordering. I don't want clients to depend on one server's specific ordering, but rather give them the option of depending the property that there is **a** self-consistent ordering for one repository. I also wasn't really happy with this wording though - suggestions?

For example, say I'm creating a UI back by `SearchReads` which displays paginated results with `pageSize=10`, and the first matches are 100 different reads which are all mapped to the same position. If I refresh the page, I would like the first page to contain the same content every time - rather than a random permutation of 10 reads from those 100.

jeromekelleher: I see, this is a good point. How about we split this into two sentences then? "..are returned in an unspecified order. This order is required to be consistent with a server implementation, however, such that the list of results obtained from the same query is always the same". Or something along those lines?

diekhans: Please include the rationale for ordering as paging in the
doc.  Prevents asking the same questions about why.

calbach: I've revised this comment. It was a bit of struggle to find a middle-ground between providing rationale, keeping semantics weak enough to allow flexibility of implementation, and readability. I started to add more detailed rationale, but it felt very out of place. This is developer documentation - I don't think it's appropriate to rationalize all design decisions here unless that information also happens to be useful in writing client code or implementing a server. If someone really cares in this case, they can look at the blame and find my response above.

Hopefully this is also more self-explanatory with the way I've rephrased it. Please take a look.

jeromekelleher: I like this version, and I agree that rationalising would be out of place here. We should have a higher level point somewhere that ordering of result streams (regardless of page sizes) are expected to be stable throughout the API, and can attach the rationale behind this requirement there.

jeromekelleher:
+1

calbach:
Any more comments? Is this good to merge?

macieksmuga:
+1 Reads well. I believe it's good to merge too.

jeromekelleher:
We have the requisite +1-age and elapsed time, so I'm merging. Thanks @calbach, this is a good clarification.

Issue #343, opened 2015-06-16
How much work would be really needed to shift away from AVRO to a proposed protobuf based workflow? (from #287, item 10)


Relequestual:
(Referencing https://github.com/ga4gh/schemas/issues/287#issuecomment-111578618 )

It was discussed that the effort required to move from Avro to Protocol Buffers should be investigated. Someone at the meeting volenteered to do this (apologies, I forget who you were). Please claim this issue! @diekhans can you remember who it was?

diekhans:
Jerome and CH are working on in.

jeromekelleher:
Yes, I'm working on a quick proof of concept to see how much work would be involved in porting the reference server and client to protobuf. I'll report back here when there's some news.

Relequestual:
@jeromekelleher Is there any update on this issue?

diekhans:
It is stalled waiting for ProtoBuf 3 JSON support

Relequestual:
The ProtoBuf 3 protocol, or the libraries that implement it?

diekhans:
Library to implement it for python

Relequestual:
@diekhans Is there any news or ETA on the protobuf 3 python library? If not, is it worth considering Swagger? There's already good support for Swagger2 in most languages, including python. http://swagger.io/open-source-integrations/ 

I am happy to help with converting AVRO to Swagger2, but I'm afraid I don't know python, so wouldn't be able to help with writing the test implementation.

jeromekelleher:
We're still waiting on proto3 support in Python @Relequestual --- no concrete updates for now, but we'll have something more definite by the NY meeting.

Swagger looks interesting @Relequestual --- could you outline what you would imagine the protocol definitions and other tooling would look like using this?

Relequestual:
I hear @buske might be at the NY meeting, so hopefully he can feed back! =]

Swagger is indeed interesting. The Swagger editor allows you to easily see many examples of the spec in use, and the corrsponding documentaion that's generated (http://editor.swagger.io). Click file, open example.
The protocol definition can be split over multiple files, as it is currently. These can be referenced in a master file, which would collate all files to generate the documentation, or for usage within other applications. being json/yaml based, it supports JSON Reference like so http://azimi.me/2015/07/16/split-swagger-into-smaller-files.html

There are number of tools provided by Swagger directly.
Swagger Core consists of Java libraries for generating and reading definitions.
Swagger Codegen is a CLI tool for generating client and server side code from definitions.
[...]

PR #485, opened 2015-11-10
Convert avdl -> proto3

calbach:
As alluded to in several GA4GH discussions and [issues](https://github.com/ga4gh/schemas/issues/343), this PR converts the GA4GH schemas from Avro to [protocol buffers](https://developers.google.com/protocol-buffers/docs/overview). This is ready to be discussed and reviewed, but a prerequisite for merging will be an agreed plan for corresponding changes to the reference server and compliance suite (in the works, but need an explicit plan for sequencing the PRs).

I'll highlight a few factors which motivate this proposal:
- Protocol buffers have defined JSON serialization, which was sorely missing in our use of Avro. Libraries for protobuf <-> JSON conversion are defined at http://github.com/google/protobuf. See also the [specification](https://developers.google.com/protocol-buffers/docs/proto3#json).
- Gives the GA4GH a straightforward pathway to support an optimized binary wire format in the future.
- (Slightly subjective) Generally more community adoption / tooling / client library support.

At a high level, my aim with this PR was to make the minimal possible functional / documentation / wire JSON / code structure change while still winding up with valid proto3. I have a list of natural follow-up changes given proto as the schema - but I think these make more sense as separate PRs.

That being said, there are some changes from Avro -> proto3 which are unavoidable:
- proto3 does not have optional/default semantics, e.g. Avro's `union {string, null} foo = ""`
  - In places, this may change the JSON wire format slightly - e.g. null -> "" or vice versa.
  - As far as I can tell, the only place this really causes a problem is for `SearchVariantsRequest.callsetIds`, which will be resolved separately by https://github.com/ga4gh/schemas/pull/488.
  - Technically you can still get "nullable" semantics by using a wrapper message, at the cost of ugliness and stutter in the API.
  - If you're using the generated proto libraries, this shouldn't make a difference.
- I converted all 'methods' into proto RPC service methods.
  - For now, this would be for documentation purposes only, much as the Avro methods are.
  - In order to support the GET operations, I've added small ‘request’ protos (with a single ID) for those endpoints, in addition to a referenceId on the ListBasesRequest.
- All fields now have [tag numbers](https://developers.google.com/protocol-buffers/docs/proto3#assigning-tags).
- In proto3, enum 0-values should be a sane 'unset' default.
  - The reason behind this is that the 0-value is the default when you initialize the request proto, or if a client sends you nothing. In general, it is better to disambiguate this case.
  - For this reason, added `${ENUM_NAME}_UNSPECIFIED` to all enums. This makes sense in most places, though feels the least natural in Cigar Operation.
- map\<[]string\> becomes map\<google.protobuf.ListValue\>.
  - This is the best way of generating consistent JSON with the previous approaches, but the ListValue message is actually more flexible than []string (elements could be number, message, etc).
  - For now, the requirement of a `[]string` payload is enforced only by the comment.
  - An alternative would be to relax this documentation to allow all values of `ListValue` or to use the more freeform google.protobuf.Struct, which represents arbitrary JSON.
  - I don't have a strong opinion here, again this first pass is what I believe to be the minimal functional delta.
- JSON wire format change: "sourceURI" -> "sourceUri"
- In a few places, nested message definitions to make the hierarchy clearer e.g. CigarUnit { Operation {} operation }. This mainly aesthetic, though it also may impact the generated libraries.

Also, for reference, the Python generated classes can be created by the following process:
1. Install protoc >3.0 from https://github.com/google/protobuf/releases
2. `$ cd src/main/resources && protoc -I . --python_out=. proto/{,*/}*/*.proto`


diekhans:
Given our current guidelines, the best approach is to create a
topic branch and pull this into the topic branch.  We will then
have other topic branch for the server and compliance suites.

dglazer:
@calbach -- thanks for the legwork here, and for enumerating the short (but non-zero) list of wire format changes. Looks good to me, assuming no surprises as we update the other affected code.

I strongly agree that we _"need an explicit plan for sequencing the PRs"_ across the schemas/reference/compliance repos, and the documentation branch. I'd support any plan that the owners of the respective conversions (you, Jerome, Herb, Mark) are comfortable with; let us know if/when you want input.

jeromekelleher:
This is great, thanks @calbach! I haven't gone through this in detail. 

The wire format changes look entirely reasonable to me. The only contentious issue I can see is the ``callSetIds`` field in ``SearchVariantsRequest``. I think by far the simplest and best option here is to drop the "null==everything" semantics. There is no need for a shortcut to request all calls.

calbach:
I will split out a PR for removing null vs empty list behavior on `SearchVariants` to minimize noise here.

Also still TODO for this PR (but shouldn't block review of the content):
- Update pom.xml to get mvn working for protobuf
- Ensure CI covers protobuf compilation

src/main/resources/proto/ga4gh/reads.proto
```diff
@@ -0,0 +1,246 @@
...
+  // `alignedSequence` and `alignedQuality` may be shorter than the full read
+  // sequence and quality. This will occur if the alignment is part of a
+  // chimeric alignment, or if the read was trimmed. When this occurs, the CIGAR
+  // for this read will begin/end with a hard clip operator that will indicate
+  // the length of the excised sequence.
+  repeated int32 aligned_quality = 15;
+
+  // The mapping of the primary alignment of the `(readNumber+1)%numberReads`
+  // read in the fragment. It replaces mate position and mate strand in SAM.
+  Position nextMatePosition = 16;
```
IanRogers-LShift: snake_case for consistency?

src/main/resources/proto/ga4gh/reference_service.proto
```diff
@@ -0,0 +1,168 @@
...
+  // The continuation token, which is used to page through large result sets.
+  // To get the next page of results, set this parameter to the value of
+  // `next_page_token` from the previous response.
+  string page_token = 5;
+}
+
+// This is the response from `POST /referencesets/search` expressed as JSON.
+message SearchReferenceSetsResponse {
+  // The list of matching reference sets.
+  repeated ReferenceSet referenceSets = 1;
```
IanRogers-LShift: snake_case for consistency?

src/main/resources/proto/ga4gh/reference_service.proto
```diff
@@ -0,0 +1,168 @@
...
+message GetReferenceSetRequest {
+  // The ID of the `ReferenceSet` to be retrieved.
+  string reference_set_id = 1;
+}
+
+// ****************  /references  *******************
+// This request maps to the body of `POST /references/search` as JSON.
+message SearchReferencesRequest {
+  // The `ReferenceSet` to search.
+  string referenceSetId = 1;
```
IanRogers-LShift: snake_case for consistency?

calbach:
Thanks @IanRogers-LShift , not sure how I missed those. I've made the corrections.

I took another pass through all the files, so I think there shouldn't be any more cases of this.

src/main/resources/proto/ga4gh/metadata.proto
```diff
@@ -0,0 +1,108 @@
+syntax = "proto3";
+
+package ga4gh;
+
+import "proto/google/protobuf/struct.proto";
```
jeromekelleher: We've hit a minor issue on the server concerning the layout of the protos @calbach: ga4gh/server#876. Basically, we want to re-root to get rid of this ``proto`` prefix here. Is this possible?

Also, as an aside, would it be controversial to drop the ``main/resources`` prefix as well? Seems pretty pointless having such a deep directory structure when there's only one thing in it.

calbach: If it works I don't have any real objection to either. IIRC, we put proto there to avoid conflicts with the package "google" which had already been imported into the reference server.

I don't know the history around main/resources. Removal of those intermediate folders seems reasonable at a glance, though I don't know why they exist in this form currently.

jeromekelleher:
``GAException`` seems to have disappeared here @calbach --- is there a reason for removing this?

src/main/resources/proto/ga4gh/variant_service.proto
```diff
@@ -0,0 +1,175 @@
...
+
+// This is the response from `POST /variantsets/search` expressed as JSON.
+message SearchVariantSetsResponse {
+  // The list of matching variant sets.
+  repeated VariantSet variant_sets = 1;
+
+  // The continuation token, which is used to page through large result sets.
+  // To get the next page of results, set this parameter to the value of
+  // `next_page_token` from the previous response.
+  string page_token = 2;
```
jeromekelleher: Should be next_page_token

jeromekelleher:
Looks pretty close to me @calbach. The PR in the server (ga4gh/server#799) is nearly read to merge, so I think we're almost there. The only serious problem I've hit is the incompatible JSON in the ``info`` fields ( https://github.com/ga4gh/server/issues/879). For example in reads, now we have
```json
{"info": {"MD": ["63"] }
```
Under the current protobuf implementation they look like
```json
{"info": {"MD": {"values": [{"stringValue": "63"}]}}
```
which is definitely a breaking change.

Is this a necessary change? If there is any way we can keep the ``info`` fields wire-compatible I think we should do it, even if the proto definitions are (temporarily) ugly.

calbach:
RE: `GAException`. Yes, I failed to include this in the change log. Could you expand on how the reference server was using this, if it was? I've always seen this as an artifact of using Avro, as I don't believe it ever had a formal specification for how it would translate to HTTP. I believe that the GA4GH API error handling specification has just been 'HTTP statuses'. That said, there is a [proto equivalent](https://github.com/google/googleapis/blob/master/google/rpc/status.proto) that we can either pull into here, or possibly reference via dependency.

RE: `info`, this is definitely not expected. I will comment on the issue.

jeromekelleher:
For GAException, we're using this to return JSON to the user on error conditions, as well as the correct HTTP status. Does the Google API not do this?

calbach:
Ah, I understand now. Yes the Google API returns this same JSON, though I hadn't realized that this was part of the GA4GH spec (I don't see it documented anywhere, but perhaps I missed something).

diekhans:
The GAException method is a standard way to encode detailed
error information in the HTTP body.  This is in addition to the
HTTP response code.

src/main/resources/proto/ga4gh/variant_service.proto
calbach: Fixed.

calbach:
OK, so issues on my plate for this PR currently:

- Switch up the import paths and verify things still work, per Jerome's request.
- Get [rpc.Status](https://github.com/google/googleapis/blob/master/google/rpc/status.proto) into this PR, or reference these common protobufs somehow (I'll figure out the best way to go about this). This would replace `GAException`.
- Per comments on ga4gh/server#879, the info tags issue is in progress to be fixed and I will provide an ETA once I have one.

Lastly, I will likely delete this PR and make a new one, originating from the `protobuf` schemas branch (this PR is still originating from my personal fork). Upon doing so, I will summarize any pending comments from here in the new PR.

calbach:
I've created #547 which pulls from the protobuf topic branch instead.

PR #488, opened 2015-11-10
Drop "null callsets array = all calls" semantics for SearchVariants

calbach:
The issue is being forced by the [proposal to switch to proto3](https://github.com/ga4gh/schemas/pull/485); proto3 does not [disambiguate](https://developers.google.com/protocol-buffers/docs/proto3#default) between the cases of `null` and `[]`.
## Updated notes (2/11/16)

We're now simply removing the ability to retrieve all variant calls via the `callSetIds` parameter.
## Original notes

This has the side-effect that it is no longer possible to retrieve **only** the variants (sans calls). My preference is to eventually add support back in via cross-cutting response field masking for the API, e.g. via the proto3 [FieldMask](https://github.com/google/protobuf/blob/master/src/google/protobuf/field_mask.proto#L41) specification.

Alternatives considered:
1. Add another boolean field, e.g. `mask_calls` which could be passed to get the same behavior. Again, I think it makes more sense to avoid one offs and just have cross-cutting API support for field masking, but it could be acceptable in the interim, if the community feels strongly that we should not lose this functionality for an unknown duration.
2. Drop 'return all variant calls' semantics. I feel this would be surprising default behavior, since if I compose a minimal query, I expect to get almost everything back. In medium scale analysis even, e.g. thousands of callsets, having to pass all of these callset IDs will be cumbersome.

### DIFF
--- a/src/main/resources/avro/variantmethods.avdl
+++ b/src/main/resources/avro/variantmethods.avdl
@@ -71,10 +71,9 @@ record SearchVariantsRequest {
 
   /**
   Only return variant calls which belong to call sets with these IDs.
-  If an empty array, returns variants without any call objects.
-  If null, returns all variant calls.
+  If unspecified, return all variants and no variant calls objects.
   */
-  union { null, array<string> } callSetIds = null;
+  array<string> callSetIds = [];
 
   /** Required. Only return variants on this reference. */
   string referenceName;


jeromekelleher:
-1.

Why do we need to have this short-cut at all? From having used this API quite a lot, I *strongly* prefer the sane semantics of having [] correspond to no calls, ["csId1"] correspond to the callset with ID "csId1", etc. It is a semantic mess to have the empty list correspond to all calls, and also leads to ambiguity --- what order should the calls be returned in?

calbach:
Thanks Jerome. I fully expected the -1 since this proposal is counter to what you suggested on previous threads. I'd like to crystallize the arguments and get more input.

> also leads to ambiguity --- what order should the calls be returned in?

Actually, unless I missed something the documentation doesn't specify the behavior you describe even if you've provided call set IDs. The only statement I can find about relative call ordering is this [caveat for servers which don't populate callsetId](https://github.com/ga4gh/schemas/pull/488/files#diff-8fa5b209218e5652cabcaf1f631bc402L115) (which I hadn't actually read before). We could make the documentation stricter on this point, but does it actually matter anyways to a client? Are there other ambiguities this would introduce?

> a semantic mess

Could you elaborate? In my mind, I would explain this endpoint as follows: you're searching for variant calls within a `VariantSet`. `callSetIds` can also be provided to further filter the results. My proposed semantics for `[]` I think follow naturally.

diekhans:
Generally, +1, on doing away with this magic behavior, -1 on
this approach.

This functionality was added to address a problem UCSC's had.
The API was simply too slow for an brca exchange browser due the
the volume of calls returned.  The brca exchange current goes
directly to VCFs because of this.

Another approach is to change the data structure.  having the
calls contain in variants seems less that an ideal
representation.


Mark

calbach:
@diekhans I think you are mainly arguing that we retain the ability to retrieve only the variants. Does this mean that you would be satisfied with both alternatives (1) and (2) that I listed? I'm going to consider fundamental changes to the `Variant` record as out of scope for this PR, though I agree this is the likely root cause of the API's awkwardness in this case.

mbaudis:
@diekhans
> having the calls contain in variants seems less that an ideal representation

+n on this. When starting to look at structural changes, there are "no" recurring variants; searches are addressed at some feature (breakpoint, DUP | DEL ...) occurring in your query interval. So searches are against calls, or in VCF terms against variants-with-dermined-call-positivity.

Therefore, the best scenario for query targets would be to have per sample "variant calls".

Sorry for not being up to speed with everything going on in this area, but I recently realised that the current variants.avdl certainly isn't optimal even for e.g. simple copy number changes.

diekhans:
I agree with Jerome, that we should have a flag to return all
calls and an empty array or null return no calls.  This seems
like cleaner semantics.  Although, we also need to look at
consistency across the API (don't know the queries well
enough to comment).  Is it a filter or request?

The field mask is interesting, however would want to understand
what this means for the overall semantics of the API.  Does this
cause cases were objects with omitted fields are passed to code
that misinterprets the omitted fields and no data?  How does
impact caching of objects?

Agreed that simplifying the object structure is a separate task,
but something that needs looked at.

calbach:
> Although, we also need to look at
> consistency across the API (don't know the queries well
> enough to comment).  Is it a filter or request?

I did a quick scan and I don't see anything equivalent. Again, this is probably a special case due to the nested structure of Variant/Call.

> The field mask is interesting, however would want to understand
> what this means for the overall semantics of the API.  Does this
> cause cases were objects with omitted fields are passed to code
> that misinterprets the omitted fields and no data?  How does
> impact caching of objects?

I think there is minimal impact to the API semantics, if supported it would be an optional parameter to use on any request - but by default the behavior would be the same as today. The concerns you raised are the concerns of a client who has decided to utilize/support the field masking feature. But yes, for clients using field masking, they should take extra care to ensure that:

- they are using the feature correctly
- they are not passing the full record to another library which expects additional fields to be populated
- if they are also caching records, the field mask is accounted for in the caching semantics

diekhans:
Thanks for confining my suspicion that there isn't other cases
like this in the API.

It would be best if field masking needs to be handled
separately.  I would suggest going ahead with a short term flag
in the request that says all variants.  Feel free to start a
ticket to discuss masking, but it needs some careful
consideration.

I will save concerns and approaches to addressing them for
for the field mask PR.

diekhans:
BTW, this needs to be implemented in the reference server and compliance suite before merging.

jeromekelleher:
I'll say it again --- why do we need a mechanism to return all calls? The client can make a ``SearchCallSets`` request to find the IDs, and can then make a ``SearchVariants`` with these callSetIds as a parameter. There is no need for a second mechanism to do this.  We **do** need a mechanism to get variants with no calls, and there is already a sane, semantically clean way to do this in the API. Therefore, Occam's Razor says we drop the null==everything semantics, and move on.

I usually agree with pretty much everything you say @calbach, so I'm genuinely confused about where you're coming from here. To answer your questions:

> Actually, unless I missed something the documentation doesn't specify the behavior you describe even if you've provided call set IDs. The only statement I can find about relative call ordering is this caveat for servers which don't populate callsetId (which I hadn't actually read before). We could make the documentation stricter on this point, but does it actually matter anyways to a client? Are there other ambiguities this would introduce?

Well, the documentation is wrong on this point then. The ``callSetIds`` argument is a list. The returned ``calls`` field is a list. Surely the expected behaviour is that there is a one-to-one mapping between the ``callSetIds`` and the ``calls`` lists: they should have the same length, and they should have the same order. Given list of ``callSetIds``,  I can't imagine any server doing anything other than returning the ``calls`` list in the same order: it just doesn't make sense to do anything else. 

This matters to clients. There are efficiency gains to be had from knowing the order in which the calls are returned. If I'm generating some statistics for the calls I'm interested in,  I can allocate some arrays and know that the index corresponding to callSetId ``x`` will always be the same. Otherwise, I have to maintain a dictionary, indexed by callSetId, and I have to look this up every time. I can't mask out ID field from the ``Call`` object, because it's always necessary.

Unless we require that calls are returned in the order of the specified list of callSetIds, there is a lot of ambiguity and unnecessary hassle for the client. Is the ordering of calls stable within a server implementation, or can it vary from page-to-page, variant-to-variant? This would be a source of nasty and annoying client bugs, as well as portability issues (clients will write code that depends on the ordering of a given server).

This is part of what I mean by a semantic mess. If we depart from the obvious bijection between the input and output lists, there are all sorts of twisty loopholes that open up. We gain nothing from opening up these loopholes.

By far the worst semantically is equating the empty list with the list of all calls. This makes no sense, and is mathematically horrible. The empty list is a perfectly well defined object, with perfectly well defined semantics, which also happen to correspond exactly with something that we *need*. We *need* to be able to return variants with no calls. We have no need for a mechanism to return all calls in some arbitrary order.

You make the point that you would expect all the calls to be returned by default. I would expect exactly the opposite. Unless I specifically ask for the calls, I don't want them. I don't want a flood of thousands (or hundreds of thousands) of calls coming through, eating bandwidth and slowing parsing down to a crawl. The variant is a summary of the information that we know, and it's a reasonable thing to just return this summary unless more specific information about particular calls is requested.

diekhans:
excellent points @jeromekelleher

One concern about this and the field mask, is the semantic
meaning of the GA4GH objects. Are the objects shopping bags or
are they part of a data model?  This is something I have never
understood about GA4GH.

I am worried about the shopping bag approach without any kind of
indication what as put in the bag.  Does an Variant object with
no calls indicate there were no calls in the VCF or that they
were not requested?  I have never programmed a data model with
this kind of ambiguity.  It seems error prone.

benedictpaten:
On Wed, Nov 11, 2015 at 4:46 AM, Jerome Kelleher <notifications@github.com>
wrote:

> I'll say it again --- why do we need a mechanism to return all calls? The
> client can make a SearchCallSets request to find the IDs, and can then
> make a SearchVariants with these callSetIds as a parameter. There is no
> need for a second mechanism to do this. We *do* need a mechanism to get
> variants with no calls, and there is already a sane, semantically clean way
> to do this in the API. Therefore, Occam's Razor says we drop the
> null==everything semantics, and move on.
>
Strongly +1 this comment. Get rid of the all calls option. Passing in the
list of CallSets to the method is no more expensive than the return value
when all callSets are specified.

haussler:
This seems important. When you request info from a record, there must be a
distinction between "that field was null" and "you didn't ask for the value
of that field". This seems to be tied up in scoping the query and the
problem of how to return only part of a coherent data structure. I guess
this means a distinction between two types of missing values: "this part is
missing in the actual data structure" and "this part is missing because you
didn't ask for it so we didn't want to waste bandwidth sending it to you".

calbach:
OK, I've gone a bit deeper on this issue and I've discussed with my team. @jeromekelleher given your premises I think I agree with your conclusion. However I think there may be some more fundamental disagreement about those premises, and possibly there is another dimension to `SearchVariants` which hasn't been discussed yet - I'd like to resolve these points before continuing.

1) A `Variant` need not have a `Call` for **all** `CallSets` in the `VariantSet`.

Note that these semantics are looser than those of a server where 1 VariantSet == 1 VCF file, since the VCF spec requires 1 `Call` / sample for all variants (I'm including 'no call' here, which is still explicitly modeled by a `Call` resource). The current Variants API is ambiguous on this point, but my assertion is supported by the [variants diagram](http://ga4gh.org/assets/images/documentation/variants.png). These semantics allow for a more compact representation of a variant set, but I also think that the requirement of a full JOIN of `Variant`/`CallSet` within a `VariantSet` may be a technically unreasonable invariant for a server to enforce (especially if a `CallSet` is associated with more than one `VariantSet` as is allowed today - though I don't agree with that and maybe that's slated for removal).

What are the semantics of having no corresponding `Call` resource for a `CallSet` on a given `Variant`? We can be explicit in documenting this, but I think it naturally follows the same semantics as not representing a `Variant` in your `VariantSet` at all.

We should clarify the spec on this point and improve the corresponding documentation one way or the other.

2) If you accept my assertion in (1), we now need to clarify whether `SearchVariants()` is performing an OUTER or INNER JOIN of `Variant`s and `CallSet`s.

More concretely, if I request `{callSetIds: ["callsetA"]}`, do I get back:
a. All `Variant`s, even those with 0 matching calls (OUTER JOIN).
b. Only `Variant`s that have at least one matching call (INNER JOIN).

If it's (a), then the semantics you propose follow naturally. If it's (b), then I find my proposal to be more natural. Google's API currently implements (b) - I feel that (b) is the desired behavior in this case, as it allows for a terse response which includes only the relevant `Variant`s. If I'm searching by `CallSet`, I care about the `Call`s - the `Variant` is just side-data used to interpret these `Call`s.

>  When you request info from a record, there must be a
distinction between "that field was null" and "you didn't ask for the value
of that field".

Yes, this is a better articulation of my concerns around returning 0 calls by default. This is also generally a problem with the SearchVariants endpoint whenever you are filtering by callSetId. The root of the issue is that a `Variant` is not really a standalone resource as the API currently defines it. Rather there are `Variant`s and `Call`s which have been fused into a single representation for the sake of convenience. I think we will have to live with this caveat unless we are willing to change that representation.

jeromekelleher:
This is a great analysis @calbach, and has brought up a number of things I hadn't realised. My position would be, make the semantics OUTER JOIN, return a NO CALL for every missing call that is requested, and keep the request semantics I've argued for above. We absolutely need to clarify the points you've raised though, and I think we need some wider input here.

The `convenience' of merging variants and calls is looking more questionable all the time...

calbach:
Who else should we be getting input from on this? @lh3 maybe?

Perhaps @sarahhunt can comment on what EBI does here with their implementation of the variants API (though I'm guessing this is also closely tied to VCF)?

The 'NO CALL' is not a bad solution (though not my first preference). We would have to make this clear in the documentation since I think this generally indicates a lack of sufficient quality data to make a call - compared to the semantics of not including a variant at all, which generally means that you just didn't see any non-reference calls in your study (unless this is a gVCF). Returning a reference call for all missing calls is probably overstepping and could easily lead to errors.

diekhans:
IMHO, containing the Call objects in Variant is not
the best representation due to it's unbounded size.

A query to return the calls as a separate step would be give the
ability to filter what could be a very long list and result.

sarahhunt:
@calbach - you are correct, our implementation is VCF based and we return 'no call's for missing data. If the call was 'attempted' but there was not enough information to make it, it makes sense to return this information.
My preference would be to move calls out of Variant and add a calls search with a default 'return all in set' mode and optional filtering on callset.

msackman:
I've now hit this problem with the compliance suite: `VariantsGetByIT.checkVariantsGetResultsMatchSearchResults` is failing because of this problem.

The protobuf `variant_service.proto` includes the following under `SearchVariantsRequest`:

    // Only return variant calls which belong to call sets with these IDs.
    // If an empty array, returns variants without any call objects.
    // If null, returns all variant calls.
    repeated string call_set_ids = 2;

AFAIUI, that comment is impossible. The default value for a repeated field is the empty list. There is no distinction between empty list and null possible. If you wish to permit such a distinction (I've read through the above discussion briefly but I'm not sure there was a conclusion), one way of expressing this in protobuf would be:

    message CallSetIds {
      repeated string call_set_ids = 1;
    }
    oneof call_set {
      bool all = 2;
      CallSetIds ids = 3;
    }

That bool in there is completely pointless, but sadly protobuf doesn't support a void type that I can find (capnproto does, for example).

Fixing up the server is possible: `ga4gh/datamodel/variants.py` `getVariants` just needs the first branch conditional changing to `if not callSetIds:` but that breaks the server's own tests as they assume there is a distinction between `None` and the empty list.

calbach:
I've reverted this PR to what Jerome originally suggests on the thread, i.e. `[]` == variants only. No one else on the thread seems particularly keen on the 'return all variant calls' use case, and this `oneof` proposal would unfortunately cause JSON wire format incompatibility:

`{ids: {callSetIds: ["1", "2"]}}` vs `{callSetIds: ["1", "2"]}`

I think lets minimize the diff right now to unblock the proto work, and we can re-evaluate adding the 'get all calls' functionality at a later juncture.

This PR also does not touch the issue of INNER vs OUTER JOINs or call set ordering. The semantics here will remain undefined for the time being, as it sounds like general opinion is that we should fundamentally change the API rather than update the semantics of what we have today.

calbach:
I've also updated the `protobuf` branch to reflect this.

jeromekelleher:
Great, thanks @calbach! 

+1

calbach:
The score for this PR is currently at 0. @diekhans if you flip your -1 -> +1, we should have consensus to merge.

jacmarjorie:
I'm developing a tangentially related API (with proto3), and have been working heavily with searchVariants() for quite some time now. I ran into the similar null, [], [id1, id2] issue for a different set of objects and was hoping to follow the pattern you were using and ran into this thread... With that disclaimer, I can provide a little bit of feedback here given my experience with the SearchVariantsRequest (apologies for duplicating conversation, but additional input shouldn't hurt):

My primary use case for searchVariants is `{callSetIds: null }`. Under the filter possibilities you propose, I will have to pass a list with all callSetIds to get the equivalent behavior, assuming I want to access the actual call information across a set of samples (which I frequently do over 8000 samples). Note that our group has assumed the callSetId filter is performing an INNER join, not an outer join. Also, CallSet ordering in the calls object has never been a concern. 

As you have already noted a bit above, the issue with CallSetIds could be solved by:

1. proceed with what you have and make the client responsible for constructing the complete list of callsets.

1. (again, as you mention could eventually be replaced by field masks - that looks like the right solution assuming the group decides to stick with these schemas)
```
message CallSetIds {
  bool return_calls = 1;
  repeated string callset_ids  = 2;
}

message SearchVariantsRequest {
  ...
  CallSetIds callset_ids = n;
 ...
}
```
Note that the return_calls could also just be an optional bool in the SearchVariantsRequest object you have now, as well. 

with option 2 you happen to get an additional bonus behavior, which is not seen in the API today (which could be useful to some). Here are the four cases:
1. All variants, no call info: empty CallSetIds (bool default, return_calls = false)
2. All variants, all call info: empty CallSetIds (return_calls = true)
3. Only variants belonging to a subset of CallSets, with call info: (return_calls = true, callset_ids not empty)
**4. Only variants belonging to a subset of CallSets, without call info: (return_calls = false, callset_ids not empty)** 

with option 1, you proceed with the current api implementation and are always returning call info given a callset filter. 

Perhaps the question is: would the community find behavior 4 useful?

calbach:
Thanks @jacmarjorie for the feedback. I think in general the usefulness of (4) will depend on the semantics of 'call absence' in the variant set; it translates into a light-weight call to get all variants which are 'relevant' for a given subpopulation. I think it's probably not a huge win in terms of functionality, but in my opinion it's a logical mode to support.

Currently I think there's not much agreement on the primary use cases, perhaps you can add some more insight here from your experience? I'll list out my primary use cases in order of importance as reference from my experience serving the API and interacting with customers:

1. Get variants+calls for all call sets.
2. Get variants+calls for a small number (1-3) of call sets.
3. Get variants-only (for either all call sets, or a limited number).
4. Get variants+calls **excluding** some callsets, (e.g. this `CallSet` may have issues with QC).

So can you technically achieve my primary use case using my current PR? Sure, e.g. by including 8K call set IDs in the request. Is that made easy for the client? Not really. Does that matter? It depends on whether people agree that this is a primary use case.

My latest approach with this PR has been to remove the more controversial aspects of dealing with INNER vs OUTER JOIN semantics by dropping the "short-hand" notation for getting all variant calls. I still believe that the API needs an opinion on this, and I think it *should* have the short-hand, but I may try to tackle this in a follow-up PR if this is needed to unblock proto work. Of course, I'll also be happy if consensus changes quickly and I can change this PR to something similar to what you describe.

diekhans:
How many samples do the Google customer VCFs tend to have?  The
problem with the `all calls' default behavior was that it
generated slow performance because some VCFs had thousands of
samples.

diekhans:
An option that explicitly specifies the behavior rather than
implicitly specifying it in an option would be both clearer and
eliminate these trade offs were overriding the implicit behavior
is bad for someone.

I would rather have much clearer semantics than avoid having an
extra option.

jacmarjorie:
After trying to catch up with the thread, I believe the scope needs to be reduced in order for @calbach to proceed with the protobuf definitions. So can an executive decision be made about **how to support returning calls (or not returning calls)** and move all other things (ordering ad inner/outer joins) to a separate issue? 

I believe that @diekhans is suggesting that the presence of a bool flag in the SearchVariantsRequestObject would provide a functionality that would explicitly say *return calls* or *not return calls*. This seems to be in line with what @haussler is saying in making a distinction between a null value and not asking for a value of a certain field. I believe this approach would allow @calbach to continue and would address all possible use cases associated with variants/call objects being returned. This would leave inner and outer join up to the implementation for now until the group wants to make a more firm decision on if they want to support inner, outer, or both. 

Here is some discussion on the topic, but want to stress the importance of the question @calbach wants answered in ordered to merge this PR. 

TL;DR.

Summary of possible use cases (all of which are valid and probably should be supported by the API) ignoring ordering, that is:
**1. all variants, no call info**

**2. all variants, all call info**
  **a. call info only contains calls from callsets with that variant (INNER JOIN)**
  b. call info contains all callsets in the variant set, absense of the reported variant on a callset gives [0,0] genotype? or the call list is empty? (OUTER JOIN)

**3. only variants belonging to a subset of callsets, with call info**

4. only variants belonging to a subset of callsets, without call info

(The bold are the use cases I frequently encounter, however I have performed 2b. client side).

Note the documentation appears to argue for an outer join (at least when call set id is not present), which would be consistent with a multi-sample VCF:
```
  /**  The name of the call set this variant call belongs to.  If this field is not present, the ordering of the call sets from a  `SearchCallSetsRequest` over this `VariantSet` is guaranteed to match  the ordering of the calls on this `Variant`.  The number of results will also be the same.  */
```

With all that - I don't think the argument is whether one of these use cases is more valid or not because a case could be argued for each one, and removing one of the above would limit the API or require additional processing on client side (as I have done with 2b before). 

If you think about how one is pulling down variants right now through the API, the standard procedure is /datasets/search, /variantssets/search, /variants/search, it would add yet an additional call (to essentially retrieve similar data to separately request the calls - and even then you would still want the variant information associated with these calls). That makes it hard to argue for an additional /calls/search function which would essentially duplicate /variants/search with call information.

And finally, I'm happy to open these up as an issue (if there isn't already one created).

diekhans:
Instead of a boolean flag, I would suggest a enum to define the
operation. There are current four proposed use cases.  Even with
two cases, a enum is clearer and allows for expansion.

The array of ids then become an argument to the operation.

calbach:
@diekhans - thousands of `CallSet`s per `VariantSet` seems to be typical from the public and/or widely available datasets I'm familiar with. Not sure if the question was informational, or a possible argument against this feature, if it's the latter: at the end of the day, if the client actually wants all calls, lack of API support is just putting a roadblock in their way and not avoiding the performance issue.

Note: The pagination semantics at least allow you some flow control on the server; in the most extreme case you could limit yourself to returning 1 variant (w/ all calls) per page to negate performance issues. Google's implementation will do this response sizing based on the number of calls returned.

@jacmarjorie thanks for adding those, this is very useful to the discussion. I think the INNER vs OUTER JOIN issue should perhaps be split off from this PR; feel free to create an issue if you'd like to. Note: if it's an OUTER join, the interaction between a possible `retrieveCalls` and `callSetIds` field is dubious - but perhaps this can be overlooked for the time being.

RE: boolean field
---

There's also a small question of whether it should be `retrieveCalls` or `maskCalls`, i.e. what is the default behavior. My preference is for the latter, following my personal rule of 'the most obvious query should yield the most obvious result' - in other words don't drop fields which are documented in the response value unless that behavior is explicitly requested. Based on previous comments, I would guess others may disagree with this as the default behavior.

Sounds like there is some vague support for adding this boolean flag. I'd like to hear @jeromekelleher's thoughts as well.

jacmarjorie:
@calbach Current behavior is callSetIds = null as default. This would be equivalent to `retrieveCalls`, no? And would be consistent to not dropping fields that are documented in the response.

diekhans:
@calbach the question was just informational.  I was curious
about the data that google genomics users create, if there are
any patterns there that could be informative to the right
default.

The flow control wasn't the issue with performance.  It was the
cost of converting VCF to JSON and then read large data that
isn't needed.

Personally, I think variants containing calls vs a query to
return them isn't the best representation.  It's an
indeterminate sized object requiring these kind of queries.

calbach:
@jacmarjorie yes, the current default behavior is to return all variant calls which is equivalent to setting `retrieveCalls` to true. In proto3, the default value for a boolean field is `false` if left uninitialized, which is why I was suggesting the inverse, i.e. `maskCalls` defaults to `false` so all calls are returned by default. The client would need to explicitly set this to `true` to get variants-only.

RE: enum

I don't think there is a need to support both INNER and OUTER JOINS, I think we need to decide which one is correct (I feel strongly that INNER is correct) and make this clear in the documentation. Note that this distinction should be irrelevant to a VCF-backed implemented, as all variants are guaranteed to have a `Call` entry for all `CallSet`s.

diekhans:
RE: default behavior

I don't know what the right default behavior is, however I think
we should talk to people who do variant analysis as part of
their day to day work, like @lh3, and not optimize for either
the google data center or the reference server.

RE: enum

I don't know if we need outer join or not, however enum makes it cleaner
to extend for the future.

Either enum or boolean (with the possibility of adding more booleans)
is better than implicit behavior based on what is specified.  Although
all of this makes SQL look eloquent.

Summary:
Lets talk to people who actually use VCF.

jeromekelleher:
I'm afraid I still don't understand why we need the 'return all calls' special case. The INNER JOIN vs OUTER JOIN semantics is an orthogonal question, since we must define the JOIN semantics for any subset of all callsets in any case (won't there be variants that don't have calls even if we have only one callSetId?). Having the possibility of different semantics when we get all callsets is yet another argument for not having what is (in my opinion and experience) an ugly and problematic special case.

david4096:
@calbach Given the switch to protobuf can we close?

PR #493, opened 2015-11-16
removed unused kruft to minimize diffs with documentation branch

reece:
NOTE: I removed the contrib/ directory, which is absent on the documentation branch and (I assumed) unused. If that's not right, do not merge.


diekhans:
+1 contrib is for the doxygen extraction, which has been replaced.

macieksmuga:
+1 Thanks Reece!

calbach:
+1 since the relevant parties have said this is no longer needed.

I would vote to waive the 'wait at least 3 days' merge rule here since it's a code cleanup which doesn't affect the actual API.

jeromekelleher:
What about ``make_avrodoc.sh`` though? Is this something we need for the  upcoming 0.6.0a1 release?

adamnovak:
@jeromekelleher may have a point. I think the docs we get out of Sphinx (like at http://ga4gh-schemas.readthedocs.org/en/latest/schemas/readmethods.html) are a suitable substitute, but we might have to do work to get them up to release quality (at least fixing the doubling). Until then, maybe we should keep the avrodoc script around in the scripts folder?

Alternatively, if we're so close to release that we don't want to remove stuff that we want to replace in the next release (i.e. Avrodoc), should we maybe create an 0.6.0a1 release branch?

reece:
I think I've missed a memo about the release plans, so I'm flying a bit blind about what's safe to remove.

I *thought* that the next milestone was to merge the documentation branch into master. Once that happens, we don't need contrib, right?

Or, are you saying we're going to push a release of master *before* the documentation branch merge? If so, then I can see why contrib needs to be kept around.

Help!

diekhans:
We are a little short on memos; but I would love to get the
documentation branched merged before releasing.  It seems
really close.

reece:
... which means that 86'ing contrib is okay, right?

jeromekelleher:
OK, let's nuke it then. It's easy enough to find the scripts if we need them.

+1.

Going with @calbach's idea of waiving the three day wait as this doesn't affect the API.

PR #494, opened 2015-11-16
removed tools/sphinx/generate_sphinx_docs.sh to minimize diffs with m…

reece:
Minor changes to minimize non-rst diffs with master.

IMPORTANT: removes tools/sphinx/generate_sphinx_docs.sh, which is currently a wrapper around the top-level Makefile. Should be okay, but Mark may know of callers that will fail upon removal.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,11 +5,6 @@ target
 doc/source/schemas/*.avpr
 build
 
-# remove this if we start pushing autogen'd rst files
-# in order to build at readthedocs
-doc/source/schemas/*.rst
-
-
 #********** windows template**********
 
 # Windows image file caches
--- a/pom.xml
+++ b/pom.xml
@@ -179,7 +179,10 @@
           </execution>
         </executions>
         <configuration>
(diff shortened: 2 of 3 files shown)


diekhans:
generate_sphinx_docs.sh is used for maven to generate the doc:

./pom.xml:182:          <executable>tools/sphinx/generate_sphinx_docs.sh</executable>

macieksmuga:
Perhaps this is one change that should wait until we merge docs into master?

adamnovak:
We definitely can't merge it if it won't build.

@reece : can you make Maven call make directly?

reece:
@adamnovak, @diekhans -- modified as suggested by Adam.
It feels very inside out (to me) to have mvn calling make. Mark suggested that we can move to make when move to protobuf. I like that. And make can always call mvn for java building, of course.

adamnovak:
+1 on this since it builds now.

dglazer:
@diekhans -- is this mergeable?  (Just trying to tidy up loose ends, so we can get closer to retiring the branch.)

PR #495, opened 2015-11-17
Sync variants.avdl and variantmethods.avdl with main schema (only all…

pcingola:
Sync variants.avdl and variantmethods.avdl with main schema (only all…elic model, no graph)


sarahhunt:
Good idea!

PR #498, opened 2015-11-20
Add variant annotation schema diagram

andrewjesaitis:
This is the current schema diagram for the variant annotation team. It matches the variation_annotation branch as of November 20.


dglazer:
@andrewjesaitis -- nice diagram.

@diekhans -- process question -- where should documentation like this, for features that aren't yet in master, go? My prior is it should live in the appropriate topic branch, not in this documentation branch, since we want to merge everything in this branch into master shortly. Your call, though.

diekhans:
Since this is specific to the variant annotation work, lets
merge this into the variant annotation branch.

diekhans:
please make this a pull request against the variant annotation  task team branch.

mbaudis:
@andrewjesaitis In your diagram, ```Metadata``` is neither in sync with the current master nor the metadata branch. I really would like to touch base regarding the requirements, which could be helpful in getting branches in sync (e.g. ```Sample``` => ```BioSample```,  ```Analysis```?).

andrewjesaitis:
@mbaudis Right, I see that looking at master directly. It's actually something I struggled a bit with trying to sync this up with the schema in the variant_annotation branch. Originally, this was a internal diagram just to help understand the current state of the schema across branches. I simplified it to only use the variation_annotation branch, but I see at least two inconsistencies currently. Maybe it's best to hold off on the PR until these differences get incorporated so that the schema is stable when documented. But, I'm glad to have the diagram serve as a starting point for highlighting inconsistencies and getting things in sync.

Also note that I deliberately abbreviated protocols outside of the variant task team's core area, as I imagined  this diagram being in the annotation section of the docs. Ultimately, I think a master diagram will need to be created using the Adam's scripts in contrib, but I was having some problems getting them to generate all the new modules, so I'll have to dig into it some more.

@diekhans Sure thing.

PR #499, opened 2015-11-21
Revert "Added Fragment to Reads"

calbach:
This reverts commit 468b5af6555c2f88b66a3d0fdb36c3052ae69738.

The `Fragment` record has been checked in as an unused orphan for over a year. This resource needs more design work and supporting API methods before being proposed as an addition to the reads API.

### DIFF
--- a/src/main/resources/avro/reads.avdl
+++ b/src/main/resources/avro/reads.avdl
@@ -177,18 +177,6 @@ record LinearAlignment {
 }
 
 /**
-A fragment represents a contiguous stretch of a DNA or RNA molecule. Reads can
-be associated with a fragment to specify they derive from the same molecule.
-TODO: this Fragment object is essentially unused, and may be removed in a future PR.
-*/
-record Fragment {
-
-  /** The fragment ID. */
-  string id;
-
-}
-
-/**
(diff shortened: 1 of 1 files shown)


dglazer:
+1 -- agree with the rationale.

calbach:
@jeromekelleher please take a look when you have a chance. I believe this is the change you requested on the release PR.

jeromekelleher:
Yeah, definite +1 here.

calbach:
We have two +1s, though one is also from Google (the same organization to send the PR). Could we have a third party merge this? UCSC perhaps - @benedictpaten @diekhans ?

pgrosu:
Just to be sure, probably might it might be good to check with @saupchurch and @delagoya on this to be sure based on the initial request on #212, since I don't see their comments here.

saupchurch:
We need this eventually, but the initial observation is true.  In the interests of cleaning up the repositories and moving to our "everything is tested and integrated together in a concerted way"
+1

pgrosu:
Sounds reasonable.

Thanks,
~p

-----Original Message-----

jeromekelleher:
This sounds like consensus to me; merging.

PR #502, opened 2015-11-23
merge documentation branch into master

reece:
This is the moment we've all been waiting for!

Major changes:
- new documentation outline
- significant new content and editing
- semi-automated doc builds (make docs)

**Terrific community effort everyone!**

The documentation branch will be retired after this merge.
(I think this is done with git push upstream --delete documentation... but let's wait until the merge!)

I have verified that docs build correctly from this PR.

PR #503, opened 2015-11-23
Documentation

reece:
Changes since #502:
* removed sample_code stub from appendix/index.html (rst content remains, but is unreferenced)
* removed Makefile-broken
* removed ga4gh-objects.png
* removed empty schemas/intro.rst (and reference in index.rst)
* removed ref to documentation branch in doc/README.rst
* updated rst files derived from avdl files


dglazer:
+1 -- good cleanup. Let's get it in, and then re-do the merge-to-master PR.

(I didn't read the diffs in the auto-generated files closely.)

reece:
What's our protocol... may I merge my own PR on non-master branches after someone +1s? 
@benedictpaten @diekhans @calbach

calbach:
The official protocol is 'the task team which owns the topic branch may decide on their own review policy'. In this case, there's not really a task team in charge and I think Mark has been the defacto merge-master for this repo. 

I don't know if Mark is around today and this seems uncontroversial to me, so I'll just go ahead and click merge. We can follow-up with Mark if there should have been a different process followed here.

reece:
On Mon, Nov 23, 2015 at 3:44 PM, CH Albach <notifications@github.com> wrote:

> I don't know if Mark is around today and this seems uncontroversial to me,
> so I'll just go ahead and click merge. We can follow-up with Mark if there
> should have been a different process followed here.


Thanks. Just want to make sure that don't go off the reservation.

PR #506, opened 2015-11-30
Update metadata.rst

mbaudis:
Cleaning up some errors / wrong descriptions (e.g. every object belongs to one "Dataset", which had never been defined thusly), and adding of some format definitions.

### DIFF
--- a/doc/source/api/metadata.rst
+++ b/doc/source/api/metadata.rst
@@ -66,12 +66,6 @@ The specification of a time point is given through the concatenation of
 * time stamp in milliseconds in YYYY-MM-DDTHH:MM:SS.SSS
     2015-02-10T00:03:42.123Z
 
-**Implementations**
-
-* recordCreateTime (G2P, metadata)
-* recordUpdateTime (G2P, metadata)
-* many proposed in the metadata branch
-
 ===========
 Durations
 ===========
--- a/doc/source/api/metadata.rst
+++ b/doc/source/api/metadata.rst
@@ -10,20 +10,117 @@ Goals and Scope
 
 The metadata API provides information on the primary data objects
 available via the GA4GH API, and facilities to organize primary data
-objects.
+objects. Main goals of the metadata definitions are to provide
 
-The current metadata API is immature and will evolve in future.
+* standardized use of common attributes/values
+* a schema of how (core) objects relate to each other
+
+The  metadata API is under continuous development.
 
 
-Datasets
-@@@@@@@@
+Common Attribute Names and Formats
+@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+
+Throughout the schema definitions, a consistent use of common attributes should
+be enforced. The following list should serve as guidance for schema developers.
 
 
-All GA4GH data objects are part of a *dataset*. A dataset is a
+========================= ======================================================
+Attribute                 Note
+========================= ======================================================
+id                        the objects ID, used for references; (at least) locally unique
+name                      a more descriptive object label or legacy ID; should *not* be used as reference
+accessions                an ``ARRAY`` containing other names, IDs, URIs of the object
+description               a string describing aspects of the object; *not* a list or nested object itself
+recordCreateTime          time at which this record was created in ISO 8601 (see **Date and Time...**)
+recordUpdateTime          time at which this record was updated in ISO 8601 (see **Date and Time...**)
+========================= ======================================================
+
+Date and Time Format Specifications
+@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+
+Date and time formats are specified as ISO8601 compatible strings, both for
+time points as well as for intervals and durations.
+The required granularity has to be specified as part of the respective
+attributes' documentations.
+
+===========
+Time points
+===========
+
+The specification of a time point is given through the concatenation of
+
+* a date in YYYY-MM-DD
+* the designator "T" indicating a following time description
+* the time of day in HH:MM:SS.SSS form, where "SSS" represents a decimal fraction of a second
+* a time zone offset in relation to UTC
+
+**Examples**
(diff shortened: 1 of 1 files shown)


doc/source/api/metadata.rst
```diff
@@ -10,20 +10,117 @@ Goals and Scope
 
 The metadata API provides information on the primary data objects
 available via the GA4GH API, and facilities to organize primary data
-objects.
+objects. Main goals of the metadata definitions are to provide
 
-The current metadata API is immature and will evolve in future.
+* standardized use of common attributes/values
+* a schema of how (core) objects relate to each other
```
diekhans: what is a `core' object?  Throwing in a new term is confusing.

doc/source/api/metadata.rst
```diff
@@ -10,20 +10,117 @@ Goals and Scope
...
+    2015-02-10
+
+* time stamp in milliseconds in YYYY-MM-DDTHH:MM:SS.SSS
+    2015-02-10T00:03:42.123Z
+
+**Implementations**
+
+* recordCreateTime (G2P, metadata)
+* recordUpdateTime (G2P, metadata)
+* many proposed in the metadata branch
```
diekhans: G2P is not currently part of master.  Drop the section Implementations; this will be quickly out of date and confusing

doc/source/api/metadata.rst
mbaudis: Removed. Basically referred to the concept of main objects in the hierarchy
	Individual <=> BioSample <=> (VariantCall or such)

… vs. additional information objects (Experiment, Ontologies …).

doc/source/api/metadata.rst
```diff
@@ -10,20 +10,117 @@ Goals and Scope
...
+
+* experimental intervention of 30min in YYYY-MM-DDTHH:MM/YYYY-MM-DDTHH:MM
+    2014-12-31T23H45M/2015-01-01T00H15M
+
+
+Object: Dataset
+@@@@@@@@@@@@@@@
+
+GA4GH data objects can be part of one or several *dataset*(s). A dataset is a
 data-provider-specified collection of related data of multiple types.
```
diekhans: This needs to be discussed.  i believe the original design is that every object belong to one and only one data set.

doc/source/api/metadata.rst
mbaudis: > Drop the section Implementations; this will be quickly out of date and confusing

But then, I have always been told that things don’t get into master if they aren’t called anywhere?

Removed.

doc/source/api/metadata.rst
```diff
@@ -10,20 +10,117 @@ Goals and Scope
...
+
+* age with date of birth in YYYY-MM-DD/PnYnMnD
+    1967-11-21/P40Y10M05D
+
+* anchored 3 month interval, e.g. a therapy cycle in YYYY-MM-DD/YYYY-MM-DD
+    2015-04-18/2015-07-17
+
+* experimental intervention of 30min in YYYY-MM-DDTHH:MM/YYYY-MM-DDTHH:MM
+    2014-12-31T23H45M/2015-01-01T00H15M
+
```
diekhans: Time intervals are not currently implemented in master.  The documentation should be in a topic branch until it's needed.   We can create a separate metadata formats push request to launch a discussion

doc/source/api/metadata.rst
mbaudis: > This needs to be discussed. i believe the original design is that every object belong to one and only one data set
> 

It will be discussed, but the concept is crazy. What do you do with your “Dataset"s of “childhood cancers” and “medulloblastomas”? And so on.

There is also absolutely no procedural treason for limiting to one. Also, since objects have uniqueIDs ...

diekhans: Oh!  I like the idea of "procedural treason".

This goes back to the function of `Dataset',  using dataset
in place of metadata queries will result in an explosion
of datasets. One view is that dataset as something very
similar to NCBI BioProject.

mbaudis: Oh my. It is more of a confusion stemming from the "IndividualGroup" object, I guess, and then the use of "Dataset" in reads for grouping reads, and then we agreeing on a wider scope (w/ me and then @dglazer writing some definitions), and now seeing this turned into a "BioProject" or "Series" like object, w/o real definition in other parts of the schema.

The problem with "Dataset cf. BioProject" is, that in such a usage it would be extremely vendor dependent - and actually there are repositories with data not being part of a traditional project (e.g. OpenSNP ...).

Therefore, it makes much more sense to have this as an open aggregator, instead of a monolithic & required structural element.

mbaudis: I’ve tried to fix the BioSample PR, by fixing reads/variants (which were obviously erroneous - why didn’t anybody take care of this???).

https://github.com/ga4gh/schemas/pull/397

This was supported before, only didn’t get in due to your objections since it wasn’t “bioSampleId” etc.

=> ???

Michael.

diekhans: We should setup a topic branch for bring in new metadata
objects, as we need to implement them in the server and
compliance suite before moving to master.  maybe metadata_stage ?

mbaudis: BioSample is not a new object; there was Sample for a while, and then all non referenced objects were moved out of master and only kept in the metadata branch, but actually reads/variants either had references in them or introduced them after it was moved out, and then we had these "sample" discussions so we renamed Sample to Biosample and ...

So there is absolutely no reason for keeping BioSample out of the release. It is basically only due to your aversion that I didn't push it... Not accepting this in is the single biggest obstacle to soooo much development. And if anybody objects, it is on me. But besides some technical demonstration, all the schema is absolutely worthless with any reference object (even ifit is rather void in this PR, in contrast to the metadata branch).

PR #509, opened 2015-12-08
Clarify mappings of SAM flags to read fields

dcolligan:
Based on discussion here https://github.com/ga4gh/server/pull/789

### DIFF
--- a/src/main/proto/ga4gh/reads.proto
+++ b/src/main/proto/ga4gh/reads.proto
@@ -143,6 +143,9 @@ message ReadAlignment {
   bool duplicate_fragment = 5;
 
   // The number of reads in the fragment (extension to SAM flag 0x1).
+  // A value of 1 means SAM flag 0x1 is unset.
+  // A value of [2, infinity) means SAM flag 0x1 is set.
+  // TODO what do other values mean?
   int32 number_reads = 6;
 
   // The observed length of the fragment, equivalent to TLEN in SAM.
@@ -151,13 +154,19 @@ message ReadAlignment {
   // The read ordinal in the fragment, 0-based and less than numberReads. This
   // field replaces SAM flag 0x40 and 0x80 and is intended to more cleanly
   // represent multiple reads per fragment.
+  // A read_number of 0 represents SAM flag 0x40 being set and 0x80 being
+  // unset.
+  // A read_number of 1 represents SAM flag 0x80 being set and 0x40 
+  // being unset.
+  // A read_number other than 0 or 1 means both SAM flags 0x40 and 0x80 are 
+  // set.
   int32 read_number = 8;
 
   // The read fails platform or vendor quality checks (SAM flag 0x200).
   bool failed_vendor_quality_checks = 9;
 
-  // The alignment for this alignment message. This field will be null if the
-  // read is unmapped.
+  // The alignment for this alignment message. This field will be unset if the
+  // read is unmapped (i.e. if SAM flag 0x4 is set).
   LinearAlignment alignment = 10;
 
   // Whether this alignment is secondary. Equivalent to SAM flag 0x100.


src/main/resources/avro/reads.avdl
```diff
@@ -190,7 +190,12 @@ record ReadAlignment {
   /** The fragment is a PCR or optical duplicate (SAM flag 0x400). */
   union { null, boolean } duplicateFragment = null;
 
-  /** The number of reads in the fragment (extension to SAM flag 0x1) */
+  /**
+   The number of reads in the fragment.
+   A value of 1 means SAM flag 0x1 is unset.
+   A value of 2 means SAM flag 0x1 is set.
+   TODO what do other values mean, especially null?
```
macieksmuga: can somebody familiar with reads comment on this? The null union here seems spurious to me based on the documentation so far.

macieksmuga:
+1 modulo the question about null union in number of reads.
Since this PR is purely about improving the quality of documentation (not changing the meaning of any fields), I believe it's safe not to wait on this to be merged for the 0.6.0a1 release.

src/main/resources/avro/reads.avdl
```diff
@@ -202,6 +207,12 @@ record ReadAlignment {
   The read ordinal in the fragment, 0-based and less than numberReads. This
   field replaces SAM flag 0x40 and 0x80 and is intended to more cleanly
   represent multiple reads per fragment.
+  A readNumber of 0 is equivalent to SAM flag 0x40 being set and 0x80 being
+  unset.
+  A readNumber of 1 is equivalent to SAM flag 0x80 being set and 0x40 being
+  unset.
+  A readNumber other than 0 or 1 means both SAM flags 0x40 and 0x80 are set.
```
calbach: The wording throughout this PR seems off to me. SAM flags are not 'set' in the GA4GH APIs. Some property of the GA4GH API may be equivalent to a SAM flag, for instance our comment for `properPlacement`: `(equivalent to SAM flag 0x2)`.

src/main/resources/avro/reads.avdl
```diff
@@ -190,7 +190,12 @@ record ReadAlignment {
   /** The fragment is a PCR or optical duplicate (SAM flag 0x400). */
   union { null, boolean } duplicateFragment = null;
 
-  /** The number of reads in the fragment (extension to SAM flag 0x1) */
+  /**
+   The number of reads in the fragment.
+   A value of 1 means SAM flag 0x1 is unset.
```
calbach: Technically the mapping to SAM would be: 

```
1 -> 0x1 unset
[2, infinity) -> 0x1 set
```

I would guess that `null` means an unknown number of reads, or it shouldn't be allowed.

src/main/resources/avro/reads.avdl
macieksmuga: I _think_ that's what this PR is already saying. My naive understanding is that `readNumber` should correctly represent an arbitrary number of reads per fragment, as opposed to SAM's two-flag mechanism used to represent 0, 1 or many... (or should it be 1, 2 or many?)

richarddurbin: I think that the current comment is clear and OK.  This is saying very clearly that each GA4GH API value of 
readNumber is equivalent to a particular combination of SAM flag states - there is no single equivalence to
any one tag.  It doesn’t matter whether flags are set or unset in the API - they are in SAM, and it is
their SAM state that is being referred to.
 
> On 16 Dec 2015, at 02:47, CH Albach <notifications@github.com> wrote:
> 
> In src/main/resources/avro/reads.avdl <https://github.com/ga4gh/schemas/pull/509#discussion_r47731145>:
> 
> > @@ -202,6 +207,12 @@ record ReadAlignment {
> >    The read ordinal in the fragment, 0-based and less than numberReads. This
> >    field replaces SAM flag 0x40 and 0x80 and is intended to more cleanly
> >    represent multiple reads per fragment.
> > +  A readNumber of 0 is equivalent to SAM flag 0x40 being set and 0x80 being
> > +  unset.
> > +  A readNumber of 1 is equivalent to SAM flag 0x80 being set and 0x40 being
> > +  unset.
> > +  A readNumber other than 0 or 1 means both SAM flags 0x40 and 0x80 are set.
> The wording throughout this PR seems off to me. SAM flags are not 'set' in the GA4GH APIs. Some property of the GA4GH API may be equivalent to a SAM flag, for instance our comment for properPlacement: (equivalent to SAM flag 0x2).
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/ga4gh/schemas/pull/509/files#r47731145>.
> 




-- 

david4096:
@dcolligan when you have a chance could you review, close, or update this?

dcolligan:
@david4096 updated

src/main/proto/ga4gh/reads.proto
```diff
@@ -151,13 +154,19 @@ message ReadAlignment {
...
+  // A read_number other than 0 or 1 means both SAM flags 0x40 and 0x80 are 
+  // set.
   int32 read_number = 8;
 
   // The read fails platform or vendor quality checks (SAM flag 0x200).
   bool failed_vendor_quality_checks = 9;
 
-  // The alignment for this alignment message. This field will be null if the
-  // read is unmapped.
+  // The alignment for this alignment message. This field will be zero if the
```
david4096: `unset` maybe? does it come out 0?

david4096:
Thanks @dcolligan there is still a lot of complexity hidden in the conversion from BAM to GA4GH read representation and giving implementors more to go on is helpful.

src/main/proto/ga4gh/reads.proto
david4096: Saying the field will be `0` confused me. I think you mean to say if it isn't set perhaps?

dcolligan: yes, you're right... changing...

david4096:
Thanks, reminds me that we need a way to get back only unmapped reads!

PR #514, opened 2015-12-15
updated version of the beacon api

mcupak:


buske:
If two samples have the same variant, does that count as 2 variants or 1 variant toward the `variantCount` field?

mcupak:
@buske That's a very good question. The number I was going for is 2. Open question: Should we add this to the description of the field or add another field to capture this value?

mcupak:
Feature requests raised on the last beacon call:
- support for multiple-dataset queries
- consent codes

mfiume:
AFAIK, GA4GH typically uses Variant to refer to a unique position + allele pair, so I would interpret variantCount to be 1. What metric would be most useful? If the total number of "calls" is useful, should we add a callCount?

mcupak:
I hope all the issues raised above have been addressed:
- I added a `callCount` field to have data for both variants and calls.
- I added support for multiple datasets, which required some other changes, such as error handling.
- I added `datause.avdl` for consent codes. The data is linked to a datasets.

PR #515, opened 2015-12-15
updated version of the beacon api

mcupak:
Complement to #514 without datasets.


mcupak:
As per discussion on the call today, this has been rejected in favour of #514. Closing.

Issue #516, opened 2015-12-16
Review terminology for predicted molecular impact


sarahhunt:
Discussion of the variant annotation documentation revealed concerns and potential confusion about the current classification of predicted molecular impact and in particular the term 'MODIFIER' to describe variants with no predicted impact on protein function.

This issue records the need to address these concerns and seek a prefered term.

majensen:
I wanted to +1 @andrewjesaitis view expressed on the call today that the schema can represent "known" unambiguous facts from SO regarding variants, which can be grouped in any desired way at the application level. Also to +1 the ability nevertheless to store in the schema groupings that are commonly used or have other evidence or support (along with the provenance of the groupings). The concern I would have, and that was mentioned in the, is that if the _putative_ effects are baked into the schema without caveat, then that might be interpreted as an endorsement by GA4GH. 

One razor the group could apply might be, for any schema-derived assertion, would a citation of "Variant Annotation schema, GA4GH" be acceptable support to you as a review of a paper. "This is a frameshift variant" would probably fly, but "This is a variant of high effect in disease X" would have me looking for more evidence.

gabeiscoding:
I was also +1 @reece comment about simplifying as much as possible.

My proposal on the call was to change the `impact enum {HIGH, MODERATE, LOW, MODIFER}` to a `sequenceOntologies array<enum> { <all_so_terms> }` as that provides an external authority to define the meaning of this term.

Deanne mentioned these annotations should have the program's providence provided, but I believe that is already the case as TranscriptEffect is provided as a part of VariantAnnotation, which references the varinatAnnoationSetId containing the "Analysis" meta-data that can include the program name, version and parameters.

While users are used to asking for "High" impact or "Loss of Function" variants from analysis tools, the goal of this API is to provide a data representation of called variants and community driven attributes that can be algorithmically assigned to them such as HGVS names and Sequence Ontology terms. Since the [SO site](http://www.sequenceontology.org/browser/obob.cgi) does not provide these groups as part of the standard, and its easy to provide them as a grouping of SO terms at an application level, why place them in the schema?

pcingola:
Correct me if I'm wrong, but the main idea of "putative impact" is to
create a
simple categorization for fast filtering purposes having sensitive
defaults.
  It is true that people can use an enumeration of SO terms, but that
becomes
inconvenient very soon as the filter expressions grow. Removing this
category
would reduce the schema at the cost of increasing the work for everyone
else trying to filter variants.
[...]

PR #519, opened 2015-12-19
Variant Annotation re-merge

sarahhunt:
This PR re-merges the Variant Annotation API first discussed in #302, now with additional documentation, a reference server implementation and compliance tests.

Major new developments have been avoided. The differences are:
- removal of all graph-related functionality
- rename AnnotationSet to VariantAnnotationSet for clarity
- a VariantAnnotationSet is now derived from a specific VariantSet, rather than a Dataset
- AlleleLocation overlapEnd can be null
- SearchVariantAnnotationsRequest feature_ids key renamed to featureIds to match the standard
- avro corrections - addition of missing null values


diekhans:
+1  this longs sane and a good step forwards within the restrictions of the current variant schema.

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,205 @@
...
+  long end;
+
+  /**
+  Only return variant annotations for any of these features
+  Features may include specific transcripts, genes or regulatory elements
+  This or a location (referenceName/referenceId plus optional start and end)
+  must be supplied.
+  If null, return all variant annotations in specified window.
+  */
+  union { null, array<string> } featureIds = null;
```
diekhans: The behavior of the feature search is not clearly specified.  If a gene feature id is specified, will the search include all of it's children ids?

david4096: @sarahhunt As far as I know `featureIds` are taken verbatim from a variant annotation file and are not GA4GH generated, is this correct? We don't have a taxonomy of features to draw from other than the raw data.

@diekhans could you expand on what does and does not work with this approach? May we suffice to say that, no, since we don't have a priori knowledge of a feature taxonomy that we are simply matching strings?

sarahhunt: @david4096 - that's how it looks from the test data, but we have databases of gene and transcript sets which we will be making available through the GA4GH API.

@diekhans - Yes - if a gene is specified I would expect the search include all of it's child features ( currently transcripts), as available to the VariantAnnotationSet. I'll change the comment to make this more clear.

doc/source/api/alleleAnnotations.rst
```diff
@@ -0,0 +1,93 @@
...
+The ``VariantAnnotation`` data model, is based on the results provided by variant 
+annotation programs such as VEP, SnpEff and Annovar and others, as well as the 
+VCF's `ANN format <http://snpeff.sourceforge.net/VCFannotationformat_v1.0.pdf>`_ . 
+
+
++---------------------+---------------------------------------------------------------------------------------------------------------------+
+| Record              | Description                                                                                                         |
++=====================+=====================================================================================================================+
+| VariantAnnotationSet| A VariantAnnotationSet record groups VariantAnnotation records. It represents the comparison of a VariantSet to a   |
+|                     | specified reference data set using specified algorithms. It holds information describing the software and reference |
```
jeromekelleher: Both 'reference' and 'data set' have specific meanings within the schema - are we talking about precisely these objects?

doc/source/api/alleleAnnotations.rst
```diff
@@ -0,0 +1,93 @@
+
+Variant Annotation API
+!!!!!!!!!!!!!!!!!!!!!!
+
+See `Allele Annotation schema <../schemas/alleleAnnotations.html>`_ for a detailed reference.
```
jeromekelleher: I'm probably being dense, but what is the distinction between `allele` and `variant` here? They seem to be used more-or-less interchangeably in this documentation. Why `Variant Annotation API` and then the `Allele Annotation schema`? If they are interchangeable, perhaps it would be better to standardise on one (Variant?)?  If not, I think we need some discussion at the start to clarify this point.

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,190 @@
...
+
+*/
+
+/**
+An AnalysisResult record holds the output of a prediction package such
+as SIFT on a specific allele.
+*/
+record AnalysisResult {
+
+  /** The analysis record for this result (defined in metadata schema)  */
```
jeromekelleher: Why is the parenthetical comment needed here?

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,190 @@
...
+  Analysis analysis;
```
jeromekelleher: Should the `Analysis` be embedded here, or would it be more consistent to have an `analysisId` referring to an object on the server? Aren't there going to be a lot of AnalysisResult objects, and wouldn't it be wasteful to include the full Analysis object in each one?

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,190 @@
...
+
+  /** The text-based result for this analysis */
+  union { null, string} analysisResult;
```
jeromekelleher: Do we need the 'analysis' prefix on both `analysisResult` and `analysisScore`? The type is already called AnalysisResult, so It would seem sufficient to have just `result` and `score`.

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+An allele location record holds the location of an allele relative to a
+non-genomic coordinate system such as a CDS or protein and holds the
+reference and alternate sequence where appropriate
+*/
+record AlleleLocation {
+
+  /** Relative start position of the allele in this coordinate system */
+  int overlapStart ;
```
jeromekelleher: Why do we need the `overlap` prefix here? All other location references use `start` and `end`.

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,190 @@
...
+  string id;
+
+  /** The ID of the variant set to which this annotation set belongs */
+  string variantSetId;
+
+  /**
+  Analysis details. It is essential to supply versions for all software and
+  reference data used.
+  */
+  Analysis analysis;
```
jeromekelleher: Is there a good argument for embedding the Analysis object here rather than having an `analysisId`, and having the usual methods for `Analysis` objects? I think we need to have at GET method for the Analysis objects anyway, so I'm not sure there's much to be gained from embedding them into the `VariantAnnotationSet` container.

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,190 @@
...
+/**
+A VariantAnnotationSet record groups VariantAnnotation records. It is derived
+from a VariantSet and holds information describing the software and reference
+data used in the annotation.
+*/
+record VariantAnnotationSet {
+
+  /** The ID of the variant annotation set record */
+  string id;
+
```
jeromekelleher: Should VariantAnnotationSet have a `name` field?

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,190 @@
...
+  union { null, string} alternateBases = null;
+
+  /** Effect of variant on this feature */
+  array<OntologyTerm> effects;
+
+  /** Highest Impact from the predicted effects */
+  Impact impact;
+
+  /** HGVS formatted annotation at genomic level  */
+  union { null, string } HGVSg = null;
```
jeromekelleher: HGVSg, HGVSc and HGVSp aren't the most obvious names... perhaps there should be a HGVSAnnotation (which contains the corresponding fields) that could be embedded in here?

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,190 @@
...
+  represents the effect of a single allele on a single transcript.
+  */
+  array<TranscriptEffect> transcriptEffects = [];
+
+  /**
+  The IDs of other variants which are co-located with this variant.
+  these can use used to look up disease associations, ClinVar statuses,
+  allele frequencies in reference panels, etc
+  */
+  array<string> coLocatedVariants =[];
```
jeromekelleher: What do we mean by co-located here? Are these overlapping variants, or variants that have precisely the same coordinates? Why do we need to provide this when a client and just run a query to find out the same information?

src/main/resources/avro/metadata.avdl
```diff
@@ -114,4 +135,42 @@ record Dataset {
...
+  The time at which this record was last updated.
+  Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
+  */
+  string recordUpdateTime;
+
+  /** The type of analysis. */
+  union { null, string } type = null;
+
+  /** The software run to generate this analysis. */
+  array<string> software = [];
```
jeromekelleher: How does this related to the existing ``Program`` field? Shouldn't we try to reuse this framework?

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,205 @@
...
+*/
+record SearchVariantAnnotationsRequest {
+  /** Required. The ID of the variant annotation set to search over. */
+  string variantAnnotationSetId;
+
+  /**
+  Only return annotation for variants which have exactly this name
+  (case-sensitive, exact match).
+  */
+  union { null, string } name = null;
```
jeromekelleher: This method should mirror SearchVariantsRequest as closely as possible I think. Searching by name has been removed from SearchVariantsRequest, so we should either remove it here also  (or add it to SearchVariantsRequest).

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,205 @@
...
+  name. One of this field or `referenceId` or `features` is required.
+  (case-sensitive, exact match)
+  */
+  union { null, string } referenceName = null;
+
+  /**
+  Only return variants with reference alleles on the reference with this
+  ID. One of this field or `referenceName` or `features` is required.
+  */
+  union { null, string } referenceId = null;
```
jeromekelleher: Why do we have `referenceId` and `referenceName`? Again, this should mirror SearchVariantsRequest as closely as possible, and that currently only has `referenceName`. (It should probably be just `referenceId`, but that's a different issue.)

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,205 @@
...
+
+  /**
+  Required if referenceName or referenceId supplied.
+  The beginning of the window (0-based, inclusive) for which variants with
```
jeromekelleher: Shouldn't this be always required (like SearchVariantsRequest)?

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,205 @@
...
+/**
+This request maps to the body of `POST /variantannotationsets/search` as JSON
+*/
+record SearchVariantAnnotationSetsRequest {
+
+  /**
+  If non empty, will restrict the query to variant annotation sets within the
+  given dataset.
+  */
+  union { null, string } datasetId = null;
```
jeromekelleher: Why do we need to search through both datasets and variantSets? This seems redundant to me --- a client can get the list of variantsets for a given dataset in a seperate query, and then search for the annotation sets for each of these.

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,205 @@
...
+
+  /**
+  If non empty, will restrict the query to variant annotation sets within the
+  given variant set. This takes precedence over any dataset id supplied.
+  */
+  union { null, string } variantSetId = null;
```
jeromekelleher: This should be a mandatory argument, following the pattern of `SearchVariantSets`, etc.

doc/source/api/alleleAnnotations.rst
sarahhunt: I've changed 'reference data' to  to gene annotation data to 'gene annotation data' in #521 to clarify this.

doc/source/api/alleleAnnotations.rst
sarahhunt: This was probably less confusing before we removed the graph component. Annotation - such as protein sequence change - happens at the allele level.  For consistency I've changed the documentation to refer to the Allele annotation API in #521.

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: Thanks - removed in #521

src/main/resources/avro/metadata.avdl
sarahhunt: We are just re-using the metadata defined record here. It probably makes sense to move Program out of reads.avdl and merge the two.

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: When we created this, SearchVariantsRequest allowed search by both referenceId and referenceName, which is why they are both here. Since, referenceId has been removed leaving only referenceName, but as the plan is to switch to referenceId only, we have left both for now.

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: The group were keen to allow the convenience method of searching by featureIds, which is an alternative way to to specify a genomic region.

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: Agreed - this is updated in #523

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: Agreed - this is updated in #523

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: The Analysis object is embeded because that's pretty much all the VariantAnnotationSet is. You would always want the Analysis object info when extracting a VariantAnnotationSet.

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: Agreed - added in  #524

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: Agreed - these are updated in  #522

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: Co-located means at precisely the same location - I've changed the comment to make this clear. I see what you mean, but it's there to reduce the number of queries the client needs to run.

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: Agreed - these are updated in  #522

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: Agreed - updated in #525.

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: I don't know...the comment makes it clear what they are and they are wonderfully succinct.

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: I need to check on this one.

sarahhunt:
Thanks for the detailed review @jeromekelleher! I've responded to the individual comments and created a few PR with the suggested changes. I'll merge these shortly unless there are further comments.

doc/source/api/alleleAnnotations.rst
jeromekelleher: That's much clearer, thanks.

src/main/resources/avro/alleleAnnotations.avdl
jeromekelleher: OK, fair enough. I guess we should provide a GET endpoint for the Analysis objects as well though, since we clients will have access to the IDs? It seems unnecessary to provide a full search endpoint for them.

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: I've checked - this is removed in #526

src/main/resources/avro/alleleAnnotations.avdl
jeromekelleher: Someone reading client code shouldn't have to look up the API documentation to find out what something is though --- they should be able to just look at it and know what it is. If I saw:
```python
transcriptEffect.hgvsAnnotation.genomic
```
in client code it would be much more obvious. 

It would be good to get @reece's opinion here, since he knows a lot more about HGVS than I do!

Either way, I wouldn't block on the point. It can be followed up as an issue after merge IMO.

src/main/resources/avro/alleleAnnotations.avdl
jeromekelleher: I see, but it does complicate things server side quite a bit. Is it really a big enough problem that it needs to go into the first version of the schema? We can always add things to the schema later, but removing things is much, much, much harder. Performance optimisations can be added later when we have hard data proving that they are necessary...

src/main/resources/avro/metadata.avdl
jeromekelleher: @mbaudis, @diekhans any thoughts on this? What's the pragmatic way of getting some decent metadata in here?

src/main/resources/avro/alleleAnnotationmethods.avdl
jeromekelleher: In the interest of simplicity and consistency then I'd remove referenceId. It won't be hard to update both queries at the same time when/if we're changing over to referenceId.

src/main/resources/avro/alleleAnnotationmethods.avdl
jeromekelleher: Have you an example of how this would work? Is there a 1-1 mapping between genomic coordinates and featureIds?

jeromekelleher:
Thanks for the updates and comments @sarahhunt, these were very helpful. I think this API makes a lot of sense, and hopefully we can get some wider input on it so that it can be merged into master soon.

@dglazer, @calbach, have you any thoughts on this?

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: It's not a problem to hold this one for a later version in the interest of speed. Looking at the test data, it's not in the reference server code yet.

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: Yes - there is a 1-1 mapping - GA4GH Features have specific locations on specific References.

src/main/resources/avro/metadata.avdl
diekhans: Yes, please move program to metadata.

src/main/resources/avro/alleleAnnotationmethods.avdl
diekhans: It seems very useful to have both referenceName and referenceId
queries.

We need to work on use cases for the all of GA4GH API to
understand what is needed for these type of queries.

src/main/resources/avro/alleleAnnotationmethods.avdl
jeromekelleher: Is this a critical feature for the first version of the schema then? Conveniences can be added in incrementally...

src/main/resources/avro/alleleAnnotationmethods.avdl
jeromekelleher: That may be @diekhans, but we should keep the API consistent. If we add it to one query for variants it should be in the other as well. This should be done in a follow-up PR, as it's not on the critical path for getting variant annotations into the schema.

diekhans: Completely agree with keeping it consistent!  We need to have a day with much coffee
making the whole API consistent.

pgrosu: I prefer adding a caffeine synthesis pathway to our genome, or this might work in the meantime :)

coffee

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: OK - that one is removed in #529 and I've created an issue to add it back later.

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: @diekhans - is that a vote for or against keeping referenceId?

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: @jeromekelleher - I would be very reluctant to hold this one for later - it was the key use case the task team identified - given a gene or transcript return annotation.

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: Agreed - this can go in the new metadatamethods file @david4096 has just created.

sarahhunt:
@jeromekelleher - thanks for the further comments. I've had a second pass through this and merged the updates into branch so it is cleaner for others to review.

src/main/resources/avro/alleleAnnotationmethods.avdl
diekhans: Until sequence annotations is integrated, there is no way to generate a featureId and no way to implement the query.  I would suggest we just comment out the query definition (with an explanation).  Then we add it back in as soon as sequence annotations are integrated.  I would much rather do this and delay merging VA.

src/main/resources/avro/alleleAnnotationmethods.avdl
diekhans: @sarahhunt, lets see

 Variant has referenceName;
 SearchVariantsRequest takes referenceName

 LinearAlignment.position has referenceName
 SearchReadsRequest uses referenceId

Which is rather inconsistent.  So at this point, I don't care
what alleleAnnotationmethods implements as long as a ticket is
filed to make use of id/name consistent across the entire API.

I think what alleleAnnotationmethods is what is required.

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: Thanks @diekhans - I've followed your suggestion and commented it out of the schemas and compliance suite, but left it in the Ensembl dev implementation as I assume features will be merged soon.

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: Hi @jeromekelleher - I've switched this to a nested record in #531 - it does look neater.

andrewjesaitis:
+1 This looks good to me. I just made a PR to the variant_annotation branch updating documentation by adding the schema diagram.

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,195 @@
...
+//  must be supplied.
+//  If null, return all variant annotations in specified window.
+//  */
+//  union { null, array<string> } featureIds = null;
+
+  /**
+  Only return variant annotations including these effects (SO terms).
+  If null, return all variant annotations.
+  */
+  union { null, array<org.ga4gh.models.OntologyTerm> } effects = null;
```
calbach: How does matching work here exactly? Do all fields of the OntologyTerm have to match exactly?

src/main/resources/avro/alleleAnnotations.avdl
calbach: Is `Analysis` 1:1 with a `VariantAnnotationSet`? If so I would do away with giving it its own ID, seems inconsistent with the rest of the API to have a uniquely addressable resource which is also embedded here.

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/**
+A HGVSAnnotation record holds Human Genome Variation Society descriptions
+of the sequence change with respect to genomic, transcript and protein
+sequences. See: http://www.hgvs.org/mutnomen/recs.html.
+*/
+record HGVSAnnotation {
+
+  union { null, string } genomic = null;
```
calbach: Can something be documented about which of these fields you expect to be populated? e.g. at least one (but presumably more than one is allowed)

src/main/resources/avro/metadata.avdl
```diff
@@ -114,4 +135,42 @@ record Dataset {
 
 }
 
+/**
+An analysis contains an interpretation of one or several experiments.
+(e.g. SNVs, copy number variations, methylation status) together with
+information about the methodology used.
+*/
+record Analysis {
+  /** The analysis UUID. This is globally unique. */
```
calbach: UUID? Why is this a requirement? IMO this would ideally just follow the same description as other IDs in the API.

calbach:
This looks reasonable to me. I added a few minor comments inline.

src/main/resources/avro/metadata.avdl
mbaudis: This is formulated differently in the metadata branch. You should pick up the version from there:

https://github.com/ga4gh/schemas/blob/metadata/src/main/resources/avro/metadata.avdl#L767

Also, IDs are described in https://github.com/ga4gh/schemas/blob/metadata/doc/source/metadata_intro.rst. Not sure if this the right place in the long run...

doc/source/schemas/alleleAnnotations.rst
```diff
@@ -0,0 +1,562 @@
...
+
+  :field id:
+    The call set ID.
+  :type id: string
+  :field name:
+    The call set name.
+  :type name: null|string
+  :field sampleId:
+    The sample this call set's data was generated from.
+      Note: the current API does not have a rigorous definition of sample. Therefore, this
```
mbaudis: It is shouldn't read ```Sample``` but ```BioSample```, and this is defined both in PR https://github.com/ga4gh/schemas/pull/397 (in a basic version) as well as in the metadata branch https://github.com/ga4gh/schemas/blob/metadata/src/main/resources/avro/metadata.avdl#L441.

It would really be helpful if somebody from reads/ref/var would take care of merging 397 into master; it is basically needed by enough objects now, right?

doc/source/schemas/alleleAnnotationmethods.rst
```diff
@@ -0,0 +1,700 @@
...
+    The variant ID.
+  :type id: string
+  :field variantSetId:
+    The ID of the `VariantSet` this variant belongs to. This transitively defines
+      the `ReferenceSet` against which the `Variant` is to be interpreted.
+  :type variantSetId: string
+  :field names:
+    Names for the variant, for example a RefSNP ID.
+  :type names: array<string>
+  :field created:
```
mbaudis: Time formats should all be ISO8601. See https://github.com/ga4gh/schemas/blob/metadata/doc/source/metadata_intro.rst for reference/explanation.
Also, attributes should be ```recordCreateTime``` and ```recordUpdateTime```, and the type ```string```.

doc/source/schemas/alleleAnnotationmethods.rst
```diff
@@ -0,0 +1,700 @@
...
+    The sample this call set's data was generated from.
+      Note: the current API does not have a rigorous definition of sample. Therefore, this
+      field actually contains an arbitrary string, typically corresponding to the sampleId
+      field in the read groups used to generate this call set.
+  :type sampleId: null|string
+  :field variantSetIds:
+    The IDs of the variant sets this call set has calls in.
+  :type variantSetIds: array<string>
+  :field created:
+    The date this call set was created in milliseconds from the epoch.
```
mbaudis: Time formats should all be ISO8601. 
Attributes should be ```recordCreateTime``` and ```recordUpdateTime```, and the type _string_.

See https://github.com/ga4gh/schemas/blob/metadata/doc/source/metadata_intro.rst for reference/explanation.

These formats are used below (Experiment, Analysis ...).

This change should be performed throughout the schema.

doc/source/schemas/alleleAnnotationmethods.rst
```diff
@@ -0,0 +1,700 @@
...
+  :field id:
+    The ID of this VariantAnnotation.
+  :type id: string
+  :field variantId:
+    The variant ID.
+  :type variantId: string
+  :field variantAnnotationSetId:
+    The ID of the variant annotation set this record belongs to.
+  :type variantAnnotationSetId: string
+  :field created:
```
mbaudis: ```recordCreateTime```,  ```recordUpdateTime```,  type ```string```, format ISO8601

doc/source/schemas/alleleAnnotations.rst
```diff
@@ -0,0 +1,562 @@
...
+      field actually contains an arbitrary string, typically corresponding to the sampleId
+      field in the read groups used to generate this call set.
+  :type sampleId: null|string
+  :field variantSetIds:
+    The IDs of the variant sets this call set has calls in.
+  :type variantSetIds: array<string>
+  :field created:
+    The date this call set was created in milliseconds from the epoch.
```
mbaudis: ```recordCreateTime```,  ```recordUpdateTime```,  type ```string```, format ISO8601

doc/source/schemas/alleleAnnotations.rst
```diff
@@ -0,0 +1,562 @@
...
+    The variant ID.
+  :type id: string
+  :field variantSetId:
+    The ID of the `VariantSet` this variant belongs to. This transitively defines
+      the `ReferenceSet` against which the `Variant` is to be interpreted.
+  :type variantSetId: string
+  :field names:
+    Names for the variant, for example a RefSNP ID.
+  :type names: array<string>
+  :field created:
```
mbaudis: ```recordCreateTime```,  ```recordUpdateTime```,  type ```string```, format ISO8601

doc/source/schemas/alleleAnnotations.rst
```diff
@@ -0,0 +1,562 @@
...
+  :field id:
+    The ID of this VariantAnnotation.
+  :type id: string
+  :field variantId:
+    The variant ID.
+  :type variantId: string
+  :field variantAnnotationSetId:
+    The ID of the variant annotation set this record belongs to.
+  :type variantAnnotationSetId: string
+  :field created:
```
mbaudis: ```recordCreateTime```,  ```recordUpdateTime```,  type ```string```, format ISO8601

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,194 @@
...
+  /** The ID of this VariantAnnotation. */
+  string id;
+
+  /** The variant ID. */
+  string variantId;
+
+  /** The ID of the variant annotation set this record belongs to. */
+  string variantAnnotationSetId;
+
+  /** The date this annotation was created in milliseconds from the epoch. */
```
mbaudis: ```recordCreateTime```,  ```recordUpdateTime```,  type ```string```, format ISO8601

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: Analysis is 1:1 with a VariantAnnotationSet, but it is a generic pre-existing model we are re-using here, which is why it has it's own id.

src/main/resources/avro/metadata.avdl
sarahhunt: @calbach - we took this from the Analysis record in master when we started to create the Variant Annotation API. It's currently unchanged, but it makes sense to change the comment to the one in metadata_intro.rst.
@mbaudis - we can't move to the new structure in the metadata branch as there often won't be information available for the required fields - for example individualIds, biosampleIds. At this stage, it would be simplest to use the old record the switch to something similarly light weight later when the Analysis model is updated. There is already the plan to look at Program and move this out of reads, so this may be a good way to go.

src/main/resources/avro/alleleAnnotations.avdl
jeromekelleher: I think Analysis needs an ID here @calbach because there are references to it in the AnalysisResult.

src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -0,0 +1,200 @@
...
+  /**
+  This filter allows variant, transcript combinations to be extracted by effect
+  type(s).
+  Only return variant annotations including any of these effects and only return
+  transcript effects including any of these effects. Exact matching across all
+  fields of the Sequence Ontology OntologyTerm is required.
+  (A transcript effect may have multiple SO effects which will all be reported.)
+  If null, return all variant annotations.
+  */
+  union { null, array<org.ga4gh.models.OntologyTerm> } effects = null;
```
macieksmuga: @sarahhunt, @diekhans Passing in whole JSON objects representing OntologyTerm into a query seems a bit strange. In Sequence Annotations, I replaced this with an `array<string> ontologyTerms = [];` so that a search could just be performed on the (presumably standardized) terms. 

I don't know which approach is correct, but I believe we should standardize on a single way of searching ontologies across GA4GH API's.

diekhans: Right now, only the id is required for OntologyTerm.

If this is hard to implement, I would favor delaying
implementation.  I am no sure passing in OntologyTerm is right,
however passing in generic strings isn't right;

I am pushing the metadata task team on refining OntologyTerm; if
you say "we don't know how to implement something", that is
*really* helpful.  Please create tickets and add MetadataTaskTeam
label.

mbaudis: The ```id``` is the only attribute in ontologies which definitely resolves. We could require the ```term``` attribute, too; but IMHO it makes more sense to code/document on the server side (e.g. resolving effects to their SO identifiers, while presenting the corresponding terms).

macieksmuga: Sounds like I should modify this accordingly in Sequence Annotation branch as well.

pcingola:
+1 time to merge?

jeromekelleher:
I'm +1 also. The server [implementation requirements](https://github.com/ga4gh/server/issues?utf8=%E2%9C%93&q=milestone%3Avariation_annotation_implementation+) have been met, so the schemas have been proved to be implementable. @david4096, can you comment on the compliance tests please?

One minor query that I have is about ISO formats for times. We should commit to a complete changeover to ISO format for times in the API before we release the next version (it would be confusing to have a mixture, where variation annotation part uses ISO 8601 and the rest uses unix time). Is there a plan to make this conversion with server and compliance tests implementations?

diekhans:
Yes, we are planning to do a separate PR to convert the other
times and make the field names consistent.

jeromekelleher:
> Yes, we are planning to do a separate PR to convert the other times and make the field names consistent.

Great. Definitely +1 on this PR then!

david4096:
@jeromekelleher Compliance branch is also ready to be merged (thx @sarahhunt !).

https://github.com/ga4gh/compliance/pull/141

diekhans:
+1  lets move forward on this

mbaudis:
+1

diekhans:
@jeromekelleher this has more than enough +1s, do you want to do the merger?

jeromekelleher:
@calbach, have you any thoughts here?

jeromekelleher:
@calbach, as one of the people who reviewed this it would be good to get your final opinion on it before we merge. However, the clock is ticking, and I think we should merge this quickly.

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,198 @@
...
+/**
+A HGVSAnnotation record holds Human Genome Variation Society descriptions
+of the sequence change with respect to genomic, transcript and protein
+sequences. See: http://www.hgvs.org/mutnomen/recs.html.
+Descriptions should be provided at genomic level. Descriptions at transcript
+level should be provided when the allele lies within a transcript. Descriptions
+at protein level should be provided when the allele lies within the translated
+sequence or stop codon.
+*/
+record HGVSAnnotation {
```
reece: Do we have a sense of how these will be used and whether they will be searched?

For example, how do we handle ENST v. RefSeq transcripts? One answer might be that we're agnostic, but then searching will be hampered which kinda defeats a core tenet of GA4GH.

In other words, is the representation here unambiguous, or might it reasonably depend on an end-user preference? If the latter, it might be better to let end-user software deal with the transformations.

diekhans: GA4GH is intended to be providing an API that is data-set agnostic, so
both RefSeq and Ensembl should be equal valid to include in a data
set.

Of course, being able to resolve annotations against multiple
versions of dataset would be ideal.

jeromekelleher: @reece, is this merge-blocker? If not we can open an issue to clarify the documentation on this point.

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+Impact is a simple prioritization for the effect of an allele which is used
+in the annotation record.
+IMPORTANT:
+ Prioritization methods are a crude estimates and are not assumed to be
+ reliable: a 'HIGH' Impact may actually not cause any disruption
+ in protein function or expression.
+*/
+enum Impact {
```
reece: We have not yet decided whether to include Impact. The issue is discussed in #516 and will be discussed in the VATT call on March 9th. In the meantime, let's remove impact now to unstick protobuf and compliance work. We can reconsider at a later date.

src/main/resources/avro/alleleAnnotations.avdl
reece: Not a merge blocker. I'm sure that this issue will come up naturally during use. Let's not create an issue now; instead, we can wait the use case/problem report to motivate our response.

src/main/resources/avro/metadata.avdl
```diff
@@ -9,6 +9,41 @@ protocol Metadata {
 import idl "common.avdl";
 
 /**
```
reece: My instinct is that OntologyTerm it should go either in alleleAnnotations.avdl or perhaps in a separate file.

diekhans: OntologyTerm is will be used by many different modules, its a
core part of metadata.  metadata.advl is going to become types
to support metadata; with the functional area metedata being
part of the appropriate modules.

PR #525, opened 2016-01-07
Variant annotation - update AnalysisResult

sarahhunt:
hold analysisId in AnalysisResult not the full record


jeromekelleher:
+1

PR #528, opened 2016-01-07
Create a metadata methods file and move dataset operations to it

### LINKED ISSUE #527: Dataset methods are described in readmethods.avdl
Datasets have a base endpoint described in readmethods yet the Dataset record itself is within metadata.

https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/metadata.avdl#L95

https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/readmethods.avdl#L182

david4096:
Dataset methods were being declared in readmethods. This PR moves dataset methods to a new file called metadatamethods.avdl.

Notice that fragmentID has been removed from the readmethods.rst. When making changes to the schema, the docs need to be updated as well and it appears that since fragment was removed this update hadn't been performed. Fixes #527


diekhans:
Does the server and compliance suite processing for the schemas still work?

david4096:
Here is the PR that would bring compliance in line with these changes. https://github.com/ga4gh/compliance/pull/159

diekhans:
+1  thanks for noticing this 1/2 done work

jeromekelleher:
+1

sarahhunt:
+1

david4096:
As far as server goes, I think these do not make any code changes. I ran the new schemas through the `process_schemas.py` and the `_protocol_definitions.py` files were identical ! @jeromekelleher sound about right?

jeromekelleher:
Good, thanks for verifying this @david4096. As we're only moving definitions around within files this shouldn't affect the schema definitions in any way.

Since we have the requisite +1s, the required time has elapsed and this is a minor change, I'm going to merge.

PR #531, opened 2016-01-11
reformat HGVS records

sarahhunt:


src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -107,6 +107,22 @@ record VariantAnnotationSet {
 }
 
 /**
+A HGVSAnnotation record holds Human Genome Variation Society descriptions
+of the variant allele with respect to a transcript.
+*/
```
jeromekelleher: A stupid question: are there any requirements about which of these fields can be null at the same time? I.e., can we have genomic, coding, or protein set, in any combination?

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -107,6 +107,22 @@ record VariantAnnotationSet {
...
+record HGVSAnnotation {
+
+  /** HGVS formatted annotation at genomic level  */
```
jeromekelleher: I'm afraid I don't find these comments hugely enlightening. I already know that it's a "HGVS formatted annotation" since it's held in this struct. Would it be possible to get a slightly higher-level description? Perhaps some links to the HGVS details?

sarahhunt:
@jeromekelleher - I see what you mean - I've removed the redundant comments. It would be cumbersome to describe HGVS in great detail here, so I've added a link to the recomendation. What do you reckon?

jeromekelleher:
Looks good to me @sarahhunt, +1

diekhans:
+1 on the hgvs change.  Linking to canonical external
documentation, when it exists, is far better that writing
a condensed `reader's digest' version

PR #533, opened 2016-01-12
Adds variant annotation schema diagram

andrewjesaitis:
This adds the variant schema diagram to the documentation. Additionally, I added allele annotations to the toctree so that you are able to navigate to the page from the API overview.

@reece In terms of versioning the diagram I was thinking that it would probably be most maintainable if these diagrams/documentation updates get checked in with the task team's overall merge. That way we leave the versioning to git and as long as the update goes with the overall merge it should be clear what version it is on. Also, it would seem to me that the checksum that should be referenced in the svg is the merge hash going into master, which doesn't exist yet. Alternatively, I can make this PR against master directly after the variant annotation PR is accepted. It just seemed cleaner to keep the documentation updates with the schema updates. Happy to do it either way.


david4096:
+10 for diagrams!

sarahhunt:
Yes - this is great @andrewjesaitis! 

I agree - adding this to the branch and merging everything to master together sounds a clean approach.

There has been a minor update since you generated the diagram, but there is still the Ontology record to reconcile between the Sequence Annotation and VA PR, so an update may be premature.

andrewjesaitis:
Sounds good. I'll keep my eye on the main PR to master. And update the diagram once the dust settles.

sarahhunt:
@andrewjesaitis - it looks like the dust has settled, so do go ahead and update. Thanks!

andrewjesaitis:
@sarahhunt Just updated the diagram. Thanks for letting me know!

sarahhunt:
Thanks @andrewjesaitis !

PR #535, opened 2016-01-13
Add a method for finding variant annotations by ID

### LINKED ISSUE #534: Get Variant Annotation by ID
There is a method for finding Variant Annotation Sets by their ID, but not Variant Annotations themselves.

https://github.com/ga4gh/schemas/blob/variation_annotation/src/main/resources/avro/alleleAnnotationmethods.avdl#L194

david4096:
Currently can get Variant Annotation Sets by ID, but not the variant annotations themselves. fixes #534


david4096:
Mistakenly done against master...

PR #538, opened 2016-01-15
clarify comments

sarahhunt:
@calbach - are these comments clearer?

It would be good to do searches including child terms in the ontology later, but it may be better to add a flag to trigger this behaviour.

I'd make HGVS.genomic required, but I don't think the guidelines cover all possible variants (though they probably cover eveything the current Variant model handles).

### DIFF
--- a/src/main/resources/avro/alleleAnnotationmethods.avdl
+++ b/src/main/resources/avro/alleleAnnotationmethods.avdl
@@ -76,8 +76,8 @@ record SearchVariantAnnotationsRequest {
 //  union { null, array<string> } featureIds = null;
 
   /**
-  Only return variant annotations including these effects (SO terms).
-  If null, return all variant annotations.
+  Only return variant annotations including these effects (SO terms). Exact
+  matching is required. If null, return all variant annotations.
   */
   union { null, array<org.ga4gh.models.OntologyTerm> } effects = null;
 
--- a/src/main/resources/avro/alleleAnnotationmethods.avdl
+++ b/src/main/resources/avro/alleleAnnotationmethods.avdl
@@ -77,7 +77,8 @@ record SearchVariantAnnotationsRequest {
 
   /**
   Only return variant annotations including these effects (SO terms). Exact
-  matching is required. If null, return all variant annotations.
+  matching across all fields of the OntologyTerm is required. If null, return
+  all variant annotations.
   */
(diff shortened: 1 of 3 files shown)


src/main/resources/avro/alleleAnnotationmethods.avdl
```diff
@@ -76,8 +76,8 @@ record SearchVariantAnnotationsRequest {
 //  union { null, array<string> } featureIds = null;
 
   /**
-  Only return variant annotations including these effects (SO terms).
-  If null, return all variant annotations.
+  Only return variant annotations including these effects (SO terms). Exact
+  matching is required. If null, return all variant annotations.
```
calbach: This still seems a bit ambiguous to me. Maybe further clarify that an `OntologyTerm` matches if and only if all fields of the `OntologyTerm` match exactly (at least, that's my interpretation of what this means).

src/main/resources/avro/alleleAnnotations.avdl
```diff
@@ -110,6 +110,10 @@ record VariantAnnotationSet {
 A HGVSAnnotation record holds Human Genome Variation Society descriptions
 of the sequence change with respect to genomic, transcript and protein
 sequences. See: http://www.hgvs.org/mutnomen/recs.html.
+Descriptions should be provided at genomic level. Descriptions at coding level
```
reece: In the pre-ncRNA world, "coding variant" and "transcript variant" were essentially the same thing. That's no longer true: protein coding sequences are a subset of transcribed sequences. Although this distinction is routinely overlooked, it seems to me that the ga4gh interfaces should do the technically right thing by calling these "transcript variants" rather than "coding variants".  Specific suggestion:

`union { null, string } transcript = null;`
in lieu of
`union { null, string } coding = null;`

This means that transcript variant strings might contain c. or n. variants.

src/main/resources/avro/alleleAnnotationmethods.avdl
sarahhunt: Thanks @calbach - I've updated the comment.

src/main/resources/avro/alleleAnnotations.avdl
sarahhunt: Thanks @reece - I've made that change.

calbach:
+1, thanks.

PR #540, opened 2016-01-16
Beacon

maximilianh:
This is another suggestion for a beacon API. It replaces most fields with three key-value maps.


diekhans:
Suggest have the values in key-value maps as arrays of strings.
This removes the need to come up with magic syntax that then has
to be parsed for both complex values and multi-valued
attributes.

mbaudis:
@maximilianh @mcupak Can we remove this in favour of current Beacon development?

mcupak:
I don't think I have the permissions needed to close this, but whoever does, please do so. This is out of date.

mbaudis:
Closed in favour of current Beacon development efforts.

Issue #142, opened 2014-09-10
GA4GH APIs need to address scientific reproducibility (propose immutable datatypes)


benedictpaten:
Consider a researcher who writes a script against the GA4GH APIs, accesses data and publishes the results. The current APIs do not guarantee that subsequent researchers will get the same result when running the original script, therefore the published results are not assured to be reproducible. 

If GA4GH APIs are to really going change the way bioinformatics is done they need to facilitate the reproducibility of results. In order for results to be reproducible one needs to be able obtain the exactly the same data and associated metadata that were used in an experiment.  For the GA4GH APIs this means that every time a given data object is returned it is always the same. This means that APIs must present data as immutable.  Data objects are never modified, instead new derived versions are created.

Mark Diekhans, David Haussler and I think this is important to address and that it would be relatively straightforward to implement immutability into an update of the v0.5 API. What do people think?

delagoya:
Closing this issue for lack of PR or recent comments. It seems that #167 takes precedence for this issue.

awz:
@delagoya is the Containers and Workflows task-team working on this issue?  Maybe wait until they make progress before closing this issue?  It seems to be pretty important and have quite a lot of content that is referenced by #167.   Also pinging  @fnothaft and @tetron since they may feel happy to close this.

PR #542, opened 2016-01-22
Added a line so that identifiers are URL safe.

david4096:
Although the specific identifier scheme is left up to the implementation, identifiers should be URL safe. If not, GET requests could require URL encoding. For example, we should avoid URLs that look like `http://ga4gh/referencesets/aGczNw%3D%3D` in favor of characters that are neither reserved nor need URL encoding.

```
   * ids must not include characters that require URL encoding or
     characters that are reserved for URL schemes (?, $, =, @, etc.).
```

### DIFF
--- a/doc/source/api/apidesign_intro.rst
+++ b/doc/source/api/apidesign_intro.rst
@@ -24,6 +24,9 @@ Object Ids
     across server restarts -- once a user is given an ID for data
     stored on a server, the id remains valid for as long as the server
     is still storing that data
+    
+  * ids should not include characters that require URL encoding or
+    characters that are reserved for URL schemes (?, $, =, @, etc.).
 
 * Many objects, including most ‘container’ objects, also have a
   **name** field. The name is user-defined, isn’t programmatically


jeromekelleher:
+1.

jmarshall:
-1.  This is not a **must not** (in the [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) sense).  It's a "implementations may wish to take cognisance of", which is at most **should not**.

jeromekelleher:
You're right @jmarshall, "should not" is probably more appropriate. Just out of interest though, why not make it a "must not"? I'm trying to think of cases where a server would need to use URL-unsafe IDs and not really coming up with anything. It seems to me like well-formed clients will have to always check the IDs for URL-safeness before creating a GET request if we don't make this a requirement, and so we're making clients more complex if we make it a "should not" rather than a "must not".

diekhans:
If the goal is interoperability, its seems it should be a "must
not". Either that or recommend that ids always be URL encode.

I am +1 on making it a "must not" without an argument against it.

dcolligan:
I'm in favor of "must not"

calbach:
+1. Completely agree this should be a **must not**.

macieksmuga:
+1 If we want identifiers on URL's then they MUST be URL compliant. What am I missing here?

mbaudis:
+1 for must

Also, implies rule based id normalization/checking. There is a wild world out there ...

diekhans:
@macieksmuga if we don't define identifiers as being URL safe, then 
we are require people to URL encode them.

While for programmatic access, one should got through a function
to builds the URL, in reality people don't always do this,
especially for command line

pgrosu:
Most digests as these id's most likely will end up, can be `hexlify`'ed from their binary.  Here's a link that might help:

https://docs.python.org/2/library/hashlib.html

The reference for handling UUIDs stems from a previous discussion here - which got referenced and extended into other ones:

https://github.com/ga4gh/schemas/issues/142

Ideally most of these the processing should be handled among machines where the end-users just get the high-level results of the analysis.

Hope it helps,
~p

mbaudis:
@pgrosu That's what @diekhans was talking about. Any mechanism normalizing ids (and we are not talking UUIDs here, which is a different can of worms) would imply some tool specificity, or lead to ambiguous outcome; whereas a rejection-but-not-conversion mechanism seems safe.

Or one specifies a very clear mechanism, e.g. expressed as a regex based substitution. But there are certainly plenty of examples where this has been shown to fail.

pgrosu:
@mbaudis Nice, glad we're all in agreement :smile:  I only mentioned UUID since the initial post mentioned `referencesets` in the url (`http://ga4gh/referencesets/aGczNw%3D%3D`) and those methods usually are defined by an `id` that should stay fairly static across repositories, in order for them to "talk" the same standard.

jmarshall:
@jeromekelleher You are right that for the statement to have a practical effect for client writers, it would have to be **must not**.

I guess my point is that IMHO this is not the sort of detail that a grown up specification should be concerning itself with.  For me, it pretty trivially runs afoul of that RFC's [§6 guidance](https://www.ietf.org/rfc/rfc2119.txt).  One of the benefits of layering on top of existing standards like HTTP and URI is that we don't have to worry about request escaping and encoding issues, and there's an extent to which it IMHO behoves us not to.

Perhaps in future we'll add some free-text or similar request to the schema, one which could not reasonably be restricted to use only URI-unreseved characters.  And then clients will need to be doing percent-encoding anyway.  Oh well, if that happens we can come back and delete this line again. :smile:

sarahhunt:
An example of a request which cannot be restricted to use only URI-unreserved characters is a  GET request by HGVS variant description. It's only been been raised as a possiblity so far - there is no immediate plan to support it.

jeromekelleher:
@jmarshall, I think I must be missing something here. All we are specifying here is that ID strings must be URL safe. The use-case is, I have (say) a ``referenceId='2?x=1&y=2'`` embedded within a variant object and I want to use a HTTP GET request to find out some details about that reference. I construct a URL ``http://example.com/references/2?x=1&y=2``, which won't work. I therefore have to URL encode my referenceId value on the client side. Other (non-pathological) servers will return sensible, URL safe ID strings. Therefore, for interoperability, we must either specify that (a) all clients must URL encode identifiers before putting them into query strings (which I guess is implicit, and is where you are coming from?); or (b) we specify that IDs must be URL safe.

From an overall interoperability perspective it seems to me that requiring URL-safeness of ID strings is perfectly reasonable, and falls directly within the guidance of RFC 2119 section 6.

diekhans:
@sarahhunt  this only applies to the id attribute,

david4096:
@jmarshall can I call that a reserved +1? This is adhering to standards and won't limit the types of queries that can be performed.

Gory details here: http://tools.ietf.org/html/rfc3986#section-2

jmarshall:
> @jmarshall can I call that a reserved +1?

That would be an astonishing misrepresentation.

In future, we'll likely have requests such as the one Sarah described.  At that point, clients will have to percent-encode that request but are instructed that they can get away without percent-encoding this request.  This is needless specification complexity that helps no-one.  It is not _**actually required** for interoperation_ [my emphasis], to use RFC 2119 §6's phrase.

However in this instance the effect is fortunately minor and y'all seem inexplicably intent on it.  So you can call my position **-0 whatevs**.

jeromekelleher:
OK, @jmarshall does have a point here:  URLs should be URL encoded, or else they're not really URLs. The client is always responsible to constructing URLs correctly, and us requiring that one fragment of particular URLs is safe doesn't really help. So, is this pointless?

saupchurch:
I'm in agreement with @jmarshall and I think that @jeromekelleher has the right idea now.  In the world of web clients it is the responsibility of the client to encode the URL.  If it doesn't then it is not sending the server a well formed URL with the intended data and the server response should be an error of some sort.

david4096:
OK I've changed `must not` to `should not`.

```
 +  * ids should not include characters that require URL encoding or
 +    characters that are reserved for URL schemes (?, $, =, @, etc.).
```

david4096:
@jmarshall @jeromekelleher @saupchurch Switching must to should, can we +1 this?

jeromekelleher:
I'm still on the fence on this one to be honest. I'll give it a -0 --- I'm slightly negative about the need for it, but I won't block others on it.

david4096:
Closing due to lack of consensus. Odd characters in IDs will not be common practice.

diekhans:
reopening This needs to be resolved one way or another.    While everyone should URL encode arguments, the fact that people often don't means if should either be documented one way or another.

david4096:
@diekhans In practice you don't need to specify this. I was convinced after this conversation that implementors will have considered this and that it won't harm interoperability to leave it open. Since we use very simple paths throughout the API, an argument for tidy looking URLs becomes less convincing.

diekhans:
I have no idea why the `tidy looking URL' issue is.

This is about either making it easy for sloppy programmers or
warning them/

PR #543, opened 2016-01-23
Re-added sequenceAnnotations,

macieksmuga:
Including sequenceAnnotationMethods and OntologyTerm back into metadata. Also updated or regenerated documentation RST's accordingly as needed.


src/main/resources/avro/metadata.avdl
```diff
@@ -9,6 +9,32 @@ protocol Metadata {
 import idl "common.avdl";
 
 /**
+  An ontology term describing an attribute. (e.g. the phenotype attribute
+  'polydactyly' from HPO)
+  */
+  record OntologyTerm {
```
sarahhunt: @macieksmuga , @david4096 - I see this PR uses a different version of OntologyTerm to the Variant Annotation PR.

Here we have:
{"ontologySourceName":"Sequence Ontology",
"ontologySourceID":"http://www.sequenceontology.org/browser/release_2.5.3/term/SO:0001583",
"ontologySourceVersion":"release_2.5.3"}

and in #519 we have:
{"ontologySource":"Sequence Ontology",
"name":"missense_variant",
"id":"http://www.sequenceontology.org/browser/release_2.5.3/term/SO:0001583"}

The former is from the current metadata branch and the latter was in the  master branch last July.

@mbaudis - was it intentional to drop the name from the record? It seems a useful piece of information for the API to return.

mbaudis: @sarahhunt Good question; we're currently (i.e. last/this week) discussing the ```OntologyTerm``` structure, to stabilize & document it. That's an item where other people (probably) have strong opinions, and since it is related to many records we have to get this with a good consensus score. Pls. follow up on this!

sarahhunt: Thanks @mbaudis! Is there an issue? I would strongly prefer the name to be included. I would also suggest dropping the 'ontology' prefix from all attributes for consistency with the rest of the API.

Usig the asbove example, this would look something like:

{"sourceName":"Sequence Ontology",
"sourceVersion":"release_2.5.3"
"name":"missense_variant",
"id":"http://www.sequenceontology.org/browser/release_2.5.3/term/SO:0001583"}

mbaudis: Yes, I was just writing something along those lines ;)

diekhans: We pick up the new one from the metadata branch.  However, I name
is essential for making sequenceAnnotations useful (this is how
GFF3 works).

Lets add name back in and then work this out with the metadata TT.

mbaudis: @diekhans We seem to converge on "term" instead of "name"; but this should be a drop-in replacement.

macieksmuga: There's both a `metadata` and a `metadata-working` branch, each of which has its own version of `OntologyTerm`. Which one should I settle on?

diekhans: @mbaudis comments?  I think metadata-working is closer, but can
you tell us for sure.

mbaudis: I did some house cleaning ...
* ```metadata-working``` is gone, since the cleanup/development has been moved to https://github.com/mbaudis/schemas (e.g.https://github.com/mbaudis/schemas/pull/4)
* ```OntologyTerm``` reflects the current point of discussion; especially ```id``` and ```term``` seem to be stable.

macieksmuga:
Ok, I have several comments of my own on the new definitions in this PR, and unless I hear quickly otherwise, I'll implement the changes they imply:

1. There are no get/search methods for `FeatureSet`s. At best, it's inconsistent with the rest of the GA4GH API's, at worst, since I need to provide a `FeatureSetId` to a `SearchFeaturesRequest`, it makes the API unusable. I propose to add these in a way consistent with the get/search for `VariantSet`.  Doing so will also allow me to remove this sad little comment from the bottom of `sequenceAnnotationmethods.avdl`:
```
// TODO: AnnotationSet methods.
```

2. Search on references is done by `referenceName` in variants, and by `referenceId` in reads. Here, it's done optionally by both. I suggest removing the search by `referenceName` here:
```
    /**
    Only return features with on the reference with this name.  One of this
    field or `referenceId` is required.  (case-sensitive, exact match)
    */
    union { null, string } referenceName = null;
```

3. Until ontology trees are properly represented in GA4GH, I propose we resolve the following ambiguity:
```
    // TODO: Fix this field. Clarify semantics around how OntologyTerm
    // matching works, or change the Ontology term field on the feature.
    /**
    If specified, this query matches only annotations which match one of the
    provided feature types.
    */
    array<org.ga4gh.models.OntologyTerm> features = [];
```
By stipulating that only exact matches be returned for now. Also, this is part of a query, not a response, so let's do this instead:
```
  /**
  If specified, this query matches only annotations which match one of the
  provided ontology terms.
  */
  array<string> ontologyTerms = [];
```

4. In `sequenceAnnotations.avdl`, in the definition of the `Feature` record, add in the field:
```
  array<Feature> children = [];
```
This would be nonempty for discontinuous features such as CDS, and would list all the 
CDS regions in order. It would also signify that the range given in this feature shouldn't
be used for computing if a base is outside of this feature, as proper membership in the feature
is determined by examining the union of ranges of its children.

5. Also in the definition of the `Feature` record, add in `strand`.

diekhans:
Maciek Smuga-Otto <notifications@github.com> writes:
>  1. There are no get/search methods for FeatureSets.

Since the basic schema was needed quickly, many queries were
left undefined.  Please fill out and make consistent with the
rest of the API.


>  1. Search on references is done by referenceName in variants, and by
>     referenceId in reads. Here, it's done optionally by both. I suggest
>     removing the search by referenceName here:

I have no idea which is right; but id should at least be there,
so go for it.

>  1. Until ontology trees are properly represented in GA4GH, I propose we
>     resolve the following ambiguity:

Ontologies normally use lookup services as opposed to custom implementations,
so this will make the implementation straight forward.  For example:

   http://www.ebi.ac.uk/ontology-lookup/

> By stipulating that only exact matches be returned for now. Also, this is part
> of a query, not a response, so let's do this instead:

Good plan, however please include in the documentation the fact
that it is a short-term approach and will be replaced by a true
ontology lookup.

>   array<Feature> children = [];
> 
> This would be nonempty for discontinuous features such as CDS,
> and would list all the CDS regions in order. It would also
> signify that the range given in this feature shouldn't be used
> for computing if a base is outside of this feature, as proper
> membership in the feature is determined by examining the union
> of ranges of its children.

Please discuss this with me, this is a bit off.  Discontinuous
features are part of the tab separated format and we are better
of getting rid of the concept and doing it in a simpler way.

>  1. Also in the definition of the Feature record, add in strand.

This use to be there, something happened.

macieksmuga:
@diekhans @sarahhunt changes introduced as discussed above. Please let me know if it's ready for your +1's (and if not, why)

src/main/resources/avro/sequenceAnnotationmethods.avdl
```diff
@@ -0,0 +1,165 @@
...
+    join of circular genomes are represented as two requests one on each side
+    of the join (position 0).
+    */
+    long start;
+
+    /**
+    Required. The end of the window (0-based, exclusive) for which overlapping
+    features should be returned.
+    */
+    long end;
```
sarahhunt: Should referenceId be optional when start and end are both required? I would have thought they should have the same state.

Does it mean the sequence can be defined by referenceId or parentId then the start and end are applied? That would seem complex at this stage.

doc/source/api/sequence_annotations.rst
```diff
@@ -0,0 +1,36 @@
...
+************************
+Sequence Annotations API
+************************
+For the Sequence Annotation schema definitions, see `Sequence Annotation schema <../schemas/sequenceAnnotations.html>`_
+
+
+------------------------
+Feature Based Hierarchy
+------------------------
+The central object of the GA4GH Sequence Annotation API is a Feature.  The Feature describes an interval of interest on some reference(s).  It has a span from a start position to a stop position as well as descriptive data.  A Feature has one or more parent Features which enables the construction of more complex representations in a hierarchical way.
```
sarahhunt: 'A Feature has one or more parent..'  => 'A Feature can have one or more parent..' ?
Some features are top level.

doc/source/api/sequence_annotations.rst
```diff
@@ -0,0 +1,36 @@
...
+
+---------------------------
+The Sequence Annotation Schema
+---------------------------
+TODO: insert an example annotation translation from GFF3 to GA4GH
+
+---------------------------------------
+Annotation Design - RNA Considerations
+---------------------------------------
+
```
sarahhunt: Is this comment -

'The annotation API needs to be able to report assignment at the feature level as well as aggregate assignment at the transcript or even the whole gene level if assignment is not more specific than that.'

relevant here? It seems more suited to the RNA docs

sarahhunt:
Hi @macieksmuga - looks good! I've added a few minor comments. +1 when referenceId clarified.

src/main/resources/avro/sequenceAnnotationmethods.avdl
macieksmuga: Good catch, @sarahhunt! Furthermore, with our eventual transition to protobuf, we won't have the luxury of nullable fields, which makes this a no-brainer to remove the union with null here.

doc/source/api/sequence_annotations.rst
macieksmuga: Whoops, that crept in by accident. Should have zero or one parents in current schema. Thanks!

doc/source/api/sequence_annotations.rst
macieksmuga: Good point. @saupchurch - do you have any objections to me removing this from the Sequence Annotation branch?

saupchurch: no objections - it is derived from RNA considerations

diekhans: As long as the RNA considerations is only conceptual and doesn;t
reference APIs that are not yet there, it's good to leave it it.

We need to have documented rationale.

macieksmuga:
@diekhans I added back in the RNA considerations section at your request. Please take a look to see if my writeup of the justification for single parent (as opposed to array of parentIds) makes sense to you.

diekhans:
@macieksmuga, this is clear and too the point! thank you.

Some of the work CH is doing to make sequence annotations more
type safe will provide some drawings that will help explain
this.

macieksmuga:
@saupchurch @diekhans @mbaudis I made one (hopefully minor) change to this PR, replacing the `attributes` metadata field in `FeatureSet` with a more generic `info` dictionary. This reflects the changing role of `FeatureSet` from an entity that was meant to represent certain high-level features (such as genes) to a representation of a whole GFF3 file's worth of data.

One other thing I _didn't_ propose in this PR, but would love it if you all could voice an opinion on, is the name `Feature` itself: It would make sense to me to rename it `SequenceAnnotation`, and rename `FeatureSet` to `SequenceAnnotationSet`. This renaming would replace a very generic word with a much more descriptive term, and would further make the naming scheme of this section of GA4GH consistent with the other parts (variants, reads, references). Thoughts?

diekhans:
There is some confusion here.  FeatureSet was always intended to
to be a full collection of annotations, as represented by a GFF3
file, not a single feature tree.

The use of the attributes type here is intentional, with the
reasons for it's use being what is documented.  Don't create a
second level of parsing.

I would like to make this general throughout GA4GH, the current
map of strings is not robust.  People end up making parsing
rules for different attribute names.

This would be an ideal use of polymorphism, sadly not part of
protobuf or avor.

pgrosu:
@diekhans I think protobuf could provide polymorphism, as follows:

```Protobuf
message Animal
{
    extensions 100 to max;

    enum Type
    {
        Cat = 1;
        Dog = 2;
    }

    required Type type = 1;
}

message Cat
{
    extend Animal
    {
        required Cat animal = 100; // Unique Animal extension number
    }

    // These fields can use the full number range.
    optional bool fluffy = 1;
}

message Dog
{
    extend Animal
    {
        required Dog animal = 101; // Unique Animal extension number
    }

    // These fields can use the full number range.
    optional uint32 playful = 1;
}
```

Hope it helps,
~p

diekhans:
Thanks @pgrosu ; this might be a useful approach (in V3 it is called
`any'.

Although a fixed set of derived types might be good enough.

diekhans:
Feature was chosen because this is the terminology from the existing annotation
world.

Note that `annotation' caused confusion and many showed up at the
annotation task team meeting thinking it was about sequence annotation
rather than variant annotation.

Features (discrete annotations) are not the only type supported, there are
also wiggle (continuous annotations), so the naming breakdown makes sense.

Changing Feature to SequenceFeature would make it clearer.

macieksmuga:
Thanks @diekhans for your feedback. I personally like SequenceFeature as well, and will try a version of this PR with it soon. I take you're effectively +1'ing the actual diff I proposed above (attributes -> info)?

pgrosu:
@diekhans Sure, but in proto3 the runtime libraries for `Any` are still under development, as noted here:

https://developers.google.com/protocol-buffers/docs/proto3#any

In any case - no pun intended :) - if folks want to play with the example I listed above, below are the steps:

1) Below are the contents of the `animals.proto` file:

```Bash
$ cat animals.proto
// [START declaration]
syntax = "proto2";
// [END declaration]

// [START messages]
message Animal
{
    extensions 100 to max;

    enum Type
    {
        Cat = 1;
        Dog = 2;
    }

    required Type type = 1;
}

message Cat
{
    extend Animal
    {
        optional Cat animal = 100; // Unique Animal extension number
    }

    // These fields can use the full number range.
    optional bool fluffy = 1;
}

message Dog
{
    extend Animal
    {
        optional Dog animal = 101; // Unique Animal extension number
    }

    // These fields can use the full number range.
    optional uint32 playful = 1;
}
// [END messages]
```
2) Next I produce the Python code, as follows:

```Bash
$ protoc --proto_path=. --python_out=. animals.proto
```
3) This will produce a file called `animals_pb2.py`, which looks like this:

```Python
$ cat animals_pb2.py
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# source: animals.proto

import sys
_b=sys.version_info[0]<3 and (lambda x:x) or (lambda x:x.encode('latin1'))
from google.protobuf import descriptor as _descriptor
from google.protobuf import message as _message
from google.protobuf import reflection as _reflection
from google.protobuf import symbol_database as _symbol_database
from google.protobuf import descriptor_pb2
# @@protoc_insertion_point(imports)

_sym_db = _symbol_database.Default()




DESCRIPTOR = _descriptor.FileDescriptor(
  name='animals.proto',
  package='',
  syntax='proto2',
  serialized_pb=_b('\n\ranimals.proto\"H\n\x06\x41nimal\x12\x1a\n\x04type\x18\x01 \x02(\x0e\x32\x0c.Animal.Type\"\x18\n\x04Type\x12\x07\n\x03\x43\x61t\x10\x01\x12\x07\n\x03\x44og\x10\x02*\x08\x08\x64\x10\x80\x80\x80\x80\x02\"4\n\x03\x43\x61t\x12\x0e\n\x06\x66luffy\x18\x01 \x01(\x08\x32\x1d\n\x06\x61nimal\x12\x07.Animal\x18\x64 \x01(\x0b\x32\x04.Cat\"5\n\x03\x44og\x12\x0f\n\x07playful\x18\x01 \x01(\r2\x1d\n\x06\x61nimal\x12\x07.Animal\x18\x65 \x01(\x0b\x32\x04.Dog')
)
_sym_db.RegisterFileDescriptor(DESCRIPTOR)



_ANIMAL_TYPE = _descriptor.EnumDescriptor(
  name='Type',
  full_name='Animal.Type',
  filename=None,
  file=DESCRIPTOR,
  values=[
    _descriptor.EnumValueDescriptor(
      name='Cat', index=0, number=1,
      options=None,
      type=None),
    _descriptor.EnumValueDescriptor(
      name='Dog', index=1, number=2,
      options=None,
      type=None),
  ],
  containing_type=None,
  options=None,
  serialized_start=55,
  serialized_end=79,
)
_sym_db.RegisterEnumDescriptor(_ANIMAL_TYPE)


_ANIMAL = _descriptor.Descriptor(
  name='Animal',
  full_name='Animal',
  filename=None,
  file=DESCRIPTOR,
  containing_type=None,
  fields=[
    _descriptor.FieldDescriptor(
      name='type', full_name='Animal.type', index=0,
      number=1, type=14, cpp_type=8, label=2,
      has_default_value=False, default_value=1,
      message_type=None, enum_type=None, containing_type=None,
      is_extension=False, extension_scope=None,
      options=None),
  ],
  extensions=[
  ],
  nested_types=[],
  enum_types=[
    _ANIMAL_TYPE,
  ],
  options=None,
  is_extendable=True,
  syntax='proto2',
  extension_ranges=[(100, 536870912), ],
  oneofs=[
  ],
  serialized_start=17,
  serialized_end=89,
)


_CAT = _descriptor.Descriptor(
  name='Cat',
  full_name='Cat',
  filename=None,
  file=DESCRIPTOR,
  containing_type=None,
  fields=[
    _descriptor.FieldDescriptor(
      name='fluffy', full_name='Cat.fluffy', index=0,
      number=1, type=8, cpp_type=7, label=1,
      has_default_value=False, default_value=False,
      message_type=None, enum_type=None, containing_type=None,
      is_extension=False, extension_scope=None,
      options=None),
  ],
  extensions=[
    _descriptor.FieldDescriptor(
      name='animal', full_name='Cat.animal', index=0,
      number=100, type=11, cpp_type=10, label=1,
      has_default_value=False, default_value=None,
      message_type=None, enum_type=None, containing_type=None,
      is_extension=True, extension_scope=None,
      options=None),
  ],
  nested_types=[],
  enum_types=[
  ],
  options=None,
  is_extendable=False,
  syntax='proto2',
  extension_ranges=[],
  oneofs=[
  ],
  serialized_start=91,
  serialized_end=143,
)


_DOG = _descriptor.Descriptor(
  name='Dog',
  full_name='Dog',
  filename=None,
  file=DESCRIPTOR,
  containing_type=None,
  fields=[
    _descriptor.FieldDescriptor(
      name='playful', full_name='Dog.playful', index=0,
      number=1, type=13, cpp_type=3, label=1,
      has_default_value=False, default_value=0,
      message_type=None, enum_type=None, containing_type=None,
      is_extension=False, extension_scope=None,
      options=None),
  ],
  extensions=[
    _descriptor.FieldDescriptor(
      name='animal', full_name='Dog.animal', index=0,
      number=101, type=11, cpp_type=10, label=1,
      has_default_value=False, default_value=None,
      message_type=None, enum_type=None, containing_type=None,
      is_extension=True, extension_scope=None,
      options=None),
  ],
  nested_types=[],
  enum_types=[
  ],
  options=None,
  is_extendable=False,
  syntax='proto2',
  extension_ranges=[],
  oneofs=[
  ],
  serialized_start=145,
  serialized_end=198,
)

_ANIMAL.fields_by_name['type'].enum_type = _ANIMAL_TYPE
_ANIMAL_TYPE.containing_type = _ANIMAL
DESCRIPTOR.message_types_by_name['Animal'] = _ANIMAL
DESCRIPTOR.message_types_by_name['Cat'] = _CAT
DESCRIPTOR.message_types_by_name['Dog'] = _DOG

Animal = _reflection.GeneratedProtocolMessageType('Animal', (_message.Message,), dict(
  DESCRIPTOR = _ANIMAL,
  __module__ = 'animals_pb2'
  # @@protoc_insertion_point(class_scope:Animal)
  ))
_sym_db.RegisterMessage(Animal)

Cat = _reflection.GeneratedProtocolMessageType('Cat', (_message.Message,), dict(
  DESCRIPTOR = _CAT,
  __module__ = 'animals_pb2'
  # @@protoc_insertion_point(class_scope:Cat)
  ))
_sym_db.RegisterMessage(Cat)

Dog = _reflection.GeneratedProtocolMessageType('Dog', (_message.Message,), dict(
  DESCRIPTOR = _DOG,
  __module__ = 'animals_pb2'
  # @@protoc_insertion_point(class_scope:Dog)
  ))
_sym_db.RegisterMessage(Dog)

_CAT.extensions_by_name['animal'].message_type = _CAT
Animal.RegisterExtension(_CAT.extensions_by_name['animal'])
_DOG.extensions_by_name['animal'].message_type = _DOG
Animal.RegisterExtension(_DOG.extensions_by_name['animal'])

# @@protoc_insertion_point(module_scope)
$
```

The Language Guide for proto2 is at the following link, for anyone who might be interested:

https://developers.google.com/protocol-buffers/docs/proto

Hope it helps,
~p

david4096:
+1

david4096:
Since this is a feature branch can we merge this? +1

diekhans:
+1 yes, merge into feature branch

PR #546, opened 2016-01-27
clarify documentation of effect filtering

### LINKED ISSUE #545: Transcript effect name search: are arrays logical OR or AND?
When making a request for variation annotations an array of effect names can be sent:

    /**
    Only return variant annotations including these effects (SO terms).
    If null, return all variant annotations.
    */
    union { null, array<org.ga4gh.models.OntologyTerm> } effects = null;

https://github.com/ga4gh/schemas/blob/variation_annotation/src/main/resources/avro/alleleAnnotationmethods.avdl#L79

A single variation can have a number of transcript effects. Does one expect only annotations that have all of the specified effects are returned? Or are all variants with any of the requested effects returned?

In addition, transcript effects are part of the body of an annotation. When a search for a given effect is made but the variant has multiple transcript effects, are only the effects that match the search expected in the response? Or is it safe to assume all transcript effects shall be included with the annotation, regardless of the search terms?

sarahhunt:
To fix #545

@david4096 - what do you reckon; is this clearer?

### DIFF
--- a/src/main/resources/avro/alleleAnnotationmethods.avdl
+++ b/src/main/resources/avro/alleleAnnotationmethods.avdl
@@ -76,9 +76,13 @@ record SearchVariantAnnotationsRequest {
 //  union { null, array<string> } featureIds = null;
 
   /**
-  Only return variant annotations including these effects (SO terms). Exact
-  matching across all fields of the OntologyTerm is required. If null, return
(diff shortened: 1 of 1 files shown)


david4096:
Yes, thanks +1

PR #547, opened 2016-01-28
Convert avdl -> proto3

calbach:
Opened in favor of #485.

As alluded to in several GA4GH discussions and [issues](https://github.com/ga4gh/schemas/issues/343), this PR converts the GA4GH schemas from Avro to [protocol buffers](https://developers.google.com/protocol-buffers/docs/overview). This is ready to be discussed and reviewed, but a prerequisite for merging will be an agreed plan for corresponding changes to the reference server and compliance suite (in the works, but need an explicit plan for sequencing the PRs).

I'll highlight a few factors which motivate this proposal:
- Protocol buffers have defined JSON serialization, which was sorely missing in our use of Avro. Libraries for protobuf <-> JSON conversion are defined at http://github.com/google/protobuf. See also the [specification](https://developers.google.com/protocol-buffers/docs/proto3#json).
- Gives the GA4GH a straightforward pathway to support an optimized binary wire format in the future.
- (Slightly subjective) Generally more community adoption / tooling / client library support.

At a high level, my aim with this PR was to make the minimal possible functional / documentation / wire JSON / code structure change while still winding up with valid proto3. I have a list of natural follow-up changes given proto as the schema - but I think these make more sense as separate PRs.

That being said, there are some changes from Avro -> proto3 which are unavoidable:
- proto3 does not have optional/default semantics, e.g. Avro's `union {string, null} foo = ""`
  - In places, this may change the JSON wire format slightly - e.g. null -> "" or vice versa.
  - As far as I can tell, the only place this really causes a problem is for `SearchVariantsRequest.callsetIds`, which will be resolved separately by https://github.com/ga4gh/schemas/pull/488.
  - Technically you can still get "nullable" semantics by using a wrapper message, at the cost of ugliness and stutter in the API.
  - If you're using the generated proto libraries, this shouldn't make a difference.
- int64s now encoded as strings, per the [protobuf JSON spec](https://developers.google.com/protocol-buffers/docs/proto3#json).
  - In Avro, these were presumably encoded as a JSON `Number`, though this was never strictly defined AFAICT so perhaps string was also valid.
  - See [this thread](https://github.com/ga4gh/compliance/pull/166#issuecomment-188469430) for additional context.
- In proto3, enum 0-values should be a sane 'unset' default.
  - The reason behind this is that the 0-value is the default when you initialize the request proto, or if a client sends you nothing. In general, it is better to disambiguate this case.
  - For this reason, added `${ENUM_NAME}_UNSPECIFIED` to all enums. This makes sense in most places, though feels the least natural in Cigar Operation.
- map\<[]string\> becomes map\<google.protobuf.ListValue\>.
  - This is the best way of generating consistent JSON with the previous approaches, but the ListValue message is actually more flexible than []string (elements could be number, message, etc).
  - For now, the requirement of a `[]string` payload is enforced only by the comment.
  - An alternative would be to relax this documentation to allow all values of `ListValue` or to use the more freeform google.protobuf.Struct, which represents arbitrary JSON.
  - I don't have a strong opinion here, again this first pass is what I believe to be the minimal functional delta.
- JSON wire format change: "sourceURI" -> "sourceUri"
- I converted all 'methods' into proto RPC service methods.
  - For now, this would be for documentation purposes only, much as the Avro methods are.
  - In order to support the GET operations, I've added small ‘request’ protos (with a single ID) for those endpoints, in addition to a referenceId on the ListBasesRequest.
- All fields now have [tag numbers](https://developers.google.com/protocol-buffers/docs/proto3#assigning-tags).
- In a few places, nested message definitions to make the hierarchy clearer e.g. CigarUnit { Operation {} operation }. This mainly aesthetic, though it also may impact the generated libraries.

Also, for reference, the Python generated classes can be created by the following manual process:
1. Install protoc >3.0 from https://github.com/google/protobuf/releases
2. `$ cd src/main/proto && protoc --python_out=. ga4gh/*`

TODO
- Per comments on ga4gh/server#879, the ListValue fix has been [integrated externally](https://github.com/google/protobuf/pull/1195) and we're just awaiting a protobuf release.


jeromekelleher:
This looks great @calbach. I've updated the server to use these proto definitions in https://github.com/ga4gh/server/pull/891 and it's working well. The only issue I'm having is the lack of the GAException object, which we either need to put back in (see #548) or replace with the upstream proto equivalent. I don't really mind.

msackman:
Please bump to protobuf 3.0.0 beta2 asap. The reason is as follows:

In beta1, the field name conversion is given by this function:
https://github.com/google/protobuf/blob/v3.0.0-beta-1/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java#L920
Crucially, this includes:

      } else if (Character.isDigit(ch)) {
        result.append(ch);
        isNextUpperCase = true;

So `md5checksum` becomes `md5Checksum`. However, it would appear this is a mistake: in python beta 1.1 https://github.com/google/protobuf/blob/v3.0.0-beta-1.1/python/google/protobuf/descriptor.py#L849 there is no such detection of digit so `md5checksum` stays as is. This means the compliance suite (in Java) fails to match up fields with the python server.

By beta2, both the Java and Python versions match behaviour and both do _not_ capitalise after a digit:
Java: https://github.com/google/protobuf/blob/v3.0.0-beta-2/java/src/main/java/com/google/protobuf/Descriptors.java#L1166 has now become:

        } else if (Character.isDigit(ch)) {
          result.append(ch);
          isNextUpperCase = false;

Python: https://github.com/google/protobuf/blob/v3.0.0-beta-2/python/google/protobuf/descriptor.py#L849

calbach:
Thanks for the detailed report. I've bumped the version and confirmed that `mvn clean install` is still working.

jjfarrell:
What is the convention for  handling missing values in protobuf3? For
example, in the call message, genotype is int32. How is a missing genotype
indicated?

repeated int32 genotype = 3;

​John​

calbach:
To your specific question about `genotype` which is an array of int32, it's not documented but I would guess that a missing genotype should be represented via -1. Avro had the same issue and same lack of documentation so I won't aim to change that here.

The general convention is that missing values should have identical semantics to 0-values. I believe this follows naturally for all fields in this PR, but if you feel otherwise please comment inline on a specific field.

calbach:
I added a note in the PR description of an additional wire incompatibility with this PR which I'd not made note of previously (though I was aware of the encoding): int64s are encoded as strings in proto3 JSON.

My understanding is that the reason for this encoding in protobuf JSON is that the Javascript Number representation is a 64 bit float, so it can only precisely represent integer values up to 2<sup>52</sup>. This was never rigorously specified in Avro, but presumably these 'long' fields would have been encoded as numbers. We have the option of using a int32 or uint32 instead, but the maximum values may be too low for various use cases, e.g. genomic coordinate. For now I'm leaving as int64 which I think is more in the spirit of the original Avro, even if it does lose wire compatibility.

It should also be noted that a server should accept both number and string, but should produce strings for fields of these types.

More discussion [here](https://github.com/ga4gh/compliance/pull/166).

calbach:
There are now many lines of code that I will need to merge in from the schemas master branch, e.g. variant annotations. My current understanding is that (1) this sync is not actively blocking reference server development, but will need to happen eventually and (2) this sync will not disrupt progress if it happens too soon, i.e. no one would complain if I updated this PR today to include variant annotations. I  also expect that I'm the one who will be doing this work, as I am the one proposing this PR. Based on this understanding, I expect to get around to updating the PR sometime next week.

@jeromekelleher @david4096 let me know if my timeline/expectations are misaligned with your own.

diekhans:
Update any time.  There are other schema changes coming
for sequence annotations and metadata.

It would be great to move the cigar stuff out of common.* as
long as the schema is changing.  It's not `common', it's only
used by reads.

diekhans:
Also, this is not disruptive in any way.  We would like to
restart the protocol buffer conversion work once the current
milestone is functionally complete.  So better as early as
possible rather than later.

jeromekelleher:
@calbach, this sounds like an excellent plan. The sooner the protobuf branch here is updated the better for us!

calbach:
OK I've merged in all of the latest sequence and variant annotation changes to this branch. Verified that everything compiles.

The biggest wart here is the `Attributes` message in `sequence_annotations.proto`. I tried to follow the type as closely as possible in proto, but there's no easy translation. I will defer to others on how we want to proceed there.

macieksmuga:
+1, let's get this in ASAP

jeromekelleher:
Looks like we're ready to go with this Avro->protobuf changeover. The compliance tests and server are also ready to merge at https://github.com/ga4gh/server/pull/1248 and https://github.com/ga4gh/compliance/pull/166.

+1.

kozbo:
We broke process here by merging before 3 days of review time had elapsed (my fault). Apologies and please let us know if we created a problem with this merge.

PR #548, opened 2016-01-28
Add missing GAException message

palfrey:
GAException is in the Avro descriptions, but not in protobuf

### DIFF
--- a/src/main/proto/ga4gh/common.proto
+++ b/src/main/proto/ga4gh/common.proto
@@ -8,7 +8,7 @@ package ga4gh;
 // When returning an HTTP error response, a server may also return a JSON formatted GAException
 // to better describe the error.
 message GAException {
-  // Numerical error code (should match the HTTP response code).
+  // Numerical error code
   int32 error_code = 1;
 
   // The error message.
--- a/src/main/proto/ga4gh/common.proto
+++ b/src/main/proto/ga4gh/common.proto
@@ -5,9 +5,14 @@ syntax = "proto3";
 
 package ga4gh;
 
+// When returning an HTTP error response, a server may also return a JSON formatted GAException
+// to better describe the error.
 message GAException {
-    int32 error_code = 1;
-    string message = 2;
+  // Numerical error code (should match the HTTP response code).
+  int32 error_code = 1;
+
+  // The error message.
+  string message = 2;
 }
 
 // Indicates the associated DNA strand for some data item.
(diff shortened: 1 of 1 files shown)


jeromekelleher:
@calbach Is this the simplest path to keeping compatibility with the existing API or should we go with the upstream proto equivalent? I'm easy either way.

calbach:
I guess this seems fine to me. I definitely prefer the upstream proto, but I'm not sure what the best way to include it is.

Refactoring it later should require minimal effort though, so this approach seems fine if it unblocks things. Adding some inline comments.

calbach:
On further thought, the upstream proto is not 100% compatible anyways.

src/main/proto/ga4gh/common.proto
```diff
@@ -5,6 +5,11 @@ syntax = "proto3";
 
 package ga4gh;
 
+message GAException {
```
calbach: Please comment as to how this is meant to be used (or at least port the comments from before):

```
/**
A general exception type.
*/
error GAException {
  /** The error message */
  string message;

  /** The numerical error code */
  int errorCode = -1;
}
```

My preference would probably be:

```
// When returning an HTTP error response, a server may also return a JSON formatted GAException
// to better describe the error.
message GAException {
  // Numerical error code (should match the HTTP response code).
  int32 error_code = 1;

  // The error message.
  string message = 2;
}
```

jeromekelleher:
If the upstream proto isn't 100% compatible I think we should put off refactoring to the standard proto until later (I do agree we should move to it though).

src/main/proto/ga4gh/common.proto
palfrey: Done

src/main/proto/ga4gh/common.proto
```diff
@@ -5,6 +5,16 @@ syntax = "proto3";
 
 package ga4gh;
 
+// When returning an HTTP error response, a server may also return a JSON formatted GAException
+// to better describe the error.
+message GAException {
+  // Numerical error code (should match the HTTP response code).
```
jeromekelleher: This is a new requirement, and contrary to how the reference server uses the status code. We should revert to using the old comments and update this later if necessary.

calbach: Interesting.. I figured this was a safe guess. I guess that's the problem with this documentation being extremely vague in the first place.

Reverting sounds good, though I'd like to keep the comment above GAException since it's optional, and I think accurately reflects how it is being used currently.

palfrey: Done

jeromekelleher:
Looks good to me,  +1

PR #549, opened 2016-01-28
Shift existing Avro to JSON wire compatibility with #547

calbach:
Precursor change to #547.

- Drop all Avro field defaults.
- Drop all primitive/repeated union types.
  - Message types are still nullable in proto3, primitives have 0 values, e.g. `false`, `0`, `""`.
- All `record` types are now union `null`, and default to null.
- Add ${ENUM_NAME}_UNSPECIFIED to all enums, use this as the default.
  - In proto3, enum 0-values should be a sane 'unset' default which is differentiable from the other values.
- s/sourceURI/sourceUri/
- Update any comment semantics as a result of the above changes.

Only remaining known incompatibility is #488.


src/main/resources/avro/common.avdl
```diff
@@ -128,9 +130,9 @@ record CigarUnit {
   /**
   `referenceSequence` is only used at mismatches (`SEQUENCE_MISMATCH`)
   and deletions (`DELETE`). Filling this field replaces the MD tag.
-  If the relevant information is not available, leave this field as `null`.
+  If the relevant information is not available, leave this field empty.
   */
-  union { null, string } referenceSequence = null;
+  string referenceSequence;
```
jeromekelleher: Is this the right mapping to proto semantics? Shouldn't it be 
```
string referenceSequence = "";
```
There is a difference I think. In the current version ``referenceSequence`` is optional, and if it's not provided within a JSON object it is defined as null. Here, ``referenceSequence`` is _required_ and must always be specified. 

From the [spec](https://developers.google.com/protocol-buffers/docs/proto3?hl=en#json)

> If a value is missing in the JSON-encoded data or if its value is null, it will be interpreted as the appropriate default value when parsed into a protocol buffer. If a field has the default value in the protocol buffer, it will be omitted in the JSON-encoded data by default to save space. 

So, to me, everywhere we have a ``union { null, string } x = null`` we now need to have ``string x = ""`` to be wire compatible with the proto implementation.

src/main/resources/avro/metadata.avdl
```diff
@@ -38,57 +38,57 @@ record Experiment {
...
   or slide unique identifier.
   */
-  union { null, string } platformUnit = null;
+  string platformUnit;
 
   /**
   A map of additional experiment information.
   */
-  map<array<string>> info = {};
+  map<array<string>> info;
```
jeromekelleher: Similarly, ``info`` should not now be required --- I think the correct semantics is for it to remain
```
map<array<string>> info = {};
```

src/main/resources/avro/metadatamethods.avdl
```diff
@@ -16,14 +16,14 @@ record SearchDatasetsRequest {
   Specifies the maximum number of results to return in a single page.
   If unspecified, a system default will be used.
   */
-  union { null, int } pageSize = null;
+  int pageSize;
```
jeromekelleher: Again, pageSize shouldn't be required now, but should default to 0 for proto compability. The semantics of pageSize=0 are clear here, it means 'the server decides'.

src/main/resources/avro/methods.avdl
```diff
@@ -10,7 +10,7 @@ error GAException {
   string message;
 
   /** The numerical error code */
-  int errorCode = -1;
+  int errorCode;
```
jeromekelleher: Defaults to 0 now.

jeromekelleher:
Looks good @calbach. What do you think about the defaults issue?

src/main/resources/avro/common.avdl
calbach: I guess I'm not clear on how these two cases are differentiable in Avro, so I defaulted to the syntax which *looks* closest to proto3. Isn't a default of `""` assumed if my type is `string`?

`string x = "";` vs `string x;`

That said, if the former makes things easier for the reference server or whatever Avro parsing scripts we have, I'll gladly make the change.

calbach:
Sounds like this will make the transition easier - added defaults throughout. Also ensured that all 'record' type fields now have a union of `null` and a default value of `null`, which is consistent with proto3.

src/main/resources/avro/variantmethods.avdl
```diff
@@ -67,17 +67,17 @@ record SearchVariantsRequest {
...
-  string variantSetId;
+  string variantSetId = "";
 
   /**
   Only return variant calls which belong to call sets with these IDs.
   If an empty array, returns variants without any call objects.
   If null, returns all variant calls.
   */
-  union { null, array<string> } callSetIds = null;
+  union { null, array<string> } callSetIds;
```
macieksmuga: Why not default this to null? - don't all protocol buffer fields need a default?

calbach: Oops, I actually didn't mean to modify this at all in this PR. Per the PR description, this field is meant to be resolved in #488.

src/main/resources/avro/reads.avdl
```diff
@@ -135,7 +135,7 @@ position and CIGAR array.
 */
 record LinearAlignment {
   /** The position of this alignment. */
-  Position position;
+  union { null, Position } position = null;
```
jeromekelleher: Can you explain the reasoning for this change a bit please?

jeromekelleher:
This looks good @calbach, I think should be very close to 1-1 semantics with protobuf. The only thing left to resolve really is the callSetIds issue in SearchVariantsRequest --- did we reach an agreement on this, or is it still unresolved?

I guess the next step is to start trying out proto-based clients against Avro based servers and vice-versa. I'll create a branch on the reference server to implement this PR. @macieksmuga, @david4096, can you guys do the same for the compliance suite please?

Assuming all this goes well I think we can then merge this PR, tag version 0.6.[whatever], and then swap master over the protobuf.

src/main/resources/avro/reads.avdl
calbach: In proto3, all message types are nullable and they default to `null`.

jeromekelleher: Makes sense, thanks.

calbach:
Agreed, and yes #488 is still pending. I'll reassess that PR later today.

calbach:
What's the latest here? Do the corresponding reference server / compliance branches exist? Are they in progress? Any blockers?

jeromekelleher:
The latest is that implementing this fully on the server is quite a bit of work, as we have to go right through the code to find all the places where we use None and replace this with empty string, 0, etc. The thing is, this has already been done in the protobuf branch, but we'd essentially have to do it all again for this schema. Given that this schema is purely a step to prove compatability with the proto world, I'm not convinced it's a good way to use our (stretched) developer resources. 

We can implement the client easily enough, which we can then point at the protobuf based server. This is a reasonable test. We could also use this schema to update the current Compliance Tests. Running the Avro based compliance tests against the protobuf based server would be a good test to see if anything fell between the cracks.

@diekhans, have you any thoughts on this? Should we implement this PR in the reference server so that we can make a final release of all three components base on Avro? Or can we make a final Avro based release without the corresponding server? Or do we bother with a final Avro based release at all?

diekhans:
I agree that it is not worth moving forward on making Avro
compatible.  We are very close to the protobuf merge, so
the value seems minimal.  It's not like we have thousands of
users show have to port.

This was an valuable exercise, however.

calbach:
OK, sounds good. I'm going to close this PR - but I'll leave my branch up for the foreseeable future in case it's useful as a reference or in case we change our minds.

PR #551, opened 2016-01-29
Removing autogenerated RSTs from doc/source/schemas after build.

macieksmuga:
The `make docs` (and `mvn package`) command auto-generates a collection of `.rst` files in `doc/source/schemas`, that are then committed to the repository. This is made worse by the fact that changing file `x.avdl` results in an update to the corresponding autogenerated `x.rst`, but _not_ an update of any `y.rst` for `y.avdl` which imports `x.avdl`. This in turn causes inconsistent documentation to be generated when changes to record definitions in frequently included `.avdl` files are made. 

This PR aims to solve the situation by explicitly removing the autogenerated `rst` files from git, and by adjusting the Makefiles to automatically scrub all these autogenerated files at the end of `make docs`.


david4096:
Pulled it down, ran `make docs` and was happy to see the source tree not polluted by the autogenerated .RSTs.

+1, @mbaudis as the original shell developer for this mind taking it for a spin?

diekhans:
-1 sadly.  This was added to make the readthedocs build work.  It will go away
with the conversion to protocol buffers.

diekhans:
this will be fixed in the conversion to prootocl buffers

PR #552, opened 2016-02-03
Use ISO8601 time for annotation API

david4096:
This PR switches the use of unix timestamp to ISO8601 time for records in the variant annotation API with the intention of extending to the rest of the schemas.

Most notably in alleleAnnotations.avdl

    -  /** The date this annotation was created in milliseconds from the epoch. */
    -  union { null, long } created = null;
    +  /** The :ref:`ISO 8601 <metadata_date_time>` time at which this record was created. */
    +  union { null, string } created = null;

I've also made changes to metadata.avdl to use the reference to the ISO8601 format as opposed to hard coding:

    -  Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
    +  Format: :ref:`ISO 8601 <metadata_date_time>`

Finally, the allele annotations documentation has been added to the toctree so it can be browsed to using the resulting HTML (thanks @andrewjesaitis !)

    +Allele Annotations
    +@@@@@@@@@@@@@@@@@@
    +
    +Allele annotations are additional pieces of data often generated by
    +algorithms which help to describe, classify, and understand variants.
    +
    +.. toctree::
    +   alleleAnnotations
    +
    +


src/main/resources/avro/metadata.avdl
```diff
@@ -168,13 +168,13 @@ record Analysis {
 
   /**
   The time at which this record was created. 
-  Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
+  Format: :ref:`ISO 8601 <metadata_date_time>`
   */
   union { null, string } recordCreateTime = null;
```
mbaudis: ```recordCreateTime```, ```recordCreateTime``` should be changed here to ```created```, ```updated```, too.

doc/source/api/metadata.rst
```diff
@@ -31,6 +31,82 @@ provider, users should not make semantic assumptions about that data.
...
+
+* date (e.g. date of birth) in YYYY-MM-DD
+    2015-02-10
+
+* time stamp in milliseconds in YYYY-MM-DDTHH:MM:SS.SSS
+    2015-02-10T00:03:42.123Z
+
+**Implementations**
+
+* created (G2P, metadata)
```
mbaudis: Since ```created```, ```updated``` are now universal, the specific branch pointers to G2P, metadata should be removed.

david4096:
@mbaudis You've included the `Z` timezone character in your example. Are servers expected to convert to GMT time when storing timestamps?

mbaudis:
@david4096 Don't really know.
* if they store as string, they keep the notation
* if they convert, they should adjust
* mostly, it won't matter (dates/years/intervals, and high res. times if at all matter on the actual instance)
However, this will anyway be deferred to some libraries I guess. So servers should just use their locally available way to handle this.

david4096:
@mbaudis @diekhans @macieksmuga can I get a +1/merge?

mbaudis:
+1

Sorry; the branch developments are a tad confusing. I thought this was a "managed override commit".

diekhans:
+1 and since this is the VA topic branch, you can merge yourself.

PR #559, opened 2016-02-17
Add biodata protocol and change sampleId to bioSampleId

david4096:
Also adds ontology term to metadata.avdl.

@mbaudis These are the schema changes I've plucked from your fork.

The first most important question is: which queries are we hoping to perform? Will SearchReadsRequest and SearchCallSetsRequest have the addition of an optional bioSampleId? Shall biosamples and individuals be query-able directly?


mbaudis:
That's great! The things I'm not seeing finalized here myself is the ```diseases```, ```phenotypes```, ```observations```, ```interventions```, ```developmentStage``` cluster of attributes in ```Individual```. These were intentionally put in for having discussions about structure and use (probably complex objects, but maybe fewer).

My preferred option to go forward here would be the limitation to cleared attributes: ```id```, ```name```, ```description```, ```created``` and ```updated```; as well as  ```sex```, ```species``` and ```info```.

It is a bit tricky to have a clear delineation between "forward looking/discussion" branches and the "probably o.k. to implement, pending some adjustments" PRs ...

diekhans:
I think it's very important to go forward with minimal implementation
a grow it.

Some of the missing things here seems to be more clinical
features as opposed to description of the biosample (phenotypes,
interventions), others seem very much related to the sample
(developmentStage).

Mark

helenp:
Disease and development stages can be linked.
Also developmental stage is also a feature of the organism as well as 
the sample.
It's not easy to split these, and can be use case dependent.

On 18/02/2016 15:07, Mark Diekhans wrote:
>
> I think it's very important to go forward with minimal implementation
> a grow it.
>
> Some of the missing things here seems to be more clinical
> features as opposed to description of the biosample (phenotypes,
> interventions), others seem very much related to the sample
> (developmentStage).
>
> Mark
>
> Michael Baudis <notifications@github.com> writes:
> > That's great! The things I'm not seeing finalized here myself is the 
> diseases,
> > phenotypes, observations, interventions, developmentStage cluster of 
> attributes
> > in Individual. These were intentionally put in for having 
> discussions about
> > structure and use (probably complex objects, but maybe fewer).
> >
> > My preferred option to go forward here would be the limitation to 
> cleared
> > attributes: id, name, description, created and updated; as well as 
> sex, species
> > and info.
> >
> > It is a bit tricky to have a clear delineation between "forward looking/
> > discussion" branches and the "probably o.k. to implement, pending some
> > adjustments" PRs ...
> >
> > —
> > Reply to this email directly or view it on GitHub.*
> >
>
> —
> <https://github.com/ga4gh/schemas/pull/559#issuecomment-185760458>.
>

-- 
Helen Parkinson, PhD
Team Leader - Samples, Phenotypes and Ontologies
Molecular Archives Cluster Head

-------------------------------------------
European Bioinformatics Institute (EMBL-EBI)
European Molecular Biology Laboratory
Wellcome Trust Genome Campus
Hinxton
Cambridge CB10 1SD
United Kingdom

EBI 01223 494672
For scheduling assistance please contact Lynn French lfrench@ebi.ac.uk, 01223 494 453
Skype: helen.parkinson.ebi
http://www.ebi.ac.uk/about/people/helen-parkinson

diekhans:
Good points, as always.

In my view, these reasons that relations by querying are better
than fix relations via linking.

mbaudis:
We basically need a type of wrapper object, which IMHO is identical for at least ```Disease``` and ```Phenotype```, and just contains a list of ontology terms + id, time data.
* id, so that e.g., evidences can be connected
* some time descriptor
* an array for OntologyTerm objects, called something like "characteristics", "features" or such
I am actually against making the classification between "disease" and "phenotype" - I do not understand the conceptual separation (and have said so from the beginning). The included OTs should be self-explanatory. Should be "Phenotypes" all the way. And then one could fold in "Observations" into the same general object; and also "interventions" (without the attempt to model something like complex clinical procedures, but use pharma ontologies (?) for the general classes...). So we would end up with a "Characteristics" wrapper for time(s)+id(s)+ontologies, which would have a general query mechanism etc.
Now we have to do examples (and counter-examples, if ...).
I'll try to write something ASAP.

david4096:
@diekhans I've updated this PR to include the latest metadata docs but more importantly there are changes to the queries for Reads and CallSets so that one can filter by bioSampleId.

There is also an endpoint `biosamples/search` for searching biosamples by name.

diekhans:
+1 lets get this into the topic branch

PR #560, opened 2016-02-19
G2P Topic branch

david4096:
This PR brings the schemas topic branch in line with master and includes some suggestions from https://github.com/ga4gh/schemas/pull/432.

Additionally updates ontology term and moves it to the metadata protocol following latest from metadata task team.

I would like to get this promptly merged. Since we're working on the feature branch we won't need referendum. @bwalsh if you can make sure this works for you we can move ahead. I expect the changes to ontology term will introduce (minor) bugs into the ref server g2p.

Below are a couple of issues that are *NOT* going to be addressed by this PR. We'll put up another WIP after this one is merged.

* Should `FeaturePhenotypeAssociation` contain a list of `Feature` or just featureId?
* Do we need to introduce a "g2pSet" so that there is not one G2P instance per server, but rather per dataset?


bwalsh:
+1 spoke w/ @gabrielsaldana/g2p all our tests pass.  thanks!

PR #561, opened 2016-02-23
Invert the semantics for properPlacement to improperPlacement

calbach:
This is somewhat related to the move to proto (#547), as nullability has been removed for all primitive values. For example, before this proposed change, an uninitialized `Read` proto would have a value of `properPlacement=false`.

Users of Google's implementation of the v0.5.1 Reads API have expressed that this is an inconvenient and unnatural default, as many programs will filter out reads which have not been marked as 'properly placed'. For example, when generating a simple `Read` for a unit test, one would almost always want to set the `Read` to properly placed - more natural would be only to mark the read as **improperly** placed only where needed. I tend to agree, given the usage of this field in practice. This seems to be the expected state for a read - improper placement is the exception.

### DIFF
--- a/src/main/resources/avro/reads.avdl
+++ b/src/main/resources/avro/reads.avdl
@@ -183,9 +183,9 @@ record ReadAlignment {
 
   /**
   The orientation and the distance between reads from the fragment are
-  consistent with the sequencing protocol (equivalent to SAM flag 0x2)
+  inconsistent with the sequencing protocol (inverse of SAM flag 0x2)
   */
-  union { null, boolean } properPlacement = null;
+  union { null, boolean } improperPlacement = null;
 
   /** The fragment is a PCR or optical duplicate (SAM flag 0x400). */
   union { null, boolean } duplicateFragment = null;


diekhans:
+1  agree this is clearer.

jeromekelleher:
Makes sense to me, +1.

calbach:
OK, nearly a week has passed now so I'd say this is eligible to be merged.

The one thing I'm not sure on is the reference server and compliance tests. Should I be preparing PRs to update each repo accordingly for a small change like this?

jeromekelleher:
I agree, we should merge this.

Re the server and the compliance tests, this is a good question. I don't think we have a protocol for small changes like this. IMO I think it would be overkill to require PRs against all three for something like this, as it's a trivial change for respective development teams. I'm open to arguments though, I guess there is a chance that schema changes could slip through unnoticed if we don't require code server and compliance suite changes.

diekhans:
The goal was to have working masters in all three.  If we don't
do PRs within a reasonable time frame, we will have breakage
creep in.

We don't need to rush a release just for this, but I think we
should do PRs for all three and make sure it actually works.

This doesn't mean CH needs to do this, but it's great if he does
;-)

calbach:
I've taken a stab at the reference server and compliance PRs (referenced above). They will need some more help before they can be merged, but I don't see any major blockers there. Going to merge this now.

Process-wise, it wasn't too difficult to create the other two PRs, but I'm primarily depending on the reviewers to ensure they actually work. I think that for small features like this, we should aim to not require the other two PRs unless there is significant reason to believe that it will be challenging or difficult to implement, e.g. an entire new API surface.

diekhans:
@dcolligan can you help CH get these changes integrated into the
server?

diekhans:
The goal is to keep master working in all three components, so
we need to have multiple PRs if changes are required.

However, that doesn't mean the same person has to do it.
We certainly want people to contribute and not have to
become an expert on the server or compliance suite.

You can create tickets in compliance and/or server and link them
to the PR and then pester the integration group until someone
implements it.

dcolligan:
@diekhans server PR is here: https://github.com/ga4gh/server/pull/960

diekhans:
@david4096, can you review this.

PR #562, opened 2016-02-24
Remove impact and change created to createDateTime and updated to updateDateTime

david4096:
Keeps variant annotation records in line with latest from metadata task team. @diekhans

### DIFF
--- a/doc/source/schemas/alleleAnnotations.rst
+++ b/doc/source/schemas/alleleAnnotations.rst
@@ -141,14 +141,14 @@ This protocol defines types used by the GA4GH Allele Annotation API.
   :field description:
     A description of the experiment.
   :type description: null|string
-  :field created:
+  :field createTime:
     The time at which this record was created. 
       Format: :ref:`ISO 8601 <metadata_date_time>`
-  :type created: string
-  :field updated:
+  :type createTime: string
+  :field updateTime:
     The time at which this record was last updated.
       Format: :ref:`ISO 8601 <metadata_date_time>`
-  :type updated: string
+  :type updateTime: string
   :field runTime:
     The time at which this experiment was performed.
       Granularity here is variable (e.g. date only).
@@ -219,14 +219,14 @@ This protocol defines types used by the GA4GH Allele Annotation API.
   :type name: null|string
   :field description:
   :type description: null|string
-  :field created:
+  :field createTime:
     The time at which this record was created. 
       Format: :ref:`ISO 8601 <metadata_date_time>`
-  :type created: null|string
-  :field updated:
+  :type createTime: null|string
+  :field updateTime:
     The time at which this record was last updated.
       Format: :ref:`ISO 8601 <metadata_date_time>`
-  :type updated: string
+  :type updateTime: string
   :field type:
     The type of analysis.
   :type type: null|string
@@ -558,9 +558,9 @@ This protocol defines types used by the GA4GH Allele Annotation API.
   :field variantAnnotationSetId:
     The ID of the variant annotation set this record belongs to.
   :type variantAnnotationSetId: string
-  :field created:
+  :field createTime:
     The :ref:`ISO 8601 <metadata_date_time>` time at which this record was created.
-  :type created: null|string
+  :type createTime: null|string
   :field transcriptEffects:
     The transcript effect annotation for the alleles of this variant. Each one
       represents the effect of a single allele on a single transcript.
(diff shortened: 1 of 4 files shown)


doc/source/schemas/alleleAnnotations.rst
```diff
@@ -141,14 +141,14 @@ This protocol defines types used by the GA4GH Allele Annotation API.
   :field description:
     A description of the experiment.
   :type description: null|string
-  :field created:
+  :field createTime:
```
reece: Perhaps I'm being too pedantic, but "Time" to mean datetime or timestamp seems misleading. Note that there is an ISO 8601 Time that refers to just the time.

Could we make this createdDataTime or createdTimestamp? While we're at it, we should probably require UTC.

mbaudis: So: This was ```created``` (ms from epoch) => ```recordCreateTime``` (ISO, my implementation) => ```created``` (ISO, since "created" existed somewhere in the schema) => ```createTime``` (ISO, since ```created``` is very ambiguous).

I am certainly for  ```recordCreateTime```, but flexible (had accepted ```created```, just to get ISO).

diekhans: I will take pedantic and precise over fuzzy any day

createDateTime is fine; we need to document these conventions.

Requiring UTC seems unnecessary, as the timezone should always
should included, so conversion to any timezone is possible

@mbaudis, which branch is the time format document?  I can't
find it.

mbaudis: @diekhans Time definitions are in the api -> metadata.rst / html file; in my biosample branch and derivatives. A rendered doc is here:
* http://ga4ghdata.org/Sites/github/mbaudis/biosample/html/api/metadata.html

david4096:
I've changed the time fields to createDateTime and updateDateTime. @diekhans @reece mind +1/merging when you get a chance?

diekhans:
+1

pcingola:
-1 
We should not remove IMPACT field. 
It's extremely useful.

diekhans:
Since there is disagreement on the impact field, may I suggest
that the VA pull request goes ahead without it and the VATT work
on it until some consensus is reached and it can be
incrementally added.

The other option is to put the VA integration into master on
hold which this is resolved.

pcingola:
I disagree.
  The lack of consensus is to remove it. It was agreed to be added
long time ago with consensus from all the group. I'd suggest to
go the other way around. Let's keep it (there is already an issue
opened to discuss removing it, I don't understand why it was just
removed without consulting with the group).
  On top of that, we already know how hard is to re-add things once
they've been unilaterally removed (re-adding VATT's work is taking
over quite a few months of work).

diekhans:
We though this came from the VATT.  UCSC will wait for the VATT
to decide.

reece:
We are scheduled to discuss impact and related data modeling policies at the VATT March 9 meeting. However, stalling the variant_annotation merge delays downstream activities, and therefore we won't do that.

That leaves two options: including or excluding the impact component of the schema at this time. Excluding it now leaves open the option of including it later. (And, technically, including it now leaves open the option of retracting it later, but I'm opposed to that route.)

I know of 6 opinions on this topic. @pcingola is in favor of keeping impact as-is. The other five -- @fcunningham, @sarahhunt, @jeromekelleher, @diekhans, and I -- are in favor of removing it *at this time*.

Furthermore, Daniel MacArthur, Peter Robinson, Deanna Church, Richard Durbin, and Tony Brookes commented on the ga4gh-dwg-annotation list (~Dec 9-18) with comments that I would summarize as 1) most had significant concerns regarding the impact and modifier parts of the schema as-is, and 2) some believed that these components might be valuable *with revisions* (like removing modifier). None believed that acceptance was clear cut.

All of you should correct me here, on this this thread, if you believe that I am incorrectly summarizing the data above.

So, my opinion is that impact is worth considering and that we should do so when we have a clearer understanding and consensus of terminology and utility. We can more readily consider a PR that includes impact after the March 9 meeting. In order to keep the project moving forward, we should merge this PR into variant_annotation (and, presumably, variant_annotation then into master as discussed on that PR).

+1

pcingola:
No Reece. The recommendation was to leave it,  and evaluate AFTER the merge
was completed.

-1

reece:
What recommendation and when? My polls of people were 2 days ago.

If you can point to any github comments in favor of keeping impact, I would be happy to have the record amended. I honestly know of no comments in favor of impact at this point. Apologies if I missed them.

stevenbrenner:
Thanks everyone for this discussion.  I concur with @pcingola that Impact is very important and should ultimately be part of the variant annotation.  My understanding of the community sentiment and history agrees with @reece, and therefore I agree with @diekhans to defer its inclusion (i.e., remove it _for now_).  We will discuss on the 8 March DWG VATT call and can decide how to include then.

PR #563, opened 2016-02-24
Adding FPAset and queries from @sarahhunt

david4096:
This PR addresses the need to serve multiple g2psets from a single server by plucking changes @sarahhunt proposed here: https://github.com/ohsu-computational-biology/schemas/pull/1/files.

Also adding toctree and moving G2P docs to docs/api. @bwalsh @dcolligan

### DIFF
--- a/src/main/resources/avro/genotypephenotype.avdl
+++ b/src/main/resources/avro/genotypephenotype.avdl
@@ -156,7 +156,7 @@ record FeaturePhenotypeAssociation {
   The ID of the PhenotypeAssociationSet this FeaturePhenotypeAssociation
   belongs to. 
   */
-  string phenotypeAssociationId;
+  string phenotypeAssociationSetId;
 
   /**
     The set of features of the organism that bears the phenotype.
--- a/src/main/resources/avro/genotypephenotype.avdl
+++ b/src/main/resources/avro/genotypephenotype.avdl
@@ -38,6 +38,26 @@ import idl "common.avdl";
 import idl "metadata.avdl";
 import idl "sequenceAnnotations.avdl";
 
+/** 
+A PhenotypeAssociationSet is a collection of phenotype association results. 
+Such results are grouped by data source and possibly release version or analysis 
+type. 
+*/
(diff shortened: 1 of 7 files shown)


src/main/resources/avro/genotypephenotype.avdl
```diff
@@ -131,6 +151,12 @@ such as resulting from alternative experiments and analysis.
 */
 record FeaturePhenotypeAssociation {
   string id;
+  
+  /** 
+  The ID of the PhenotypeAssociationSet this FeaturePhenotypeAssociation
+  belongs to. 
+  */
+  string phenotypeAssociationId;
```
jeromekelleher: Typo here, should be phenotypeAssociationSetId

jeromekelleher:
We definitely need something like this. Is everyone happy with the PhenotypeAssociationSet terminology?

src/main/resources/avro/genotypephenotype.avdl
david4096: Thanks nice catch @jeromekelleher , updated :)

david4096:
The other comment that I'll make here (thanks @diekhans) is that the list of `Feature` in a FPA should probably be a list of `featureId`. A query against sequence annotations with that featureId will give the entirety of the record.

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -101,6 +150,11 @@ full feature/phenotype/evidence objects.
 */
 record SearchGenotypePhenotypeRequest {
 
+  /**
+  The `PhenotypeAssociationSet` to search.
+  */
+  string phenotypeAssociationSetId;
+
   union {null, string, ExternalIdentifierQuery, OntologyTermQuery,
```
david4096: These need to be better specified and documented. For example, as @diekhans mentioned, does it make sense to have an external  ID query on `Evidence` when evidence doesn't have an external ID field? What are the strings matching on?

bwalsh:
+1 looks good

PR #570, opened 2016-03-09
Improve reads searching

david4096:
Reads search is specified in such a way that it can be computationally infeasible to implement. The current state of the schemas allows searches across ReadGroupSets through the `readGroupIds` field of a SearchReadsRequest. This is akin to performing a sort across multiple BAM files.

This PR requires that a user specify a `readGroupSetId` when searching reads. Additionally I have added a comment that states the array of read group IDs must belong to a single Read Group Set.

### DIFF
--- a/src/main/resources/avro/readmethods.avdl
+++ b/src/main/resources/avro/readmethods.avdl
@@ -15,9 +15,14 @@ specified, all queried `ReadGroup`s must be aligned to the same `ReferenceSet`.
 */
 record SearchReadsRequest {
   /**
-  The ReadGroups to search. At least one id must be specified.
+  Specifies the Read Group Set to search within.
   */
-  array<string> readGroupIds;
+  string readGroupSetId;
+  /**
+  Limits results to the given Read Group Ids. These read groups
+  must be within the specified read group set.
+  */
+  union { null, array<string> } readGroupIds = null;
 
   /**
   The reference to query. Leaving blank returns results from all
--- a/doc/source/schemas/readmethods.rst
+++ b/doc/source/schemas/readmethods.rst
@@ -442,10 +442,10 @@ Gets a `org.ga4gh.models.ReadGroup` by ID.
   :field fragmentName:
     The fragment name. Equivalent to QNAME (query template name) in SAM.
   :type fragmentName: string
-  :field properPlacement:
+  :field improperPlacement:
     The orientation and the distance between reads from the fragment are
-      consistent with the sequencing protocol (equivalent to SAM flag 0x2)
-  :type properPlacement: null|boolean
+      inconsistent with the sequencing protocol (inverse of SAM flag 0x2)
+  :type improperPlacement: null|boolean
   :field duplicateFragment:
     The fragment is a PCR or optical duplicate (SAM flag 0x400).
   :type duplicateFragment: null|boolean
@@ -522,9 +522,13 @@ Gets a `org.ga4gh.models.ReadGroup` by ID.
 
 .. avro:record:: SearchReadsRequest
 
+  :field readGroupSetId:
+    Specifies the Read Group Set to search within.
+  :type readGroupSetId: string
   :field readGroupIds:
-    The ReadGroups to search. At least one id must be specified.
-  :type readGroupIds: array<string>
+    Limits results to the given Read Group Ids. These read groups
+      must be within the specified read group set.
+  :type readGroupIds: null|array<string>
   :field referenceId:
     The reference to query. Leaving blank returns results from all
       references, including unmapped reads - this could be very large.


diekhans:
addressed by  https://github.com/ga4gh/schemas/pull/570

src/main/resources/avro/readmethods.avdl
```diff
@@ -15,9 +15,14 @@ specified, all queried `ReadGroup`s must be aligned to the same `ReferenceSet`.
...
 record SearchReadsRequest {
   /**
-  The ReadGroups to search. At least one id must be specified.
+  Specifies the Read Group Set to search within.
   */
-  array<string> readGroupIds;
+  string readGroupSetId;
+  /**
+  Limits results to the given Read Group Ids. These read groups
+  must be within the specified read group set.
```
calbach: Should be explicit about the behavior when 0 read group ids are specified.

calbach:
I'm +0 for now, though leaning towards supporting. Multi-`ReadGroupSet` search is convenient for clients and it could prove useful to push it into the server for performance gains (if any). As of recently, there continues to be [support for leaving the functionality in place](https://github.com/ga4gh/schemas/pull/253#discussion_r26700450). In the paginated API, though, it is less likely that encoding this in the API (as opposed to the client) will yield much in the way of performance gains.

I'm leaning towards supporting this because it simplifies the API, makes it more consistent with the rest of the API surface, and reduces complexity for implementers. In the future we may consider adding this functionality back if it can provably show value.

I'll also add that while I agree that this feature is complicated to implement, I don't understand the assertion that it is 'computationally infeasible' for a BAM-backed server (assuming reasonable preconditions on the BAM files...).

diekhans:
The current API requires doing a sort-merge of reads across a
large range and large number of ReadGroupSets (BAMs) is
computationally expensive.

A 5mb region across 10,000 BAMs is a very reasonable request.
40x coverage with 100 base reads is a sort-merge on about 20
billion reads seems very expensive to me.  Given that the ENA
isn't willing to re-compress start and end BAM blocks, it's hard
see how they would agree to doing the large sort-merge required
by the API.

Do we have a use case for sort-merging reads for all samples?
All use cases I know of would then de-multiplex the results into
the individual samples.  For this quantity of data to be feed
into existing applications, it would mostly like be writing
10,000 BAMs.

My experience (being responsible for ~100,000 BAMs) is that
existing data is almost always stored sorted on a
per-readgroupset basis and analyzed assuming this sorting

According to Jonathan, the Google API doesn't implement sorting
across readsgroupssets and Google at least has the compute power
to do this ;-)

Once streaming is implemented, it's very reasonable to have a
multiple readgroupset query provide the sorting requirement
within a readgroupset, not across readgroupsets.  A reasonable
client API could hide the difference between streaming and
paging, so I would not object to having the query take an array
of readgroupset ids now, as long as the sorting requirement is
removed.

diekhans:
There was very strong argument that the reads API needs to do the sort merge is a requirement for multi-sample BAM calling.  The is a strong desire to move the merge to the server, not the client.

We will update the documentation to capture this use case.

calbach:
For the record, this is not correct:

```
the Google API doesn't implement sorting
across readsgroupssets
```

Jonathan misspoke (just talked to him). The [Google API's SearchReads](https://cloud.google.com/genomics/reference/rest/v1/reads/search) accepts one or more read group set IDs and returns a  paginated stream of reads which are ordered by position.

Note: We do not support searching by arbitrary `readGroupIds` from different read group sets (only from the same read group set) as we remain skeptical that this is a useful query. We may add support in the future - this would not be a significant amount of work, but it's not been requested.

diekhans:
thanks @calbach

I am glad this is coming down to me being unclear uses cases rather
than the API being off.

pgrosu:
@calbach GA4GH is not built for the old way of analyzing data, though it can still be used that way.  So we do not want to only query based on subsets of ReadGroupSets - which can slow us down - but we want triggers attached to metadata and specific sequence variants as the data is streaming into the repositories.  

This means auto-sorting for on-the-fly pileups to allow for finding similar, rare samples. And then we want to open a genome browser and find which other rare mutations these highly-qc'ed samples are generating across the genome.

Think of this, let's say a variant is automatically detected based on subsets of auto-triggered alignments of some fresh data to subsets of reads in the system.  Has this been seen before, and were we in need of more samples like this?  Do we have a stored graph of this variant region?  If this is a cancer subtype, are there similar regions of stored matching paths that would project a molecular evolution?  Can we map it to their RNAseq activity for this region and how do they compare?  What are other diseases that map to these regions of the genome, and how similar are their variants.  I can go forever in building it up more and more, but I think you get the idea.

In my view GA4GH opens up genomic analysis to orders of magnitude larger than older tools, which requires a different and more fluid way of processing such data.

~p

pgrosu:
@diekhans I don't think you are unclear regarding use cases, and I think those should be covered too or we risk reinventing the old tools again with GA4GH.

mbaudis:
@diekhans Just a very careful question: What are, besides benchmarking variation callers, use cases for direct/federated access to reads? Sorry for my ignorance!

diekhans:
@mbaudis, we will find out, I am going to take a crack at the
use cases in the documentation.

vadimzalunin:
:+1: I'd rather have this than an uncertainty in provider's behavior. Clients don't have to deal with the specifics of each provider. 
PS sort-merge is "almost analysis" imo. If a server can guarantee sort-merge it is only a step away from providing coverage, consensus, variations etc.

PR #576, opened 2016-03-15
Switched childIds array to optional siblingRank number.

macieksmuga:
After discussion with @david4096 and @saupchurch, we feel that this will be simpler to support, and in most cases where child features are naturally ordered by genomic coordinate, the childIds field of the parent didn't add any functionality (children of a given feature can already be queried with the features/search method).

### DIFF
--- a/doc/source/schemas/sequenceAnnotationmethods.rst
+++ b/doc/source/schemas/sequenceAnnotationmethods.rst
@@ -1,0 +1,457 @@
+SequenceAnnotationMethods
+*************************
+
+ .. function:: searchFeatureSets(request)
+
+  :param request: SearchFeatureSetsRequest: This request maps to the body of `POST /featuresets/search` as JSON.
+  :return type: SearchFeatureSetsResponse
+  :throws: GAException
+
+Gets a list of `FeatureSet` matching the search criteria.
+
+  `POST /featuresets/search` must accept a JSON version of
+  `SearchFeatureSetsRequest` as the post body and will return a JSON version
+  of `SearchFeatureSetsResponse`.
+
+ .. function:: getFeatureSet(id)
+
+  :param id: string: The ID of the `FeatureSet`.
+  :return type: org.ga4gh.models.FeatureSet
+  :throws: GAException
+
+Gets a `FeatureSet` by ID.
+  `GET /featuresets/{id}` will return a JSON version of `FeatureSet`.
+
+ .. function:: getFeature(id)
+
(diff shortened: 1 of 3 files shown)


macieksmuga:
@diekhans, @saupchurch, @calbach does this look reasonable? Who else should I check with?

diekhans:
-1

We specifically went to explicit child list to make the data representation simpler and consistent.  One always needs to know the order of the children.  This way the data structure is always used in the same way.

macieksmuga:
@diekhans could I convince you to reconsider?

Logically and from a client's perspective, I believe this proposed scheme is equivalent to the existing one: Given an ordered array of `childIds` a client can compute their `siblingRank`s, and given any `feature` object, the client can easily obtain all its child features, then using their `siblingId`s (if provided) together with their genomic coordinates, can recreate the `childIds` array. Note this doesn't involve any extra steps beyond using a simple `features/search` query that specifies the `parentId`.

From an implementation standpoint, I proposed this PR since it's possible to use a line-by-line parser to generate a database that doesn't contain a `childIds` column from a `GFF3` (and presumably `BED`) file, as opposed to the current approach which needs to load a full graph of `Gff3Set` objects into memory.

diekhans:
I am very willing to consider.  Not right now; it already works.  

Client's perspective, this is trading off wire size vs easy of
use.  Google and EBI have some proposed changes that add
complexity to the implementation to make it easier for the
client to use.  We should reconsiders the approach to defining
ordering requirements when we consider the overall
sequence annotation API.

It seems the motivation is to speed up loading a large database
on a laptop.  It's not clear if this is an important goal,
however there are other ways to address this that don't require
modifying the API.

macieksmuga:
I believe the reason for doing schema development in these topic branches is so that we can easily evolve the schemas as we gain more experience with them (this PR still hasn't been accepted into the official sequence_annotations branch, therefore is still malleable). What I've discovered in the process of implementing the sequence annotations reference server is that the `childIds` array is no more descriptive than the proposed change here (I still believe the client gains little, simplicity-wise, from having the childIds array included in a feature - as argued above), while being more preprocessing-intensive and overall less flexible as new annotations are added.

Also, what are the Google and EBI proposed changes relevant to this PR? I'd very much like to make sure we include our partners' input in this discussion.

macieksmuga:
Looks like this PR won't make it in before the April 25 milestone.  I'm closing the PR for now, but want to re-open this discussion once the current schema has been merged into master.

PR #579, opened 2016-03-17
Individual in biodata

david4096:
Adds Individual to biodata protocol and updates the branch with the latest master schema (which now includes variant annotation, hence all the extra commits).

This functionality will allow more metadata to be stored regarding individual samples, as well as the individual a sample may derive from.
Endpoints added:
- POST `/biosamples/search`
- GET `/biosamples/<id>`
- POST `/individuals/search`
- GET `/individuals/<id>`

Endpoints altered:
- `callsets/search` has optional bioSampleId field
- `readgroupsets/search` has optional bioSampleId field

Most importantly:

https://github.com/david4096/schemas/blob/biodata/src/main/resources/avro/biodatamethods.avdl
https://github.com/david4096/schemas/blob/biodata/src/main/resources/avro/biodata.avdl


src/main/resources/avro/readmethods.avdl
```diff
@@ -122,6 +116,12 @@ record SearchReadGroupSetsRequest {
   Only return read group sets with this name (case-sensitive, exact match).
   */
   union { null, string } name = null;
+  
+  /**
+  Specifying the id of a BioSample record will return only readgroups
+  with the given bioSampleId.
+  */
```
macieksmuga: This comment alone doesn't make it clear that the returned `readGroupSet`s will have their `readGroup`s filtered by this field.

david4096: Updated below:
```
Only readgroups that
 +match an optionally supplied bioSampleId will be included in 
 +the response.
```

mbaudis:
@david4096 I had added some more time format documentation to ```/doc/source/api/metadata.rst```, as req. by @diekhans. Should find its way in somewhere...
https://github.com/mbaudis/schemas/blob/biosample/doc/source/api/metadata.rst

david4096:
Added @mbaudis' updated metadata docs with the modification: createTime -> createDateTime and updateTime -> updateDateTime.

PR #586, opened 2016-03-28
Biodata

david4096:
Add link to individual in biodata schema documentation. Change `createTime` to `createDateTime` and `updateTime` to `updateDateTime`.

@diekhans @mbaudis I would like to open up a PR against master soon, would be great to have the biodata protocol reviewed and edited before it is opened to the group.

### DIFF
--- a/doc/source/api/metadata.rst
+++ b/doc/source/api/metadata.rst
@@ -28,6 +28,9 @@ Metadata Records
 :ref:`BioSample<metadata_biosample>`
 ====================================
 
+:ref:`Individual<metadata_individual>`
+======================================
+
 :ref:`Dataset<metadata_dataset>`
 ====================================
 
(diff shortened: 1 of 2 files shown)


david4096:
Note: I've added the `datasetId` field to both `Individual` and `BioSample`.

mbaudis:
@david4096 This is problematic. IMO datasets should reference the included biosampleIds. Each Individual / Biosample will potentially be part of n^x Dataset's.

david4096:
@mbaudis Ok, no problem it's removed from the record so implementors are free to have biosamples/individuals across datasets. Note that when searching for biosample or individual records one must specify a datasetId.

diekhans:
maybe making the names of the create/update field consistent with reads is the way to go for now.  Adding  *DataTime  didn't really improve the clarity of what these fields really mean (create of data vs load into server).

diekhans:
This all looks great for a first pass.  Lets get it in and use it.

mbaudis:
+1

PR #591, opened 2016-03-30
Sequence annotation

macieksmuga:
Definition of `FeatureSet` and `Feature` records, together with get and search endpoints for each.


diekhans:
+1  .. while the query methods will most likely need to evolve some, it's important to get the data model out there so the methods can evolves through real use cases

david4096:
For the CI failure: https://github.com/ga4gh/schemas/pull/592

doc/source/api/sequence_annotations.rst
```diff
@@ -0,0 +1,46 @@
...
+
+------------------------------
+The Sequence Annotation Schema
+------------------------------
+
+This model is similar to that used by the standard `GFF3`_ file format.
+
+.. _GFF3: http://sequenceontology.org/resources/gff3.html
+
+The main differences concern the deprecation and replacement of discontinuous features, the replacing
```
dzerbino: Is that necessarily possible/relevant? For example in GFF a fusion transcript would have two parent genes. Duplicating the transcript seems inappropriate?

doc/source/api/sequence_annotations.rst
```diff
@@ -0,0 +1,46 @@
...
+of multi-parent features with multiple copies of that feature, and the ability to impose an explicit order on child features.
```
dzerbino: Is that necessarily possible/relevant? For example in GFF a fusion transcript would have two parent genes. Duplicating the transcript seems inappropriate?

src/main/resources/avro/sequenceAnnotations.avdl
```diff
@@ -0,0 +1,137 @@
...
+
+  /*
+  A set of sequence features annotations
+  */
+  record FeatureSet {
+    /** The ID of this annotation set. */
+    string id;
+
+    /** The ID of the dataset this annotation set belongs to. */
+   union { null, string } datasetId = null;
```
sarahhunt: There is no way to search for a FeatureSet which is not in a Dataset, but FeatureSet's don't have to belong to Datasets. Should datasetID be required here?

doc/source/api/sequence_annotations.rst
```diff
@@ -0,0 +1,46 @@
...
+************************
+Sequence Annotations API
+************************
+For the Sequence Annotation schema definitions, see `Sequence Annotation schema <../schemas/sequenceAnnotations.html>`_
+
+
+------------------------
+Feature Based Hierarchy
+------------------------
+The central object of the GA4GH Sequence Annotation API is a Feature.  The Feature describes an interval of interest on some reference(s).  It has a span from a start position to a stop position as well as descriptive data.  A Feature has one parent Feature, and can have an ordered array of child Features, which enables the construction of more complex representations in a hierarchical way.
```
sarahhunt: Should this be 'A Feature can have one parent Feature' rather than 'A Feature has one parent Feature'?

doc/source/api/sequence_annotations.rst
```diff
@@ -0,0 +1,46 @@
...
+
+Splicing (other post-transcriptional modifications?) can occur with degrees of complexity.  A ‘typical’ splice will result in a mature transcript with exon in positional (numerical) order in a head-to-tail orientation.  Back splicing (tail-to-head) can result in transcripts with the exon order reversed (1-3-2-4 instead of 1-2-3-4) and even circular RNA.  The exon order in a transcript as well as the orientation of the splice should be discoverable via the API.  In a more general case, the API should allow child features to have an ordered relationship.
+
+The annotation API needs to also be flexible enough to handle multiple references in the same gene or transcript.  This is needed to cover the cases of fusion genes or inter-chromosomal translocations.
```
sarahhunt: I don't understand this comment - is it aspirational? The Feature has a single referenceName in the schema.

doc/source/api/index.rst
```diff
@@ -43,6 +43,14 @@ system for reads and variants.
 .. toctree::
    references
 
+Sequence Annotations
```
reece: "Sequence Features" seems more descriptive (and avoids the well-worn term "annotations").

doc/source/schemas/sequenceAnnotationmethods.rst
```diff
@@ -0,0 +1,457 @@
+SequenceAnnotationMethods
```
reece: SequenceFeatureMethods?

doc/source/schemas/sequenceAnnotationmethods.rst
```diff
@@ -0,0 +1,457 @@
...
+  (0-based).
+
+.. avro:record:: ExternalIdentifier
+
+  :field database:
+    The source of the identifier.
+      (e.g. `Ensembl`)
+  :type database: string
+  :field identifier:
+    The ID defined by the external database.
```
reece: I'm always nervous about the use of accessions as a proxy for an ill-defined concept. One example is ENST: without a versioned ENST, it's quite hard to know precisely which sequence version this refers to. Furthermore, Ensembl fasta file releases typically have two very different sequences with the same accession: the cdna fastas have the cdna sequence, and the cds files have the cds, but the same accession is used in both. Example 2: NCBI transcript accessions are used as though they imply a fixed alignment to the genome, but they do not. That means that interpretations made with such an alignment are implicitly dependent on hidden information. As the GA4GH, we should be more insistent on referencing precisely defined, immutable concepts. (These comments are more my own sky-is-falling rather than a comment about the PR.)

src/main/resources/avro/sequenceAnnotations.avdl
```diff
@@ -0,0 +1,137 @@
...
+    /**
+    Identifier for the containing feature set.
+    */
+    string featureSetId;
+
+    /**
+    The reference on which this feature occurs.
+    (e.g. `chr20` or `X`)
+    */
+    string referenceName;
```
reece: We should exclusively refer to sequences by unique accession. "chr20" does not uniquely refer to a sequence. (And, on top of that, "20" -- not "chr20" -- is the correct official GRCh name for chromosome 20.)

doc/source/api/index.rst
diekhans: The name was chosen because SequenceAnnotations will include more than Features;  The second data type is continuous (wiggle) was not include in this PR to speed implementation.

doc/source/api/sequence_annotations.rst
diekhans: Ohhhhh  good point @sarahhunt , comments @saupchurch

src/main/resources/avro/sequenceAnnotations.avdl
```diff
@@ -0,0 +1,137 @@
...
+  import idl "metadata.avdl";
+
+  /**
+  Type defining a collection of attributes associated with various protocol
+  records.  Each attribute is a name that maps to an array of one or more
+  values.  Values can be strings, external identifiers, or ontology terms.
+  Values should be split into the array elements instead of using a separator
+  syntax that needs to parsed.
+  */
+  // TODO: how are multiple instances of a given attribute vs multiple values
```
reece: Do we have any examples at hand/in mind? I think it's hard to assess whether a particular choice of model is good without agreeing on some use cases.

doc/source/api/sequence_annotations.rst
macieksmuga: @sarahhunt My understanding is that features that are the roots of their hierarchy will have no parents, thus the parent is optional.

src/main/resources/avro/sequenceAnnotations.avdl
macieksmuga: Good question. My understanding is that a dataset is a very general entity, mostly related to data origin bookkeeping, and not necessarily associated with any specific reads or variants. Thus feature sets may be (and perhaps should be?) contained in a separate dataset.

doc/source/schemas/sequenceAnnotationmethods.rst
diekhans: @reece,  the sky *is* falling.   Early attempts to more formalize the data semantics were rejected with the notation that the API provides the mechanism and the data semantics can be layered on top.  That said, we should maybe sure we can represent the semantics.

src/main/resources/avro/sequenceAnnotations.avdl
macieksmuga: Using the sequence SHA/MD5 as its identifier would work with the current linear reference, but may run us into trouble with (hopefully) future graph reference sets, which are composed of lots of smaller reference segments, many of which may have identical sequences, and are differentiated only by where they join to the rest of the reference.

diekhans: this should be referenceId, not reference name.

src/main/resources/avro/sequenceAnnotations.avdl
diekhans: it don't think dataset should be nullable anywhere

doc/source/api/sequence_annotations.rst
diekhans: readthrough transcripts in gencode are either represented as a third gene or belong to one of the parent genes if that is how that gene is used in practice.

We should create a ticket on these kind of representation issues and discuss them with a wider audience, including the SO people.  [although we want to move forwards with this PR because it's blocking other work and is only alpha]

doc/source/schemas/sequenceAnnotationmethods.rst
diekhans: wiggle coming in  the future.

src/main/resources/avro/sequenceAnnotations.avdl
diekhans: multiple values are part in GFF3 (e.g. tag in gencode)

Single values that have multiple parts is a bit more speculative, but in general we have values of ExternalId and OntologyTerm.  

I think we need to consider these across all of the info files in the GA4GH API.

src/main/resources/avro/sequenceAnnotations.avdl
reece: @macieksmuga: I agree that we should plan for sequence graphs. In fact, this is exactly why I want hashes exclusively: it is the only identifier that we can be assured will 1) exist and 2) be immutable. Ultimately, all features *must* be on a linear reference ultimately otherwise the notion of a start and end are meaningless. 

A hash represents a path that is composed of sequence segments, which may also be hashed. It just so happens that the smallest genomic segment that we understand right now is a chromosome. In the future, we could easily imagine features being on small segments that we refer to by hashes. In other words, I think that human-assigned accession or name impedes graph adoption, and using a hash enables graph adoption.

But, you guys have the sequence graph brains... you should decide. (But, I'm skeptical that the right answer is sequence name or accession.)

diekhans: I suspect we will end up with a feature graph on top of of
sequence graphs.  At this point, we should just stick with
the linear assumption.  Previous attempts to integrate 

@reece, I share your enthusiasm for immutable everything.

src/main/resources/avro/sequenceAnnotations.avdl
macieksmuga: Whoops, I now understand Sarah's question. It's a mistake as written, I'm fixing it.

doc/source/api/sequence_annotations.rst
macieksmuga: Should I remove this comment altogether, or just note its aspirational nature?

src/main/resources/avro/sequenceAnnotations.avdl
macieksmuga: I now understand your point, @reece - thanks for elaborating! And yes, the referenceId should exclusively be used. I'd like to propose this (using reference ID as opposed to name) uniformly across all API endpoints as a separate PR, though, as it's a cross-cutting issue.

diekhans:
@dzerbino @sarahhunt 

I created this  issue to thinking about fusions and rearrangements per your comments above.
https://github.com/ga4gh/schemas/issues/594

src/main/resources/avro/sequenceAnnotations.avdl
diekhans: Issue is:
https://github.com/ga4gh/schemas/issues/593

doc/source/api/sequence_annotations.rst
diekhans: remove comment; it is captured in the ticket to resolve this.

macieksmuga:
Please check suggested fixes in #596

src/main/resources/avro/sequenceAnnotations.avdl
```diff
@@ -0,0 +1,137 @@
...
+  // for an attribute distinguished
+  record Attributes {
```
calbach: My initial impression is that this data type is convoluted; by default, I would assume this type of field should follow the same pattern we have everywhere else in the API i.e. `map<[]string>`. In particular, referencing an `ExternalIdentifier` does not seem like a problem unique to sequence annotations. I'd guess this is probably meant to emulate the behavior of GFF3, but given the current documentation I can't tell if this is capturing some semantic nuance in the spec, or if it's overfitting and this is an artifact of it being a file format.

This could continue to follow the normal API patterns we have elsewhere by splitting this into three fields on the `FeatureSet`:

```
map<array<string>> info;
array<ExternalIdentifier> externalIds;
map<array<OntologyTerm>> ___;
```

I don't understand the last use case well enough to propose a name for that field.

diekhans: The definition seems clear to me from a data structure
prospective (map of lists of polymorphic values) but I am
concerned it will be convoluted to use in practice.  This was
patterned on the `map<array<string>> info' approach, only
extending it to multiple types.

VariantSetMetadata, which seems to fill the a similar role, has a
very different structure to either of these approaches.  It's
uses is not clearly documented, I really don't understand it.

I completely agree that this is not limited to
SequenceAnnotations. We needed to have a single way to specify
the extended attributes in a consistent way across the API.

Given the constraints of not having a polymorphic system in avro
or protobuf, I am not sure what is the best way to encode data
types that are fundamental GA4GH types (like ExternalIdentifier
and OntologyTerm), and for that matter int and float, in a
arbitrary key/value storage.  Multiple parallel arrays seems
wrong, although maybe pragmatic.

I will start a ticket on this issue.  However, it would be great
to move ahead on the sequence annotations alpha, as we will
learn my using it.  Any objections?

src/main/resources/avro/sequenceAnnotations.avdl
```diff
@@ -0,0 +1,137 @@
...
+    map<array<union {string, ExternalIdentifier, OntologyTerm}>> vals = {};
+  }
+  
+  /**
+  Node in the annotation graph that annotates a contiguous region of a
+  sequence.
+  */
+  record Feature {
```
sarahhunt: Shouldn't Feature have a name as well as an id?

src/main/resources/avro/sequenceAnnotationmethods.avdl
```diff
@@ -0,0 +1,166 @@
...
+    The annotation set to search within. Either `featureSetId` or
+    `parentId` must be non-empty.
+    */
+    union { null, string } featureSetId;
+
+    /**
+    Restricts the search to direct children of the given parent `feature`
+    ID. Either `featureSetId` or `parentId` must be non-empty.
+    */
+    union { null, string } parentId;
```
sarahhunt: It seems strange that a location must be supplied when a parentId is supplied as the parentId does define a location.

diekhans: @sarahhunt , agreed; although it might make sense to have a different endpoint for getting features give the tree.

sarahhunt: Yes - it would be clearer as a different endpoint.

src/main/resources/avro/sequenceAnnotations.avdl
diekhans: Sarah Hunt <notifications@github.com> writes:
> Shouldn't Feature have a name as well as an id?

Feature name is a somewhat of complex entity to define.  For
instance, an Ensembl gene has both ENSG ids and HGNC names.

Maybe the right thing is to pick the one that gives the most
uniqueness.  In this example, the ENSG.

src/main/resources/avro/sequenceAnnotationmethods.avdl
macieksmuga: Also agree, it seems strange that feature queries need to supply location.

src/main/resources/avro/sequenceAnnotations.avdl
sarahhunt: @diekhans - does that mean you agree to adding a name attribute?

sarahhunt: It would also be good to add an optional md5checksum (as used in the Reference model) to show when non-wiggle features with different ids from different FeatureSets are the same. That's more a new feature than a consistency issue though, so I can put a PR in later.

macieksmuga: @sarahhunt: Can I get a +1 from you with what we have now, and defer the name discussion (important as it is) to a near-future update, or do you feel this should be addressed immediately?

src/main/resources/avro/sequenceAnnotationmethods.avdl
macieksmuga: @sarahhunt - also, is this a blocking issue for this PR, or can we fix this in a future update PR?

src/main/resources/avro/sequenceAnnotations.avdl
sarahhunt: @macieksmuga - yes - happy to +1 now and address the other details later.

saupchurch:
+1
looks good from our perspective - couple of minor issues raised can be taken to follow up issues for discussion

doc/source/api/index.rst
reece: Hmm. Okay, this is probably just means that you and I (and maybe the larger we) have different colloquial definitions of "annotations" and "features". There's probably no right or wrong here, but we should get some consensus on how the terms are used. (Ditto for the use of "allele" and "variant" semi-interchangeably elsewhere in the schema.)

doc/source/schemas/sequenceAnnotationmethods.rst
reece: Again, I think we differ in our use of "annotation" and "feature". FWIW, I think of annotations as information attached to entities in toto, e.g., a set of GO terms associated with a specific sequence, and features as information attached to specific regions of sequences. In that usage, the domain and range of features are much narrower than that for annotations (which can basically be anything attached to annything).

reece: Okay, I realize that I might be fixating, but one more observation in this regard (until the next one):

doc/source/api/sequence_annotations.rst sayeth:

> The central object of the GA4GH Sequence Annotation API is a Feature.

I think this underscores the point I'm trying to make. If the core concept is a Feature, then why isn't this the Sequence Feature API?

src/main/resources/avro/sequenceAnnotations.avdl
reece: @macieksmuga Side bar: When stitching together paths of hashed subsequences, it would be useful if there were a hash function, h, and a hash composing function, g, such that g(h(s1),h(s2)) = h(s3) for sequences s1, s2, and s3 = concat(s1,s2). In other words, the hash of the containing sequence could be computed from hashes of the subsequences (without fetching the sequences themselves).

I don't have the background for the solution, so I asked the crypto folks on stackexchange here: 
https://crypto.stackexchange.com/questions/24622/

reece:
+1.  This code seems completely appropriate for advancing development.

doc/source/schemas/sequenceAnnotationmethods.rst
diekhans: @reece because Feature (discrete annotations) are not going to
be the only type of sequence annotation.  There is also be
contiguous (wiggle).  There is a proposed data structure,
however, it's not implemented and hence not included in this
release.

doc/source/api/index.rst
diekhans: Larger we. Hence `Annotation' should always be qualified
`Sequence Annotations', `Variant Annotations', etc.  I had never
heard variant classification call `variant annotation' until I
showed up at the `Annotation Task Team' call and discovered it
wasn't my type of annotations.

diekhans: @reece, see we didn't make this up:

http://www.nature.com/subjects/sequence-annotation
http://www.uniprot.org/help/sequence_annotation
http://ensemblgenomes.org/info/data/annotation
https://seqannotationstudio.codeplex.com/
https://en.wikipedia.org/wiki/Genome_project#Genome_annotation
http://www.amazon.com/Annotation-Chapman-Mathematical-Computational-Biology/dp/1439841179

Issue #589, opened 2016-03-29
TranscriptEffect CDSLocation naming consistency and feature


diekhans:
"CDSLocation" should be named "cdsLocation" to be consistent with the rest of the API.

A featureId of the CDS feature would provide useful context

sarahhunt:
@diekhans - #600 updates the name in the schema. 

The TranscriptEffect record is deliberately minimal at the moment. The idea is that information which can be derived from the Transcript Feature should not be included. It will be easier to see how well this works when the initial sequence annotations are available, so we can revisit.

PR #600, opened 2016-04-05
change cdsLocation capitalisation for consistency

sarahhunt:
Change attribute name as per #589


david4096:
Can we close given https://github.com/ga4gh/schemas/blob/master/src/main/proto/ga4gh/allele_annotations.proto#L115 ?

PR #601, opened 2016-04-05
add Variant Annotation Task Team role to README

sarahhunt:
Text copied from developers site.


david4096:
@sarahhunt I'm not sure why the checks failed here but I've updated with the same commit in PR #657 . Thanks!

Issue #568, opened 2016-02-29
Choose consistent time format


reece:
#552 implemented ISO8601 for the variation api. CallSet, Variant, and ReadGroup have "created" and "updated" fields, but these were unchanged.

This proposal is switch remaining timestamps to ISO8601 and rename fields to createdDateTime and updatedDateTime for consistency.

@jeromekelleher, @david4096, @diekhans, @andrewjesaitis : Comments appreciated since you three probably know which monsters this will tickle in our code.

mbaudis:
+1

@reece Probably no monsters to be woken, since these attributes haven't been really used.

jeromekelleher:
Here are my thoughts:

1. We should use iso8601 everywhere for these timestamps;
2. I don't think this will cause any implementation hassles (other than actually doing it);
3. I'm not convinced by the need for the name change to `createdDateTime`. What's wrong with `created` and `updated`? We don't feel the need to put the types into the names for other fields (startLong, nameString), so why put them in here? I realise I'm probably joining the party late on this one though, so sorry if we're going over old ground on this argument!

@calbach, what's your thoughts on this, are you OK with the shift to ISO8601?

reece:
I'm happy with any of the likely outcomes that promotes consistency in type and naming. 

My weak preference is ISO8601 and updated/created.

(I had suggested DateTime previously, but only as a reaction to updateTIme, which seemed incorrectly suffix typed to me.)

mbaudis:
@jeromekelleher While I'm fine with ```created``` | ```updated```, this was seen as ambiguous in var. discussions ("updated" could be a Boolean/flag) & so @diekhans switched it to the current.

I had started the whole ISO8601 switch & [documentation](http://ga4ghdata.org/Sites/github/mbaudis/biosample/html/api/metadata.html#date-and-time-format-specifications), but very explicitly called those ```recordCreateTime``` | ```recordUpdateTime``` ... But as said, all fine with me, since the documentation should make it clear.

diekhans:
My thinking on *DateTime isn't they type (which was Unix time,
now ISO time), but clarity on the meaning of the field.
`created' sounds like a boolean to me, but what ever, as long as
it is consistent.

jeromekelleher:
I think in the interest of keeping things simple we should maintain 'created' and 'updated', as changing the names will cause friction for existing implementations. I think the advantages of `createdDateTime` over `created` are pretty subjective.

diekhans:
[...]

PR #606, opened 2016-04-05
Biosamples and individuals

david4096:
The biodata protocol allows read groups and call sets to be related to samples and individuals. This is done by adding `bioSampleId` fields to both of those records, and allowing them to be filtered using `bioSampleId` in [SearchReadGroupSetsRequest](https://github.com/ga4gh/schemas/blob/biosample/src/main/resources/avro/readmethods.avdl#L124) and [SearchCallSetsRequest](https://github.com/ga4gh/schemas/blob/biosample/src/main/resources/avro/variantmethods.avdl#L171). In addition to being able to filter callsets and readgroups, the new `biosamples` and `individuals` endpoints are added.

Individuals can be searched by name, and biosamples can be filtered by individualId and name.

Given some read find which individual it derives from:
1) Read the `bioSampleId` from a `ReadGroup` record.
2) Get the `BioSample` record by performing a get at `http://ga4gh/biosamples/<bioSampleId>`.
3) Read the `individualId` from the `BioSample` record.
4) Get the `Individual` record by performing a get at `http://ga4gh/individuals/<bioSampleId>`.

Given some individual find reads:
1) From the `Individual` record get the `individualId`.
2) Create a `SearchBioSamplesRequest` with the provided `individualId`.
3) Perform a `SearchReadGroupSetsRequest` with each returned `bioSampleId`.
4) Only read groups matching the bioSampleId will be returned. Perform a SearchReadsRequest for each read group in the `SearchReadGroupSetsResponse`. 

A reference implementation can be seen [here](https://github.com/ga4gh/server/pull/1028).


dzerbino:
This PR covers a lot of ground. Two main issues are being addressed: ontology terms and biosamples. 

Separating the concerns into two PRs (with biosamples depending on ontology terms) would help the review  process.

diekhans:
@dzerbino OntologyTerm is already in master.  While Ontology queries will need work in general, separating out doesn't seem helpful for this PR.

diekhans:
+1  while the sequence metadata needs to evolve based on developing use cases, it's vital that we get some of it in for the next alpha to have rounded functionality.  

Can we get some engagement on this ASAP as being good enough for alpha

mbaudis:
Just a question regarding the camel case: I thought that it is

* Record <= lc(Record) + Id
* e.g. ```BioSample``` <= ```biosampleId``` (not ```bioSampleId```)

But I may be confused here by the ```DataSet``` vs. ```Dataset``` discussions?

diekhans:
I don't think this is formally defined, however my impression
is it's just lower casing the first letter, not the string.

However, there are a lot of inconsistencies on the schema.

src/main/resources/avro/biodatamethods.avdl
```diff
@@ -0,0 +1,147 @@
...
+protocol BiodataMethods {
+
+import idl "metadata.avdl";
+import idl "common.avdl";
+import idl "methods.avdl";
+import idl "biodata.avdl";
+
+/******************  /individuals/search  *********************/
+/**
+This request maps to the body of `POST /biosamples/search` as JSON.
```
sarahhunt: /biosamples/search => /individuals/search

src/main/resources/avro/biodatamethods.avdl
```diff
@@ -0,0 +1,147 @@
...
+*/
+record SearchIndividualsRequest {
+  /**
+  The dataset to search within.
+  */
+  string datasetId;
```
sarahhunt: DatasetId is required in SearchIndividualsRequest, but Individual records do not have datasetIds. Can the relationship be clarified?

src/main/resources/avro/biodata.avdl
```diff
@@ -0,0 +1,133 @@
...
+   */
+   union { null, string } description = null;
+
+   /**
+   Disease annotation of the sample.
+   */
+   union { null, OntologyTerm } disease = null;
+
+   /**
+   The :ref:`ISO 8601<metadata_date_time> time at which this BioSample record
```
sarahhunt: Missing backtick

src/main/resources/avro/biodata.avdl
```diff
@@ -0,0 +1,133 @@
...
+   was created.
+   */
+   string createDateTime;
+
+   /**
+   The :ref:`ISO 8601<metadata_date_time> time at which this BioSample record was updated.
```
sarahhunt: Missing backtick

src/main/resources/avro/biodata.avdl
```diff
@@ -0,0 +1,133 @@
...
+   union { null, string } name = null;
+
+   /**
+   The biosample's description. This attribute contains human readable text.
+   The "description" attributes should not contain any structured data.
+   */
+   union { null, string } description = null;
+
+   /**
+   Disease annotation of the sample.
```
sarahhunt: Attaching disease to BioSample seems odd - many sequenced samples are not associated with a disease; if they were it would be more normal to flag this at the individual level. 
This model does not allow for individuals with multiple diseases. 
Disease is a limiting term - what about other traits which may be investigated?
Wouldn't this be better modelled as a record linking an individual to a phenotype/trait/disease with date of linkage?

src/main/resources/avro/biodata.avdl
```diff
@@ -0,0 +1,133 @@
...
+  molecules (e.g. genomic DNA, RNA, proteins) for molecular analyses (e.g.
+  sequencing, array hybridisation, mass-spectrometry) are extracted. Examples
+  would be a tissue biopsy, a single cell from a culture for single cell genome
+  sequencing or a protein fraction from a gradient centrifugation.
+  Several instances (e.g. technical replicates) or types of experiments (e.g.
+  genomic array as well as RNA-seq experiments) may refer to the same BioSample.
+  In the context of the GA4GH metadata schema, BioSample constitutes the central
+  reference object.
+ */
+ record BioSample {
```
sarahhunt: An optional tissue type attribute would be useful. How can tumour and normal samples from the same individual be distinguished?

src/main/resources/avro/biodata.avdl
```diff
@@ -0,0 +1,133 @@
+@namespace("org.ga4gh.models")
+
+/**
+This protocol defines the "biodata" objects, which can be considered data
+representations of biological correlates.
```
sarahhunt: What is a biological correlate?

sarahhunt:
Doesn't the Experiment model need to be updated? It is described as ' An experimental preparation of a sample.' so it is strange that it does not have a biosample id.

src/main/resources/avro/biodata.avdl
mbaudis: This is just ill documented. "The primary disease associated with this BioSample." It is actually correct in having the one specific to this sample ("colorectal adenocarcinoma Duke III of the right colonic flexure"), while having the multitude of possible phenotypes/diseases the individual can have with ```Individual```.
However, there are several unsolved questions with this. Chiefly, most diseases cannot be described (yet) using a single OntologyTerm. It is absolutely clear that this is an interim solution.
I recommend that you (and everybody else) has a look and contribute at the ["BioFeature" object proposal](https://docs.google.com/document/d/1aTK-hWdTPbSt3DhB-pW8jDvK5-krJW557Vsw6V3jG-s/edit#heading=h.w8b0lsmtgmwl).

src/main/resources/avro/biodata.avdl
```diff
@@ -0,0 +1,133 @@
...
+*/
+
+protocol Biodata {
```
sarahhunt: The name Biodata is not very precise - everything the API currently deals with is Biodata. Biosample may be better?

src/main/resources/avro/biodata.avdl
mbaudis: Correcting.

src/main/resources/avro/biodatamethods.avdl
mbaudis: Correcting.

src/main/resources/avro/biodata.avdl
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+    For a representation of an NCBI Taxon ID as an OntologyTerm, see
+    NCBITaxon Ontology
+      http://www.obofoundry.org/ontology/ncbitaxon.html
+    For example, 'Homo sapiens' has the ID 9606. The NCBITaxon ontology ID for
+    this is NCBITaxon:9606, which has the URI
+    http://purl.obolibrary.org/obo/NCBITaxon_9606
+    */
+    union { null, OntologyTerm } species = null;
```
sarahhunt: This is different to Reference which requires an ID (e.g. 9606->human).  Is the intention to update Reference?

src/main/resources/avro/biodata.avdl
mbaudis: Correcting.

src/main/resources/avro/biodatamethods.avdl
```diff
@@ -0,0 +1,147 @@
...
+  /**
+  The continuation token, which is used to page through large result sets.
+  Provide this value in a subsequent request to return the next page of
+  results. This field will be empty if there aren't any additional results.
+  */
+  union { null, string } nextPageToken = null;
+}
+
+/**
+Gets a list of BioSamples accessible through the API.
```
sarahhunt: BioSamples => Individuals?
Is this comment needed?

src/main/resources/avro/biodata.avdl
mbaudis: There are many additional attributes which are/have been discussed (see e.g. the metadata branch), but were left out for now. Primary intention here was to get the ```BioSample``` | ```Individual``` records in, for referencing and usage driven development (not my choice).

sarahhunt: Thanks @mbaudis - good to know!

src/main/resources/avro/biodata.avdl
mbaudis: Adding "(e.g. concrete tissue samples, individuals)"

src/main/resources/avro/biodata.avdl
mbaudis: ```BioSample``` is a record name, and "Biodata" contains more than this record type.

mbaudis:
@sarahhunt Regarding the ```Experiment```: Yes, there were changes omitted which had been done somewhere else.

mbaudis:
I've addressed some of @sarahhunt 's easier solvable points & removed some backup plot files from the _static dir.

src/main/resources/avro/biodatamethods.avdl
david4096: I had added datasetId in a previous pull request, but it was requested to leave `datasetId` out of the record itself so a sample or individual could belong to multiple datasets. https://github.com/ga4gh/schemas/pull/586

src/main/resources/avro/biodata.avdl
david4096: Added an issue to regarding this https://github.com/ga4gh/schemas/issues/610

david4096:
Added an issue regarding the Experiment model https://github.com/ga4gh/schemas/issues/611

src/main/resources/avro/biodata.avdl
david4096: Added an issue to address this: https://github.com/ga4gh/schemas/issues/612

src/main/resources/avro/biodatamethods.avdl
diekhans: Although the dataset conceptual model is fuzzy, I believe the
intention is that objects can only belong to one data set.  If
there is a use case for objects belonging to multiple datasets,
we need to revisit this model ASAP.

dzerbino: I don't understand the logic for deciding that an object belongs to only one dataset... If, say, you wish to specify which individuals you did your GWAS on, what object would you use? A dataset no? And we definitely would want individuals to be re-used in multiple cohorts.

src/main/resources/avro/biodata.avdl
dzerbino: Still, that doesn't make Biodata a good name. May I suggest Specimen, which describes both individuals and samples?

doc/source/api/ontologies.rst
```diff
@@ -0,0 +1,218 @@
...
+=====================
+
+.. _ciliopathy: http://www.ensembl.org/Multi/Search/Results?q=ciliopathy
+.. _ciliopathies: http://www.ensembl.org/Multi/Search/Results?q=ciliopathy
+.. _`orofaciodigital syndrome`: http://www.ensembl.org/Homo_sapiens/Variation/Explore?r=1:200650139-200651139;v=rs587777067;vdb=variation;vf=64599886
+
+* SNP **rs587777067** is associated with `orofaciodigital syndrome`_
+* the query for "ciliopathy_" (or "ciliopathies_") returns two different results in the current Ensembl representation
+* neither of these represents all known ciliopathies or returns rs587777067, as these queries are not performing a lookup for child terms of a functional class ciliopathy, but rather represent text queries for the two lexical variants of the term
+
```
dzerbino: These issues are currently being fixed by Ensembl and should disappear in the next release. This comment will be irrelevant in a couple of months, which is for the purpose of this document a good and a bad thing ;)

src/main/resources/avro/biodatamethods.avdl
diekhans: Nope, the purpose of dataset is to provider a bag for ownership
of objects, not grouping for analysis.  So it would not be used
to identify the individuals one did GWAS on.  Its more akin to
an NCBI BioProject.  Google uses it for billing and access
control.

A dataset can reference objects in a another dataset.  It's not
a wall.

Although the GWAS is an interesting use cases for the API.

@calbach dataset is still not clear to many people (I am not
even sure I have it right); could someone work on improving the
documentation?

src/main/resources/avro/biodata.avdl
diekhans: Daniel Zerbino <notifications@github.com> writes:
> Still, that doesn't make Biodata a good name. May I suggest Specimen, which
> describes both individuals and samples?

This is intended to contain the metadata for all of the
biological stuff that exists outside of the computer.  It will
end up with more objects than are currently there (like sequencing experiements).

I can see that Biodata maybe not the best name, however specimen
doesn't capture it either.

dzerbino: Now you have me confused:
- sequencing datasets are not a biological object
- don't they have the read api for this?

I persist that the prefix Bio is purely decorative, and naming an API
'data' is up there in coding sins along with one letter variable names.

src/main/resources/avro/biodatamethods.avdl
dzerbino: Wait how does "A dataset can reference objects in a another dataset.  It's not
a wall." fit with unique ownership of datasets?

Also, in that case, dataset should be called DataSubmission, since that is indeed unambiguous. 

Sorry, I have decided to wage a war against fuzzy terminology this morning.

diekhans: Reference is not ownership.  VCFs in one dataset can reference BAMs in another.  For instances, if you recall 1K genomes variants doesn't mean you put them in the same dataset.

DataSubmission doesn't capture it either because a dataset may not be submitted at one time
(at least there is restriction that it must be).

I think clear terminology is very important.   However, DataSet has been around since the very first version of the API.   So change the term is a bit late.  Instead, improve the documentation.
Lets start a ticket on:
  https://github.com/ga4gh/schemas/blob/master/doc/source/api/apidesign_intro.rst#dataset

diekhans:
@dzerbino  words of wisdom on the naming of the module was lost due to being inside of the code review and it being pushed.  I would agree that BioData is not a clear name.  Here is how it came about:

- _metadata _ was becoming a monolithic dumping group for everything that didn't have a clear conceptual bag across the whole API.  There was a clear conceptual grouping for everything pre-computer data file (individual, sample, library prep, sequencing machine).
- there metadata actually isn't clear; one person's metadata is a another person's data.  Clinical features were being called_ metadata._ because they were not files with sequence in them.

So, suggestions on a clearer module name? 

 `experimental'?   A common use for the white coat stuff, but we computer people also think we do experiments.

dzerbino:
IMHO, there is a clear semantic distinction between the objects of study (individual and sample, collectively I would call them specimen) from the experimental method (library prep, sequencing machine, collectively I would call those protocol). 

They would be used differently: from an analysis point of view, the former could be considered data or metadata depending on circumstances, the latter not really. I would therefore define two similar-ish schemas, 'Specimen' and 'Protocol'. If really they must to be together, how about 'LabNotebook' or 'LIMS'?

diekhans:
@dzerbino  Yes, I agree with your semantic distinction and that should be captured in the  object classes.  This is the next level of grouping, trying to make it a reasonable sized bucket.

LabNotebook' or 'LIMS' doesn't cut it as a term, but does kind of have the reasonable scope.

mbaudis:
@dzerbino These discussions have happened before, but are not easily traceable (after some months of TCs + doc editing, people, including me, write down only the outcome but not the arguments leading to it.

* "metadata": I coined the phrase "everything but the sequence", provocatively, for GA4GH San Diego 2014, and it was presented like that. The problem was that it wasn't taken seriously.
* "biodata" | "metadata" | "protocol": As above, metadata is everything but the reads themselves; I and others have made the point that "Metadata" is not a GA4GH project / branch, but a knowledge pool supporting schema development:

    *Meta data means everything but the sequence and provides meta data representation which all task teams can consume. It includes meta data for individuals, samples, analyses, instrumentation and ontology representation for meta data.*

Typical cases for "domain transgressing" metadata objects are:

* OntologyTerm
* format definitions (e.g. ids, time)
* object references

I don't really care for the "biodata" vs. "metadata"; this is more an effort to have things directly interfering with sequence/variant annotations from the stuff that is being analyzed, mostly to placate the seq/ref/var developers. In my opinion: Everything but the sequence, and everybody can shape it. 

Also, it really doesn't matter how things are named at Google or EBI.

diekhans:
Many of use thing that everything but sequence isn't a very
useful definition, since there are things that most people don't
think of as metadata fall into it. Examples are gene
annotations, clinical records, G2P evidence.

I don't know what the right breakdown is.  I am not crazy about
biodata either, and I may have suggested it.  But this is just
a container, not the objects.  

The goal for this bit was to get the metadata related to the
stuff outside of computation data released.

How about `experimentalMetdata.avdl???

mbaudis:
That's what I mean: "Everything" means everybody has to deal with it, not just a "Metadata Task Team" - but this should coordinate (and have the say so) etc.
There is no mead to change labels here. "Bioobjects" would be my preference for Individual, BioSample, but there may have been arguments before (object description vs. the real thing...). I usually talk about "technical" and "biological" metadata. But there is no perfect solution, and the current one is as good as it gets, with the usual margin of error.

diekhans:
Absolutely! metadata should not be a throw-in-over the wall because it's no
fun.

The biodata/bioobjects/experimentalMetadata is kind of the stuff
that falls out side of the task teams, because they are all
computer data oriented.

I don't have great ideas for names here...

mbaudis:
@dzerbino @diekhans It is practically impossible to have both the general concept of metadata as well as also  "metadata as separate objects" in the same bucket, without stirring up trouble.
Also, we have had these discussions about the placement of the technical objects (```Experiment```, ```Analysis```) before, and also an interlude when ```OntologyTerm``` moved to its own document ...
And although @diekhans is killing "common" (I think), IMHO it makes sense to have a place for structural objects which are used in various places.
The cleanest way to separate objects seems to me to have a three-way split of "metadata" objects:

* metadata_common (```OntologyTerm```, ```"GeoObject"```, ...)
* metadata_procedures (```Experiment```, ```Analysis```, ...)
    * we could use "metadata_protocols" here (akin to @dzerbino's note), but "protocols" may have a specific use already (queries ...)?!
* metadata_bioobjects (```BioSample```, ```Individual```, ...)

However, this is more for structural/educational/documentation purposes; technically it shouldn't matter ... I am not pushing this, but those split | unify | relabel discussions are coming up again and again. I am happy with metadata.avdl + biodata.avdl (or .proto); we just have to settle & somebody has to put a lid on it.

dzerbino:
@mbaudis: I feel that you are underplaying the structural/educational/documentation purposes: if this API makes no sense, it will not be used and we are wasting our time. In particular, it would appear that past discussions/TCs etc did not succeed to produce a clear written rationale for their decisions, therefore they will go on to be questioned. I don't think it is possible to put a lid on this discussion on pure authority. 

Splitting the schema per se is not necessary, but the descriptions and explanations need to be significantly improved. When I read that "biodata" objects are data representations of biological correlates, this makes no sense to me.

mbaudis:
@dzerbino You are certainly invited to write a more extensive documentation! However, this also exists in the API doc for biodata (which is only not strongly defined and doesn't have to be; as you said, splitting the schema isn't necessary per se).

The documentation for the "biodata" objects is at the usual place in https://github.com/ga4gh/schemas/blob/biosample/doc/source/api/biodata.rst; there is a a recent formatted version at http://ga4ghdata.org/Sites/github/ga4gh/biosample/html/api/biodata.html.

And again, you are very welcome to make concrete contributions to the documentation. The reasoning for the objects has been extensive, and the Individual/BioSample concept had been presented in DWG calls. What was lost is the trail of rejected alternatives (in naming etc.).

mbaudis:
@dzerbino ... also, I didn't say "I want to put a lid on this", but "somebody"; this was a call for positive contributions with the aim to merge in a state of maximum achievable agreement

 But ASAP, and we have  to move on; this PR just should provide some structural ground for adding usable extensions & implementations.

doc/source/api/ontologies.rst
diekhans: This is not useful documentation; please remove it.  This stuff is really complex and pointing out bug in someones system doesn't help.

dzerbino:
Ultimately we are finessing about presentation, not fundamentals.

I propose to: 
- rename the schema Experimental_Metadata (or something reasonably synonymous)
- replace "This protocol defines the "biodata" objects, which can be considered data representations of biological correlates. get rid of "biological correlates" (critically, many biological things correlate, and we don't them in our databases! ;-) with "This schema defines the biological specimen that were assayed and the practical means by which they were assayed."
- Add the above description in the documentation, where the objects but not the schema are defined.

diekhans:
@dzerbino I like your suggestions.  Presentation is very
important, as enables understanding.

mbaudis:
@dzerbino Thanks for the contributions! My take:

* I'm all for exchanging "biodata", but "experimental..." doesn't work semantically (sounds like "immature", and anyway it was noted before that experiments are performed on data etc., too). So, more suggestions, everyone, please?
* I'm not sure about this misunderstanding regarding (the noun) "correlate(s)" and "to correlate" or "correlation". "biological correlate" is a valid expression which I, personally, would not conflate with "correlating something". So, the original sentence seems al least semantically correct but, by itself, insufficient.

That being said, your suggestion seems very clear, and covers both objects and procedures ("This schema defines the biological specimen that were either directly assayed and the practical means by which they were assayed."). However, this would then not only cover ```BioSample``` and ```Individual```, but also ```Experiment``` and ```Analysis```. I actually would like to follow the meaning of your sentence, and glob those objects together, with the "metadata_common" (Dataset, OntologyTerm, "GeoObject", ...) or ..._shared or ..._base (those are the ones embedded/required by various kinds of GA4GH objects) kept separate.

Can we do it like that, with @diekhans deciding about the final file names & object distribution (since this is very much related to the general schema layout)?

dzerbino:
Sounds like a plan. @diekhans what do you think?

dzerbino:
How about, instead of experimental, assay_metadata, which covers both the method and the instance of assays.

diekhans:
This sounds like a good approach.  Improving the documentation
is key.  If @dzerbino finds it perplexing, it's only going to be
harder for people father away from GA4GH TTs.

Just `experimental' isn't great, however `experimentalMetadata'
is close to common thinking in my experience.  

assay_metadata, sounds a bit too narrow, since this include individual.
Not sure, but a good way to describe it.

The goal was this module cover everything from Individual object
to the Experiment object.  Essentially the stuff `outside' of
data access via the DWG.

Hmm, perhaps `whitecoat' ??

helenp:
Summary of actions from phone call 21/4/2016 with @dzerbino, @diekhans @mbaudis :
1. Split into 
BioMetaData - all meta data relating to Individual or BioSample
AssayMetaData - all meta data relating to how assays/expts/data were generted
2. DataSet is retained as a container (after investigation of use in other contexts) - membership of DataSets =1

@mbaudis will do the refactor as above
@diekhans will review once complete
@zerbino will review docs which need work - more detail and examples are required.

Attributes as requested by @sarahhunt  will be subject of a subsequent PR

src/main/resources/avro/biometadata.avdl
```diff
@@ -0,0 +1,131 @@
+@namespace("org.ga4gh.models")
+
+/**
+This protocol defines the :ref:`"biological metadata" objects<biometadata>`, which can be considered data representations of concrete biological samples or entities (e.g. the DNA extracted from a tissue sample, a human individual).
+*/
+
+protocol Biometadata {
```
diekhans: should be BioMetadata to match other schemas.

src/main/resources/avro/assaymetadata.avdl
```diff
@@ -0,0 +1,146 @@
+@namespace("org.ga4gh.models")
+
+/**
+This protocol defines metadata objects describing experimental or analytical procedures, performed on biological objects (e.g. biosamples) or readouts from experimental procedures.
+*/
+
+protocol Assaymetadata {
```
diekhans: should be AssayMetadata

src/main/resources/avro/assaymetadata.avdl
```diff
@@ -0,0 +1,146 @@
...
+
+  import idl "common.avdl";
+  import idl "metadata.avdl";
+
+  /**
+  An analysis contains an interpretation of one or several experiments.
+  (e.g. SNVs, copy number variations, methylation status) together with
+  information about the methodology used.
```
diekhans: wording subject to misinterpretation this is only the provenance, it does not contain the actual interperatation

src/main/resources/avro/assaymetadata.avdl
```diff
@@ -0,0 +1,146 @@
...
+
+    union { null, string } name = null;
+
+    union { null, string } description = null;
+
+    /**
+    The time at which this record was created.
+    Format: :ref:`ISO 8601 <metadata_date_time>`
+    */
+    union { null, string } createDateTime = null;
```
diekhans: changed the field name, but not the type to makes existing objects.  Adding  *DataTime didn't really make the meaning clearer.  That is `created'

mbaudis: This went through n iterations. The most verbose was ```recordCreateTime``` I think, explicitly scoping the meaning to the record, not the described object (i.e. record of a biosample, not sampling time).
To recap:
* ```created``` | ```updated``` were considered insufficient since ambiguous regarding the expected value
* same for ```recordCreated``` ..., which I had suggested before (there [had been a discussion](https://github.com/ga4gh/schemas/issues/568#issuecomment-190342645))
* not sure about the correct tense ("created" vs. "create", "updated" vs. "update")
* options
    * ```createDateTime``` | ```updateDateTime```
    * ```createdDateTime``` | ```updatedDateTime```
    * ```creationDateTime``` | ```updateDateTime```
    * ```recordCreationTime``` | ```recordUpdateTime```
    * ```timeRecordCreated``` | ```timeRecordUpdated ```
    * ```timeCreated``` | ```timeUpdated```

This should be settled ASAP, schema wide. I have no problems with ```createDateTime``` | ```updateDateTime``` myself, but also wouldn't complain about one of the other examples (or something similar).

src/main/resources/avro/assaymetadata.avdl
mbaudis: Changed. Also ```AssayMetadataMethods```.

src/main/resources/avro/biometadata.avdl
mbaudis: Changed. Also ```BioMetadataMethods```.

src/main/resources/avro/assaymetadata.avdl
mbaudis: ```Analysis``` isn't defined very well, but thought to be a relevant part of the record structure. I've stripped down the description even further, but invite others to contribute.

src/main/resources/avro/assaymetadata.avdl
diekhans: Michael Baudis <notifications@github.com> writes:
> This went through n iterations. The most verbose was recordCreateTime I think,
> explicitly scoping the meaning to the record, not the described object (i.e.
> record of a biosample, not sampling time).

recordCreateTime is clearer than the current `created' or
`createDateTime'.  In the other objects, it appears that it's
intention is like unix st_ctime, but that is only understood by
oral tradition.

However this is just a plea to make it consistent an then go
back an fix it schema wide.   

> This should be settled ASAP, schema wide.

Nothing happens ASAP in GA4GH.

mbaudis:
@david4096 +1 from me, now. Not sure what else to do? Still Travis complaints, though.

kozbo:
The plan is to merge this after the changes for RNA have been merged. Of course we need the CI tests to pass too :-)

david4096:
@mbaudis Updated the branch to pass tests. There was a name change that needed to be reflected in the reads and annotations descriptors.

david4096:
Closing in favor of https://github.com/ga4gh/schemas/pull/636

Issue #593, opened 2016-04-01
ReadAlignment links to Reference by name rather than id


diekhans:
The design pattern in the API is that object linkage is done by id, not by
name.  

The Position object uses referenceName rather than referenceId.

This should be corrected or the rationale for this inconsistenty strongly documented

david4096:
@mbaudis @ejacox @kozbo It would be great to land on one side or the other here before release. Just use reference names seems like it would make people the happiest at this point. Ideas?

Having a different way of specifying position from one protocol to the next is really confusing. To close this we might remove the alternative `reference_id` pattern from SearchVariantAnnotations as well.

ejacox:
Does this fit within our external ids discussion. The reference_name is the identifier used within a particular reference (hg38?). In that view, reference_name is fine if we indicate somewhere what the reference is.

david4096:
@dcolligan @delagoya ? opinions? We can close this by removing reference ID from SearchReadsRequest and SearchVariantAnnotationsRequest, which seems align with @richarddurbin's comments here https://github.com/ga4gh/schemas/pull/616. We would then also merge language like https://github.com/ga4gh/schemas/pull/732 to enforce that references are uniquely named in a reference set.

I believe it makes it easiest to work with some genomics data if only the reference name is required for search, since you technically don't have to have a reference set local to your instance. `1`, `2`, `3`, are fairly portable.

If we choose IDs we should implement searching references by name https://github.com/ga4gh/schemas/pull/665. It is nice to have in either case.

Note, 23andMe opted to use accession_id when specifying a reference for range searches https://api.23andme.com/docs/reference/, which falls somewhere between using reference names and server generated identifiers.

PR #616, opened 2016-04-27
Change referenceName to referenceId

david4096:
When performing searches on the Variants or Features endpoints the `referenceName` field allows the client to define which contig to search on. This is problematic because reference names are not required to be unique, and it is not immediately clear how to form the "referenceName" field. Is it "1" or "chr1"? This is further complicated by the fact that references can belong to multiple Reference Sets.

By using the `referenceId` in its place we ensure that each request is against a specific reference and we get better guarantees about the relationships in the data. The Reads API properly makes this distinction:

https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/readmethods.avdl#L26

### DIFF
--- a/src/main/resources/avro/alleleAnnotationmethods.avdl
+++ b/src/main/resources/avro/alleleAnnotationmethods.avdl
@@ -34,20 +34,20 @@ record SearchVariantAnnotationsRequest {
   string variantAnnotationSetId;
 
   /**
-  Only return variants with reference alleles on the reference with this
-  name. One of this field or `referenceId` or `features` is required.
+  Only return annotations with reference alleles on the reference with this
+  name. One of this field or `referenceId` is required.
   (case-sensitive, exact match)
   */
   union { null, string } referenceName = null;
 
   /**
-  Only return variants with reference alleles on the reference with this
-  ID. One of this field or `referenceName` or `features` is required.
+  Only return annotations with reference alleles on the reference with this
+  ID. One of this field or `referenceName` is required.
   */
+  
   union { null, string } referenceId = null;
 
   /**
-  Required if referenceName or referenceId supplied.
   The beginning of the window (0-based, inclusive) for which variants with
   overlapping reference alleles should be returned.
   Genomic positions are non-negative integers less than reference length.
@@ -57,7 +57,7 @@ record SearchVariantAnnotationsRequest {
   long start;
 
   /**
-  Required if referenceName or referenceId supplied.
+  Required if referenceName or referenceId is supplied.
   The end of the window (0-based, exclusive) for which variants with
   overlapping reference alleles should be returned.
   */
--- a/src/main/resources/avro/readmethods.avdl
+++ b/src/main/resources/avro/readmethods.avdl
@@ -20,11 +20,18 @@ record SearchReadsRequest {
   array<string> readGroupIds;
 
   /**
-  The reference to query. Leaving blank returns results from all
-  references, including unmapped reads - this could be very large.
+  The reference to query. Omitting both referenceName and referenceId
+  returns results from all references, including unmapped reads - 
+  this could be very large.
   */
   union { null, string } referenceId = null;
 
+  /**
+  The reference to query. Omitting both referenceName and referenceId
+  returns results from all references, including unmapped reads - 
+  this could be very large.
+  */
+  union { null, string } referenceName = null;
   /**
   The start position (0-based) of this query.
   If a reference is specified, this defaults to 0.
--- a/src/main/resources/avro/sequenceAnnotationmethods.avdl
+++ b/src/main/resources/avro/sequenceAnnotationmethods.avdl
@@ -84,8 +84,15 @@ protocol SequenceAnnotationMethods {
     /**
     Only return features on the reference with this name 
     (matched to literal reference name as imported from the GFF3).
+    One of referenceName or referenceId is required.
     */
-    string referenceName;
+    union { null, string } referenceName = null;
+    
+    /**
+    Only return features on the reference with this ID
+    One of referenceName or referenceId is required.
+    */
+    union { null, string } referenceId = null;
 
     /**
     Required. The beginning of the window (0-based, inclusive) for which
--- a/src/main/resources/avro/sequenceAnnotations.avdl
+++ b/src/main/resources/avro/sequenceAnnotations.avdl
@@ -62,9 +62,8 @@ protocol SequenceAnnotations {
 
     /**
     The reference on which this feature occurs.
-    (e.g. `chr20` or `X`)
     */
-    string referenceName;
+    string referenceId;
 
     /**
     The start position at which this feature occurs (0-based).
--- a/src/main/resources/avro/variantmethods.avdl
+++ b/src/main/resources/avro/variantmethods.avdl
@@ -76,8 +76,17 @@ record SearchVariantsRequest {
   */
   union { null, array<string> } callSetIds = null;
 
-  /** Required. Only return variants on this reference. */
-  string referenceName;
+  /** 
+  Only return variants for the given reference. One of
+  referenceId or referenceName is required.
+  */
+  union { null, string } referenceId = null;
+  
+  /** 
+  Only return variants for the named reference. One of
+  referenceId or referenceName is required.
+  */
+  union { null, string } referenceName = null;
 
   /**
   Required. The beginning of the window (0-based, inclusive) for
--- a/src/main/resources/avro/variants.avdl
+++ b/src/main/resources/avro/variants.avdl
@@ -202,7 +202,7 @@ record Variant {
   The reference on which this variant occurs.
   (e.g. `chr20` or `X`)
   */
-  string referenceName;
+  string referenceId;
 
   /**
   The start position at which this variant occurs (0-based).


diekhans:
+1, this is a big inconsistency in the API.

This is related to:  https://github.com/ga4gh/schemas/issues/593

richarddurbin:
Within a referenceSet the referenceNames need to be unique, of course - this was explicit in an early version of the schema.

The same reference may be in different referenceSets, and may (unfortunately) have different names in those different sets.  This is just the truth - it is called
chr1 by some people and 1 by others.  The human genetics user community will never convert to using accession numbers or other lengthy identifiers in place
of chromosome names.

If you want to use ids then you must be able to query by name to find the id, so that when I look for 1:214217-412291 I can find the referenceId
for the reference for chromosome 1.  Using the name as the identifier within Position and other structures that are unambiguously within a referenceSet 
avoids this additional lookup, and to my mind makes things more transparent and clearer, but if you want it the other way then OK.

I am concerned that the core design, over which there was a lot of careful thought by people with a lot of experience of representing genetic data, is being
or has been mangled by layers of these sorts of decisions that are being made piecemeal without joint oversight and buy-in.

Richard

> On 28 Apr 2016, at 00:50, Mark Diekhans <notifications@github.com> wrote:
> 
> 
> +1, this is a big inconsistency in the API.
> 
> This is related to:  https://github.com/ga4gh/schemas/issues/593 <https://github.com/ga4gh/schemas/issues/593>
> 
> 
> 
> David Steinberg <notifications@github.com <mailto:notifications@github.com>> writes:
> > When performing searches on the variants or feature's endpoints the
> > referenceName field allows the client to define which contig to search on. This
> > is problematic because reference names are not required to be unique, and it is
> > not immediately clear how to form the "referenceName" field. Is it "1" or
> > "chr1"? This is further complicated by the fact that references can belong to
> > multiple Reference Sets.
> > 
> > By using the referenceId in its place we ensure that each request is against a
> > specific reference and we get better guarantees about the relationships in the
> > data. The Reads API properly makes this distinction:
> > 
> > https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/ <https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/>
> > readmethods.avdl#L26
> > 
> > ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> > 
> > You can view, comment on, or merge this pull request online at:
> > 
> > https://github.com/ga4gh/schemas/pull/616
> > 
> > Commit Summary
> > 
> > • Change referenceName to referenceId
> > 
> > File Changes
> > 
> > • M src/main/resources/avro/alleleAnnotationmethods.avdl (15)
> > • M src/main/resources/avro/sequenceAnnotationmethods.avdl (2)
> > • M src/main/resources/avro/sequenceAnnotations.avdl (3)
> > • M src/main/resources/avro/variantmethods.avdl (2)
> > • M src/main/resources/avro/variants.avdl (2)
> > 
> > Patch Links:
> > 
> > • https://github.com/ga4gh/schemas/pull/616.patch
> > • https://github.com/ga4gh/schemas/pull/616.diff
> > 
> > —
> > You are receiving this because you are subscribed to this thread.
> > Reply to this email directly or view it on GitHub*
> > 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub <https://github.com/ga4gh/schemas/pull/616#issuecomment-215266594>



-- 

diekhans:
changing that to a -1, sorry when I first looked at it; I though I added reference id; now I see it replaces reference name with id in searches.  This does not solve the problem of multiple names, because you
still have to look up the id.

We need to address the inconsistencies with the API's use of names vs ids as a whole.

diekhans:
There are two issue here, somewhat related:

  - The multiple commonly used names for chromosomes.  We can't
    fix chr1 vs 1 and need to support a list of names.  Using id
    doesn't address this issue.

  - The design of the API appears to be the all object linkage
    is done by id, not name.  So when one object references
    another, it is by id.  When one finds an object initially by
    name, then goes from object to object via id.  Id is really
    an server-local pointer.  I found this very weird, but I
    maybe WEB APIs are done this way.  I am guessing at this
    design paradigm from talking to people, it's not documented,
    so I could be wrong.

The API is inconsistent, sometimes objects are linked by name,
sometimes by id.  The API does not require chromosome name to be
unique.  This is nuts, but that is what the GA4GH API explicitly
states.

We need to do a major architectural review of the API.  It is
inconsistent and the documentation is still inadequate.

mbaudis:
@diekhans So maybe someone could provide a list of occurrences of both, linking by "id" and linking by "name"? But anyway, should be "id" (and is documented like that in the [metadata docs](https://github.com/ga4gh/schemas/blob/biosample/doc/source/api/metadata.rst#common-attribute-names-and-formats) ...).

(I personally find the "search by name/alias, link by id" not so weird ...)

richarddurbin:
From my perspective “id" should be treated like an opaque handle, specific to the particular server you are talking to,
to facilitate repeated interaction, so like a pointer.  

Whereas “name” is something with external meaning, so can be used as an entry point for search, for display back to 
users, and for export of results, for example into BAM or VCF but also potentially linking across different GA4GH servers, 
which would have their own different ‘id's for the same thing.  I note that MatchMaker Exchange can only work if people 
use externally meaningful names rather than server-specific ids.

There is a case that since the ids are really private to the server, one should minimise their use in the interface.
After all they don’t mean anything to the client.  Of course this is only possible when the names are guaranteed to 
behave as unique identifiers, but that is required to be the case for References (though whether we can really 
enforce that in the current model is not clear - we could in the original model by construction).  

Once we get into this topic of names being good identifiers we get concerned with scope.  Think about variable names in 
software.  When we write and debug a program, we want to use and see meaningful variable names, not memory 
locations as in assembly code, and the same name may end up as a different location in different compiled and executed
systems (compare to different ids in different GA4GH servers).  We control name clashes with scope.  We haven’t thought 
much about scope in GA4GH, perhaps because we haven’t thought much about the write interface to the server, nor about 
compatibility (consistency?) across severs, though no doubt there will be many servers.  

Richard

> On 2 May 2016, at 17:51, Michael Baudis <notifications@github.com> wrote:
> 
> @diekhans <https://github.com/diekhans> So maybe someone could provide a list of occurrences of both, linking by "id" and linking by "name"? But anyway, should be "id" (and is documented like that in the metadata docs <https://github.com/ga4gh/schemas/blob/biosample/doc/source/api/metadata.rst#common-attribute-names-and-formats> ...).
> 
> (I personally find the "search by name/alias, link by id" not so weird ...)
> 
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub <https://github.com/ga4gh/schemas/pull/616#issuecomment-216290445>




-- 

mbaudis:
@richarddurbin Yes, agreement here. Implementation-wise, I am all for a general solution where:

* id is whatever format is used locally, guaranteeing uniqueness, stable links/pointers
* name is locally and "context based" unique (NCBI36.1; GSM198179)
* ```aliases``` (¿```aternativeNames```? ...) provides alternative names that have been used for the same data representations (hg18 ...)

In principle, for the genome editions one could enforce/request a specific value space, to which queries etc. could be converted. But this would have to be maintained through GA4GH, and wouldn't be feasible with regard to non-human genome space. And it wouldn't work as a general data structure, for different types of records.

So, IMO id + name + aliases. And ```id``` for referencing, ```name``` and ```aliases``` for search/matching.

lh3:
In my view
* `id` is like a [foreign key](https://en.wikipedia.org/wiki/Foreign_key) in SQL, which can be auto-incremental and is not long-term stable.
* We talked about `symbol` before. It still deserves its place. A `symbol` usually does not contain spaces or special characters. We may require it to be unique within a scope. For example, [HGNC](http://www.genenames.org/) defines gene symbols; compilers may have [symbol table](https://en.wikipedia.org/wiki/Symbol_table) – in C, variable and function names are all symbols, but keywords (e.g. for/break/continue) are not.
* A `name` is just a string, which is often full-text indexed for search. HGNC takes this way as well.

> There is a case that since the ids are really private to the server, one should minimise their use in the interface.

In the SQL world, ids are often joined away. Users may completely ignore id values. GA4GH does not have a concept of "join". We have to expose ids in almost every API. This complicates the meaning of id.

> We haven’t thought much about scope in GA4GH

There was a discussion before. I believe some symbols should be unique within a certain scope, but others think this adds unnecessary constraints.

diekhans:
The analogy to foreign key is perfect and it's intention is to
be used in that way.  One can jump from object to object very
easily with id, but it's two steps if you come in with a `name'.
Very analogous to join.

IMHO it would have been clearer to have `symbol' rather than
`name' and then `description'.  The id vs name has proved ver 

I believe this PR should be to add referenceId to referenceName
rather than replace it.

david4096:
I've updated this PR to include both referenceName and referenceId in the search methods. I also added `referenceName` to SearchReadsRequest, as it is [a lone case](https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/readmethods.avdl#L26) of accepting referenceId, and not referenceName.

david4096:
After reviewing this, I believe the best pattern is to remove the need for server specific IDs in favor of using reference name everywhere.

PR #619, opened 2016-04-28
first draft of api changes

bwalsh:
Initial implementation of discussion documented here: https://github.com/ga4gh/g2p-team/issues/17


diekhans:
DWG members:  we need engagement on this PR beyond the G2P TT and UCSC;  please review and comment.   It needs to be understandable to those who are not part of developing G2P.

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,319 @@
...
+  SearchPhenotypeAssociationSetsRequest request) throws GAException;
+
+
+/******************  /phenotypes/search  *********************/
+
+/**
+One or more phenotypes can be queried together.
+*/
+record PhenotypeQuery {
+  union { null, string } name; /* new field, regex */
```
jeromekelleher: What's the use case for regex here? We use exact string matching elsewhere in the API, it would be confusing to have this endpoint do regular expressions. If we do allow regex, we'd need a _lot_ more detail on what's supported. I don't think it's worth the trouble for the first version.

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,319 @@
...
+  union { null, org.ga4gh.models.OntologyTerm } type = null;
+  union { null, array<org.ga4gh.models.OntologyTerm> } qualifier = null;
+  union { null, org.ga4gh.models.OntologyTerm } ageOfOnset = null;
+  union { null, string } description = null;  /*regex*/
```
jeromekelleher: What's the use case for searching on decription? We don't do this for any other end point. I would remove it unless it's critical.

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,319 @@
...
+  union { null, array<org.ga4gh.models.ExternalIdentifier> }  externalIdentifiers = null; /* new field */
+}
+
+
+record GenotypeQuery {
+  /**
+  only those fields from feature that are `queryable`
+  */
+  union { null, string } name; /* new field, regex */
```
jeromekelleher: Ditto here for regex and descriptions.

doc/source/api/genotypephenotype.rst
```diff
@@ -0,0 +1,173 @@
...
+feature, phenotype, and evidence fields for that instance of
+association.
+
+API
+---
+
+Request
+~~~~~~~
+
+The G2P schemas define a single endpoint ``/genotypephenotype/search``
```
jeromekelleher: Is this true? I'm seeing a whole bunch of endpoints below...

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,319 @@
...
+full feature/phenotype/evidence objects.
+*/
+record SearchGenotypePhenotypeRequest {
+
+  /**
+  The `PhenotypeAssociationSet` to search.
+  */
+  string phenotypeAssociationSetId;
+
+  union {null, array<string>} featureIds = null;
```
jeromekelleher: Do we really need arrays of featureIds etc here?

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,319 @@
...
+/******************  /genotype/search  *********************/
+
+record SearchGenotypesRequest {
+
+  /**
+  The `PhenotypeAssociationSet` to search.
+  */
+  string phenotypeAssociationSetId;
+
+  array<GenotypeQuery> genotypes = [];
```
jeromekelleher: Do we really need the list of query objects here? Why not put the fields that are in GenotypeQuery directly in here, and make a series of requests?

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,319 @@
...
+  /** ECO or OBI is recommended */
+  union { null, org.ga4gh.models.OntologyTerm }  evidenceType = null;
+  union { null, string } description = null;  /*regex*/
+  union { null, array<org.ga4gh.models.ExternalIdentifier> }  externalIdentifiers = null; /* new field */
+}
+
+
+record GenotypeQuery {
+  /**
+  only those fields from feature that are `queryable`
```
jeromekelleher: This a huge number of simultaneously queryable fields. Are all of these really necessary? What are the semantics of this operation? Are these all ANDed together?

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,319 @@
...
+  union { null, string } name; /* new field, regex */
+  union { null, string } description; /* new field,regex */
+  union { null, string } featureSetId;
+  union { null, string } referenceName;
+  union { null, long } start;
+  union { null, long } end;
+  union { null, org.ga4gh.models.Strand } strand;
+  union { null, org.ga4gh.models.OntologyTerm } type; /* new field  */
+  union { null, org.ga4gh.models.OntologyTerm } featureType;
+  union { null, array<org.ga4gh.models.ExternalIdentifier> } externalIdentifiers = null; /* new field */
```
jeromekelleher: How do we use this field to match? Is it an OR?

jeromekelleher:
It's good to see the PhenotypeAssociationSet which make things fit into the overall data model much better. To the extent that I understand the domain, it seems to make sense to me. However, I'm struggling with the API design:

- Why are there so many queries over lists? Is it really necessary to have these embedded lists of objects within the queries? Take GenotypeQuery for example. How many of these do you expect to submit to the server simultaneously? 
- Why are there so many searchable fields? This is not going to be quick to implement. I think it would be worth thinking about what the minimal useful schema is here so that it can be implemented and merged quickly.
- If all these complicated queries really are necessary, then a great deal more documentation will be needed nailing down the required server semantics exactly.

Apologies if this is going over old ground, and for the stupid questions!

doc/source/api/genotypephenotype.rst
bwalsh: Good point.  I'll update the documentation

src/main/resources/avro/genotypephenotypemethods.avdl
bwalsh: The intent was to retain the ability of the original query to search for an array of features, while moving the lookup of the features to a new endpoint.

src/main/resources/avro/genotypephenotypemethods.avdl
bwalsh: Yes, the semantics of the original query 
```
The goal here is to allow users to query using one or more of
Genotype, Phenotype, Environment, and Evidence.

A query using one of the above items is to mean, by default,
that the remainder of the query is as a "wildcard", such
that all matches to just that query term would come back.
Combinations of the above are to act like AND rather than OR.
```

src/main/resources/avro/genotypephenotypemethods.avdl
bwalsh: Terms within a query are combined via AND

src/main/resources/avro/genotypephenotypemethods.avdl
bwalsh: The original query was defined as a `query by example` where an object of that type was filled in and passed in the query.  This is actually a subset of what was query-able originally.

bwalsh:
@jeromekelleher 
Thanks very much for your time and the thoughtful feedback.

I think we have simplified the queries from the original, at least in terms of `single responsibility`.  However, we should strive to simplify more.

Hypothetically, a simpler interface would be to define a search interface for each of the entities 
POST /dataset/<id>/features/search
POST /dataset/<id>/genotypes/search
POST /dataset/<id>/phenotypes/search
POST /dataset/<id>/environments/search

Then on a selected entity ask for the associations, on a single entity at a time.
GET /dataset/<id>/feature/<id>/associations
GET /dataset/<id>/genotype/<id>/associations
GET /dataset/<id>/phenotype/<id>/associations
GET /dataset/<id>/environment/<id>/associations

That would leave it up to the client to collate and sort.

The task team has been consistent about the need for wildcards.  Absent manual curation to reconcile synonyms, it seems necessary to have regex.
e.g
BRCA1	chr17:43094426:C>T	NM_007294.3:c.1105G>A	NP_009225.1:p.(Asp369Asn)

jeromekelleher:
Thanks @bwalsh, this is very useful. Your proposal to simplify further makes a lot of sense to me. 

For the querying with lists of (e.g.) features, I don't have any real sense of the volumes of data that we're talking about here. If we usually expect to query on a single feature, and are including the functionality of querying over lists of features as a convenience to the user, I would argue strongly against it. However, if we expect to be querying thousands of features simultaneously on a routine basis, then this is a different situation.

For me, I would find it very helpful to see some example transactions. The documentation is currently very good at motivating the high-level needs of the API, but I don't get much of a feel for the low-level 'flow' of the API.

bwalsh:
Thanks @jeromekelleher 
Regarding docs:
 https://github.com/ga4gh/g2p-team/issues/17 is a short summary of how we got here.
 I've updated the doc/source/api/genotypephenotype.rst in this PR

bwalsh:
@jeromekelleher 

Regarding  `If we usually expect to query on a single feature, and are including the functionality of querying over lists of features as a convenience to the user, I would argue strongly against it. `

I collaborated with @kellrott.  Today's use cases are directed at smaller, manually curated evidence sets.  This means that we could support the simplification of removing `list of \<type\>` from the entity searches.

However, the overhead of downloading all the associations and computing the intersection on the client would be prohibitive.   So, the thought was my proposal for endpoints of the form:
`GET /dataset/<datasetId>/<entity>/<id>/associations `
won't work.  We will need to retain the: 
`POST /genotypephenotype/search` endpoint.

Regarding, `Why are there so many searchable fields`.  Again, the intent was to respect the original API's abilities while separating responsibility of searching for entities vs joining evidence.    Other than `description`, are there other fields you think should be eliminated from the query?

jeromekelleher:
Thanks for the information @bwalsh, this is very useful. My main concern really is the lists of query objects rather than a single query --- to me we need a compelling reason to include them, rather than the other way around. I'm afraid I don't have much to add otherwise, as I don't know much about g2p.

src/main/resources/avro/genotypephenotype.avdl
```diff
@@ -0,0 +1,204 @@
...
+    or as minimal as the confirmed variants that are known causation
+    for the annotated phenotype.
+    Examples of features could be variations at the nucleotide level,
+    large rearrangements at the chromosome level, or relevant epigenetic
+    markers.  Relevant genomic feature types are suggested to be
+    those typed in the Sequence Ontology (SO).
+
+    The feature set can have only one item, and must not be null.
+  */
+  array<Feature> features;
```
macieksmuga: `Feature` records do not encode variants in GA4GH, but common genomic regions such as genes, exons and CDSs. Do you want an array of `Variant` records here instead?

src/main/resources/avro/genotypephenotypemethods.avdl
```diff
@@ -0,0 +1,319 @@
...
+SearchPhenotypeAssociationSetsResponse searchPhenotypeAssociationSets(
+  /** This request maps to the body of `POST /phenotypeassociationsets/search` as JSON. */
+  SearchPhenotypeAssociationSetsRequest request) throws GAException;
+
+
+/******************  /phenotypes/search  *********************/
+
+/**
+One or more phenotypes can be queried together.
+*/
```
macieksmuga: Since there's no `PhenotypeAssociationSet` mentioned in this or other queries, I assume they can span across multiple such sets?

macieksmuga:
I really like the simplified set of endpoints @bwalsh proposed a few days back in the above discussion (https://github.com/ga4gh/schemas/pull/619#issuecomment-218327604), but I'd like to take this line of reasoning further: In keeping with the spirit of the rest of the API, why not simplify the endpoints down to a pair of phenotype retrievals on feature & variant IDs:
```
/features/<featureId>/phenotypeAssociationSet/<phenotypeAssociationSetId>/phenotypes
/variants/<variantId>/phenotypeAssociationSet/<phenotypeAssociationSetId>/phenotypes
```
Then, search for phenotypes by arbitrary string match:
```
/phenotypeAssociationSet/<phenotypeAssociationSetId>/phenotypes?search=<search_string>
```
Using the phenotype ID's from that, another pair of endpoints to retrieve features and variants per phenotype:
```
/phenotypes/<phenotypeId>/features
/phenotypes/<phenotypeId>/variants
```
and a final endpoint to gather evidence for a particular association, again using IDs:
```
/evidence?phenotypeId=<phenotypeId>[&variantId=<variantId>][&featureId=<featureId>]
```
While this simplified sketch of an API doesn't allow for arbitrary qualified search queries, neither does the rest of the GA4GH API - instead it uses the already established entity (feature, variant, phenotype) IDs omnipresent in the API's other endpoints. 

I'd love your thoughts on this @bwalsh @jeromekelleher @david4096

src/main/resources/avro/genotypephenotype.avdl
kellrott: GA4GH Variant records speak about specific mutations that occur in specific samples. A phenotypic association is usually connected to the gene, which is a 'concept' shared across a multiple samples. So you have a variant in a gene that then has an association to a phenotype.

bwalsh:
@macieksmuga 

I'm following Kyle's guidance on Feature vs Variant above.  (Also, the original API was written with a dependency on Feature).

Thanks for the clarification on `phenotypeAssociationSetId` in the URI. Can you comment on its use below?   

I think the idea of a REST interface to list phenotypes associated with a feature is a good one, see mid term roadmap below.   Minimally, the endpoint would need to implement the  ``.../<entity>/<id>/associations``, even if only internally,  in order to derive the list of related entities. 

**_Proposed short term roadmap:_**


### Entity Searches

* ``/datasets/<datasetId>/features/search``
	* 	Given a SearchFeaturesRequest, return matching _features_ in the `current 'omics dataset`.
* ``/associationsets/<phenotypeAssociationSetId>/phenotypes/search`` 
	* 	Given a SearchPhenotypesRequest, return matching _phenotypes_ in the in the `current g2p dataset.` 
* ``/associationsets/<phenotypeAssociationSetId>/genotypes/search``
	* 	Given a SearchGenotypesRequest, return matching _features_ in the in the `current g2p dataset.` 
	* 	Semantically equivalent to features/search, for standalone g2p implementations

### Association Search

* ``/associationsets/<phenotypeAssociationSetId>/genotypephenotypes/search``
	* 	Given a SearchGenotypePhenotypeRequest, return matching _evidence associations_ in the `current g2p dataset.` 

**_Proposed mid term roadmap:_**

### Convenience endpoints

(As needed)

```
GET /associationsets/<phenotypeAssociationSetId>/feature/<id>/associations
GET /associationsets/<phenotypeAssociationSetId>/feature/<id>/genotypes
GET /associationsets/<phenotypeAssociationSetId>/feature/<id>/phenotypes

GET /associationsets/<phenotypeAssociationSetId>/genotype/<id>/associations
GET /associationsets/<phenotypeAssociationSetId>/genotype/<id>/features
GET /associationsets/<phenotypeAssociationSetId>/genotype/<id>/phenotypes

GET /associationsets/<phenotypeAssociationSetId>/phenotypes/<id>/associations
GET /associationsets/<phenotypeAssociationSetId>/phenotypes/<id>/features
GET /associationsets/<phenotypeAssociationSetId>/phenotypes/<id>/genotypes
```

bwalsh:
closing. superseded by #670

PR #622, opened 2016-05-13
Pluralized various method endpoints in RNA branch.

macieksmuga:
To bring into uniformity with the rest of the API.


saupchurch:
can we get a corresponding change in the server topic branch and then bring them both in

macieksmuga:
The corresponding changes should be there already... coordinate and get this in at the same time as protobuf?

macieksmuga:
This PR to be superseded by the protobuf RNA reboot #625

PR #630, opened 2016-06-07
RNA API

macieksmuga:
The RNA API adds `RnaQuantification`, `QuantificationGroup` and `ExpressionLevel` objects, as well as endpoints to get and search these.

PR #631, opened 2016-06-08
Update pull request

saupchurch:
This contains updates to address comments in #630

### DIFF
--- a/src/main/proto/ga4gh/rna_quantification.proto
+++ b/src/main/proto/ga4gh/rna_quantification.proto
@@ -34,31 +34,47 @@ message RnaQuantification {
   // Description
   string description = 3;
 
-  // ID of the ReadGroup providing the reads for the analysis.
-  string read_group_id = 4;
+  // ID(s) of the ReadGroup(s) providing the reads for the analysis.
+  repeated string read_group_ids = 4;
 
-  // List of programIds used in the analysis.
-  repeated string program_ids = 5;
+  // Program can be used to track the provenance of how read data was quantified.
+  message Program {
+    // The command line used to run this program.
+    string command_line = 1;
+
+    // The user specified ID of the program.
+    string id = 2;
+
+    // The name of the program.
+    string name = 3;
+
+    // The ID of the program run before this one.
+    string prev_program_id = 4;
+
+    // The version of the program run.
+    string version = 5;
+  }
+  repeated Program programs = 5;
 
   // List of annotations used.
-  repeated string annotation_ids = 6;
+  repeated string feature_set_ids = 6;
 }
 
-// Identifying information for annotated features.
-message QuantificationGroup {
-  // Quantification group ID
+// Identifying information for a group of feature expression values.
+message ExpressionLevelGroup {
+  // ExpressionLevel group ID
   string id = 1;
 
   // The associated RnaQuantification.
-  string analysis_id = 2;
+  string rna_quantification_id = 2;
  
-  // Name
+  // User assigned short name for the collection
   string name = 3;
 
   // Description
   string description = 4;
 
-  // A map of additional feature group information.
+  // A map of additional group information.
   map<string, google.protobuf.ListValue> info = 5;
 }
 
@@ -70,25 +86,30 @@ message ExpressionLevel {
   // Name
   string name = 2;
 
-  // The associated QuantificationGroup. 
-  string quantification_group_id = 3;
+  // Id of the feature
+  string feature_id = 3;
+
+  // The associated ExpressionLevelGroup. 
+  string expression_level_group_id = 4;
 
   // The number of reads mapped to this feature.
-  float raw_read_count = 4;
+  float raw_read_count = 5;
 
   // Numerical expression value.
-  float expression = 5;
+  float expression = 6;
 
   // True if the expression value is a normalized value.
-  bool is_normalized = 6;
+  bool is_normalized = 7;
 
   // The units of the expression value if one is given.
-  ExpressionUnit units = 7;
+  ExpressionUnit units = 8;
 
   // Weighted score for the expression value.
-  float score = 8;
+  float score = 9;
 
-  // Confidence interval on the expression value.  Expressed as a sorted array
-  // from low to high.
-  repeated float conf_interval = 9;
+  // Lower bound of the confidence interval on the expression value.
+  float conf_interval_low = 10;
+  
+  // Upper bound of the confidence interval on the expression value.
+  float conf_interval_high = 11;
 }
--- a/src/main/proto/ga4gh/rna_quantification.proto
+++ b/src/main/proto/ga4gh/rna_quantification.proto
@@ -34,9 +34,6 @@ message RnaQuantificationSet {
 
   // The RNA quantification set name.
   string name = 3;
-
-  // The RNA quantifications in this set.
-  repeated RnaQuantification rna_quantifications = 4;
 }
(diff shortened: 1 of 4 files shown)


saupchurch:
documentation updates forthcoming

src/main/proto/ga4gh/rna_quantification.proto
```diff
@@ -34,31 +34,47 @@ message RnaQuantification {
...
 
-  // ID of the ReadGroup providing the reads for the analysis.
-  string read_group_id = 4;
+  // ID(s) of the ReadGroup(s) providing the reads for the analysis.
+  repeated string read_group_ids = 4;
 
-  // List of programIds used in the analysis.
-  repeated string program_ids = 5;
+  // Program can be used to track the provenance of how read data was quantified.
+  message Program {
```
david4096: I would suggest moving the `Program` definition to the metadata.proto so it is only defined once and used by both the reads and quantification APIs.

src/main/proto/ga4gh/rna_quantification.proto
```diff
@@ -34,31 +34,47 @@ message RnaQuantification {
...
+
+    // The ID of the program run before this one.
+    string prev_program_id = 4;
+
+    // The version of the program run.
+    string version = 5;
+  }
+  repeated Program programs = 5;
 
   // List of annotations used.
```
macieksmuga: change this comment to something like `// list of annotation sets used.`

src/main/proto/ga4gh/rna_quantification_service.proto
```diff
@@ -7,44 +7,55 @@ import "ga4gh/rna_quantification.proto";
...
 
-// This request maps to the body of 'POST /rnaquantification/search'
+// ****************  /rnaquantifications/search  *******************
+
+// This request maps to the body of 'POST /rnaquantifications/search'
 // as JSON.
 message SearchRnaQuantificationsRequest {
   // If present, return only Rna Quantifications which belong to this set.
-  string rna_quantification_id = 1;
+  string rna_quantification_set_id = 1;
```
david4096: I believe this is pointing to a new message type, `QuantificationSetId`, which must be defined as well and I presume offer its own API methods (search sets, get sets).

src/main/proto/ga4gh/rna_quantification.proto
```diff
@@ -70,25 +86,30 @@ message ExpressionLevel {
   // Name
   string name = 2;
 
-  // The associated QuantificationGroup. 
-  string quantification_group_id = 3;
+  // Id of the feature
+  string feature_id = 3;
+
+  // The associated ExpressionLevelGroup. 
+  string expression_level_group_id = 4;
```
macieksmuga: Since we want to remove `ExpressionLevelGroup` ASAP, replacing it with a more appropriate entity from the sequence annotation API, I would like to propose that we not use it as a critical link between individual `ExpressionLevel`s and `RnaQuantifcation`s. To this end, I suggest we have an `rna_quantification_id` here instead.

src/main/proto/ga4gh/rna_quantification.proto
```diff
@@ -34,31 +34,47 @@ message RnaQuantification {
...
+  }
+  repeated Program programs = 5;
 
   // List of annotations used.
-  repeated string annotation_ids = 6;
+  repeated string feature_set_ids = 6;
 }
 
-// Identifying information for annotated features.
-message QuantificationGroup {
```
macieksmuga: Ok, I'll bite. For now, why not just rename this a `FeatureGroup`, which simply functions as a grouping of features, and thus consists of (in addition to a name and ID) a `repeated string feature_ids`? Uncouple it from the other entities here!

src/main/proto/ga4gh/rna_quantification.proto
macieksmuga: In tandem with my suggestion to rename `ExpressionLevelGroup` -> `FeatureGroup`, also move `string expression_level_group_id` to `repeated string feature_group_ids`, so that this expression level can participate in zero or more of such.

src/main/proto/ga4gh/rna_quantification.proto
```diff
@@ -70,25 +86,30 @@ message ExpressionLevel {
...
-  string quantification_group_id = 3;
+  // Id of the feature
+  string feature_id = 3;
+
+  // The associated ExpressionLevelGroup. 
+  string expression_level_group_id = 4;
 
   // The number of reads mapped to this feature.
-  float raw_read_count = 4;
+  float raw_read_count = 5;
```
macieksmuga: For future reference, once we release the API, numbers on the same field *cannot* change. The point of the numbers is that as we add/modify/delete individual fields, data can (largely) be mapped across different versions of the API. Thus, once `raw_read_count` is set to 4, it stays at 4. New fields can be added at the end of the message, given new numbers.

saupchurch: Noted

src/main/proto/ga4gh/rna_quantification.proto
saupchurch: well, `FeatureGroup` was the original name for this before it was changed to `QuantificationGroup`  We changed it because of a name/conceptual clash with `FeatureGroup` in `sequence_annotation`  I'm happy to change it back as long as we can keep it distinct from `FeatureSet`

src/main/proto/ga4gh/rna_quantification.proto
```diff
@@ -34,31 +34,47 @@ message RnaQuantification {
...
+    string prev_program_id = 4;
+
+    // The version of the program run.
+    string version = 5;
+  }
+  repeated Program programs = 5;
 
   // List of annotations used.
-  repeated string annotation_ids = 6;
+  repeated string feature_set_ids = 6;
```
macieksmuga: Is this really needed here? If we search on the list of all expression levels in this quantification, they'll contain features, which in turn have links to their containing feature sets.

src/main/proto/ga4gh/rna_quantification.proto
mbaudis: Support this suggestion. ```assaymetadata``` would be the right place (see structure in https://github.com/ga4gh/schemas/pull/606).

saupchurch: agreed.  Since the plan from #606 is to merge after I've pulled this into metadata and added the import hook.  Should make it easy to switch over to importing `assaymetadata`

src/main/proto/ga4gh/rna_quantification.proto
```diff
@@ -22,6 +23,22 @@ enum ExpressionUnit {
...
+  string id = 1;
+
+  // The ID of the dataset this RNA Quantification set belongs to.
+  string dataset_id = 2;
+
+  // The RNA quantification set name.
+  string name = 3;
+
+  // The RNA quantifications in this set.
+  repeated RnaQuantification rna_quantifications = 4;
```
macieksmuga: Are you sure you want these here, as opposed to storing the `rna_quantification_set_id` in each `RnaQuantification` object? Otherwise, this object may become unbounded in size.

saupchurch: good point - I'll make the change

macieksmuga:
+1 with the caveat that the documentation in the `.rst` file now needs to be updated accordingly.

PR #634, opened 2016-06-15
Update the docs build process to use Protobuf

palfrey:
This updates the docs build process to use the Protobuf schemas. The generated RST files have been removed from the repository as they're now generated during the Read the Docs build process (using [Conda](http://read-the-docs.readthedocs.io/en/latest/conda.html) to install `protoc`). The Avro Sphinx schema has mostly been reused, but we made a new Protobuf plugin to generate suitable intermediate JSON from the protocol definition files.

Demo output of this PR is available at http://ga4gh-schemas-lshift.readthedocs.io/en/protobuf-sphinx-json/index.html


diekhans:
Since we are no longer using Avro, lets rename avpr2rest.py.  I would suggest protodoc2rst.py
Note rst not rest

palfrey:
@diekhans Fixed the avpr2rest issue

david4096:
Nice! I was able to use this process to generate docs for the BioMetadata proto descriptors [here](http://schemas.readthedocs.io/en/biodataprototestdoc/schemas/bio_metadata.proto.html).

Things to note: the map entries are not useful as they appear and the links to the descriptors for the built-in proto message types don't work. The mentions of Avro throughout the code seem misleading.

The script picked up hidden files in the source directory, which caused it to fail.

I presume this is using some beta features of the protocol buffers API. Would you provide some documentation for if a new message type becomes available or the version of the protocol buffers compiler changes? What maintenance can we expect to take on from usage of this script?

palfrey:
> Things to note: the map entries are not useful as they appear and the links to the descriptors for the built-in proto message types don't work. The mentions of Avro throughout the code seem misleading.

I'm looking at the map issues. I think most if not all of the built-in refs might go away once that's dealt with. I'll see what I can do about replacing the Avro refs.

> The script picked up hidden files in the source directory, which caused it to fail.

What sort of hidden files and where? I can't reproduce this currently.

> There are a number of places throughout the documentation that suggest that the following documentation will be a JSON representation, when we are viewing protocol buffers messages. These will need to be changed.

That's something you'll need to think about, given that the underlying definition is in protocol buffer form, but IIRC the web server will be consuming/spitting out JSON.

> I presume this is using some beta features of the protocol buffers API. Would you provide some documentation for if a new message type becomes available or the version of the protocol buffers compiler changes? What maintenance can we expect to take on from usage of this script?

So, new message types within GA4GH should just work (as you've seen above). I'd be surprised if Protocol Buffers adds some other major type other than messages and enums to the spec between now and the 3.0 release. I can add some documentation regarding what the script does and how it does things, and it *should* be ok for newer versions of the Protocol Buffer compiler (especially within the 3.x releases), but if they do a version 4 then things may change. Hopefully they won't do that for some time!

Regarding maintenance, if new GA4GH messages use the existing field types, then hopefully near zero. If they use new field types, they might not be supported, but I'll reinforce some of the existing defensive coding to make that obvious if that occurs.

palfrey:
* Map issues fixed
* Avro references replaced

david4096:
To reproduce the hidden file problem try: 
```echo "1$>,:#" > src/main/proto/ga4gh/.DS_Store
make docs```

palfrey:
@david4096 Hidden file problem should be fixed now

david4096:
+1

gabrielsaldana:
Checked out the code and tested make docs command.  I then added the g2p api files and it all works fine.

I'm testing in a docker environment, and docs generate without problems.

kozbo:
+1

jeromekelleher:
+1, looks great!

Issue #611, opened 2016-04-13
Update Experiment model


david4096:
The description for `Experiment` says it's an experimental preparation of a sample. This relationship could be more explicit if this is what the Experiment record is meant to capture by adding a sampleId field. Currently experiments are an optional field of `ReadGroups`.

david4096:
Related https://github.com/ga4gh/schemas/issues/683

Issue #595, opened 2016-04-01
dataset used inconsistently across API


diekhans:
dataset is not used consistently by the datamodel.  Since we have this concept for organizing data, it is proposed that all data objects should be in some dataset and this not be optional.  It makes data discovery and access more complex if we don't have a firmly defined hierarchy.

- `ReadGroup`  has  `union { null, string } datasetId = null;`  Is there any use case where a ReadGroups in different ReadGroupSets will be in different DataSets?  Suggest removing datasetId.
- `ReadGroupSet` has `union { null, string } datasetId = null;`. Suggest making this a required field as it is with  `VariantSet`
- `ReferenceSet` does not have datasetId.  Suggest removing this one special case and require ReferenceSet to be part of a dataset.

calbach:
> ReadGroup has union { null, string } datasetId = null; Is there any use case where a ReadGroups in different ReadGroupSets will be in different DataSets? Suggest removing datasetId.

I would agree, with the caveat that we need to be explicit that `ReadGroup` belongs to a single `ReadGroupSet`.  IIRC this is a controversial topic. Otherwise it's a gray area as we can have a case where a `ReadGroup` belongs to multiple `Dataset`s via multiple `ReadGroupSet` parents. And where does the `ReadGroup` belong if it has 0 parent `ReadGroupSet`s?

> ReadGroupSet has union { null, string } datasetId = null;. Suggest making this a required field as it is with VariantSet

+1. This should be uncontroversial.

> ReferenceSet does not have datasetId. Suggest removing this one special case and require ReferenceSet to be part of a dataset.

+1 with a caveat; this is a bit more of a semantic change details on my comment [here](https://github.com/ga4gh/schemas/issues/553#issuecomment-179983633). If we're doing this, we should consider putting a `datasetId` field on `Reference`, as it has the same many:many as `ReadGroup` does.

diekhans:
Thanks @calbach

CH Albach <notifications@github.com> writes:
> I would agree, with the caveat that we need to be explicit that ReadGroup
> belongs to a single ReadGroupSet. IIRC this is a controversial topic. Otherwise
> it's a gray area as we can have a case where a ReadGroup belongs to multiple
> Datasets via multiple ReadGroupSet parents. And where does the ReadGroup belong
> if it has 0 parent ReadGroupSets?

What is the use case for readgroups being in multiple
readgroupsets? I have never seen data organized. Also, it seems
that a readgroup should always have a parent readgroupset

[...]

Issue #507, opened 2015-12-02
Role of Dataset


mbaudis:
Somehow the definition of  ```Dataset``` got mangled up. There was never agreement, at least with metadata, that an object can only belong to one ```Dataset```. For data management & mining purposes, this doesn't make any sense - in the most benign interpretation, it just requires the addition of another grouping variable.

Any object should be allowed to be a member of 0..n Dataset(s), i.e. express any number of ```DatasetIDs```.

```DatasetIDs``` should be thought of as a "tags" attached to a data object (or its parent etc.). Some typical examples for real world conflicts with unique grouping IDs (own experience):

- GEO series, where the GSE is (was) taken as the grouping parameter, leading to painful problem when adding additional data later on (leading to series and meta-series, but also frank duplications)
- a data collection focussed on childhood gliomas, focussing on DIPG but also adding other types of aggressive pediatric high grade gliomas (so you may or may not want to use the overlap of the series)

(I am actually of the opinion that it would be even better to use saved queries instead of tags|DatasetIDs, but this has been deemed too adventurous before).

So all references requiring a single DatasetId (not a match to one of several) should be fixed.

diekhans:
The discussions about dataset all seems to end up being wanting
to to use it for things that it as not intended (at least as I
understand it).

Probably the best analogy is BioProject.
http://www.ncbi.nlm.nih.gov/bioproject/docs/faq/

It's more related to who produced the data than how it is used.

Grouping objects in datasets for different use cases or data
type selection purposes will lead to a large number of datasets
and then need to create new ones.  This would be much better
handled by queries against metadata rather than grouping
into datasets.

I suspect it's going to be problematic that searches must be
restricted by dataset.

david4096:
In my understanding, providing a logical container like dataset is simply the most useful interface for interacting with data from multiple sources. They are a convenient way to stage analyses, for example, say I am doing a processing pipeline on 1kgenomes variant data in GA4GH. I want to synthesize my knowledge about variations in that population with another from the Simons dataset. I can simply "tag" the containers I want indexed as part of my new dataset with the ID for the "Simons-1kgenomes" data.

[...]

PR #636, opened 2016-06-16
BioMetadata Protocol

david4096:
The GA4GH API currently lacks a unified way for describing samples across data types. This means that it is difficult to reconcile whether a variant call or read group is from the same sample. This PR adds the ability to filter by samples to these types.

The BioMetadata protocol allows read groups and call sets to be related to `BioSample`s. This is done by adding `bioSampleId` fields to both of those records. A `bioSampleId` field is then an optional field in both in SearchReadGroupSetsRequest and SearchCallSetsRequest. In addition to being able to filter callsets and readgroups, search and get endpoints for biosamples and individuals endpoints have been added.

Individuals can be searched by name, and biosamples can be filtered by individualId and name.

Given some read find which individual it derives from:
1) Read the bioSampleId from a ReadGroup record.
2) Get the BioSample record by performing a get at http://ga4gh/biosamples/<bioSampleId>.
3) Read the individualId from the BioSample record.
4) Get the Individual record by performing a get at http://ga4gh/individuals/<bioSampleId>.

Given some individual find reads:
1) From the Individual record get the individualId.
2) Create a SearchBioSamplesRequest with the provided individualId.
3) Perform a SearchReadGroupSetsRequest with each returned bioSampleId.
4) Only read groups matching the bioSampleId will be returned. Perform a SearchReadsRequest for each read group in the SearchReadGroupSetsResponse.

To clarify the metadata roles, some reusable elements have been moved to `AssayMetadata` (`Experiment` and `Analysis`). Other assisting documentation is included in this PR to assist proper usage of Ontology Terms. Thanks to everyone for all the effort put into this @mbaudis @diekhans @jeromekelleher @bwalsh @saupchurch @sarahhunt @dzerbino


src/main/proto/ga4gh/assay_metadata.proto
```diff
@@ -0,0 +1,96 @@
...
+
+  // The name of the experiment.
+  string name = 2;
+
+  // A description of the experiment.
+  string description = 3;
+
+  // The time at which this message was created.
+  // Format: ISO 8601, YYYY-MM-DDTHH:MM:SS.SSS (e.g. 2015-02-10T00:03:42.123Z)
+  string message_create_time = 4;
```
jeromekelleher: Why message_create_time here? The work ``message`` seems ambigous, i.e., does this refer to the individual  message objects as they are being sent to the client, or to the 'record' that is stored on the server? I'm guessing that this is because of a clash with ``run_time``? I would have thought it would be better to just use ``created`` and ``updated`` just for the sake of being consistent with all the other objects (and change them all in one go to something more specific like ``creation_timestamp``, ``update_timestamp``, if/when this is done).

I realise this happened before this specific set of changes, so please feel free to point me in the direction of the relevant discussion. This isn't a criticism of the present PR, I just thought I'd bring it up while we're reorganising this metadata stuff.

src/main/proto/ga4gh/assay_metadata.proto
```diff
@@ -0,0 +1,96 @@
...
+  string created = 4;
+
+  // The time at which this record was last updated, in ISO 8601 format.
+  string updated = 5;
+
+  // The type of analysis.
+  string type = 6;
+
+  // The software run to generate this analysis.
+  repeated string software = 7;
```
jeromekelleher: How does this relate to the existing ``Program`` object? How do we envisage using a list of strings to track provenance information about an analysis? This seems pretty weak to me; perhaps we should just leave this out for now, and attack the problem of describing the software used in an analysis in a second pass?

src/main/proto/ga4gh/bio_metadata_service.proto
```diff
@@ -0,0 +1,97 @@
...
+  // Gets a `BioSample` by ID.
+  // `GET /biosamples/{id}` will return a JSON version of `BioSample`.
+  rpc GetBioSample(SearchIndividualsRequest)
+      returns (BioSample);
+}
+// *********************  /individuals ********************************
+// This request maps to the body of `POST /individuals/search` as JSON.
+message SearchIndividualsRequest {
+  // The dataset to search within.
+  string dataset_id = 1;
```
jeromekelleher: Why are we searching within a dataset here? There is no ``dataset_id`` attribute for Individuals or BioSamples, so it's not clear what this means. Should Individuals and BioSamples be considered to be "in" a dataset? I don't think so. I think these high-level metadata items should be repository level and useable within any dataset.

src/main/proto/ga4gh/bio_metadata_service.proto
```diff
@@ -0,0 +1,97 @@
...
+  // Provide this value in a subsequent request to return the next page of
+  // results. This field will be empty if there aren't any additional results.
+  string next_page_token = 2;
+}
+
+// *********************  /biosamples ********************************
+
+message SearchBioSamplesRequest {
+  // The dataset to search within.
+  string dataset_id = 1;
```
jeromekelleher: Same comment as above re datasets.

src/main/proto/ga4gh/assay_metadata.proto
david4096: It's in the master `metadata.proto`
https://github.com/ga4gh/schemas/blob/5ccf897fc1ee9c0401db0b2d117e80d87388bec9/src/main/proto/ga4gh/metadata.proto#L32 . Stubbed out this issue here: https://github.com/ga4gh/schemas/issues/611

src/main/proto/ga4gh/bio_metadata_service.proto
```diff
@@ -0,0 +1,97 @@
...
+  
+  // Returns BioSamples with the given :ref:`name <apidesign_object_names>`
+  // found by case-sensitive string matching.
+  string name = 2;
+  
+  // Returns BioSamples for the provided individual ID.
+  string individual_id = 3;
+
```
jeromekelleher: Can we add a search by  ``disease`` also? This would be very useful. I'm not sure what the semantics of searching by ontology term should be though. Would exact matching do, or do we need to be more sophisticated?

jeromekelleher:
This looks hugely useful, and fixes a lot of the shortcomings we're experiencing in the protocol when working with the WGS500 dataset. I have a couple of minor points above, but overall I think this is a massive improvement and we should try to merge it as soon as possible.

src/main/proto/ga4gh/assay_metadata.proto
david4096: This is in the existing `metadata.proto` [here](https://github.com/ga4gh/schemas/blob/master/src/main/proto/ga4gh/metadata.proto#L129). For a more detailed discussion of provenance: https://github.com/ga4gh/schemas/issues/240 https://github.com/ga4gh/schemas/issues/390

jeromekelleher: OK, can of worms. I'm happy to ignore this one for now!

src/main/proto/ga4gh/bio_metadata_service.proto
david4096: The dataset ID is an optional field in the search request. One can have the same biosample in many datasets with this model.

src/main/proto/ga4gh/assay_metadata.proto
diekhans: The time fields in the API are weakly defined.  The other
objects use Unix time and *appear* to be about when data is
loaded into the database  This is relevant to
environments like the Google cloud, but less important
when serving up an existing dataset.

src/main/proto/ga4gh/bio_metadata_service.proto
jeromekelleher: Yes, but what are the semantics of the search? We'd need more detail about what this means in the comments if this is to be useful. It's not clear that this is strictly necessary either, so I'm suggesting we leave it out for now until we have some data on real usage.

src/main/proto/ga4gh/assay_metadata.proto
mbaudis: I thought the consensus is to do ISO, and settle for a standard naming for the object housekeeping time attributes? I/m trying to find my lengthy exchange with @diekhans ...

src/main/proto/ga4gh/bio_metadata_service.proto
david4096: I think feature additions are a good thing to address once the schema is accepted. Let's move it out to an issue if we have enough +1s for the simplest sample tracking

The [VariantAnnotationSearchRequest](https://github.com/ga4gh/schemas/blob/master/src/main/proto/ga4gh/allele_annotation_service.proto#L90) has an example of searching via ontology term.

src/main/proto/ga4gh/assay_metadata.proto
jeromekelleher: OK, I'm sorry I brought this up now! It's what's in master currently, so not strictly related to this PR. I've created an issue to track it: #637.

diekhans: The time of time has not be changed yet in the other records.    

For now, lets make the name consistent the other records (created and updated), but leave the time ISO.  We can then take on global conversion to time ISO.

message_create_time is a confusing name anyway.  It should be abstracted away from protobuf.

src/main/proto/ga4gh/bio_metadata_service.proto
mbaudis: I am actually not really happy about the structure of the ```disease``` object. While the use of the ```OntologyTerm``` structure is a compromise, it clearly does not suffice for real world scenarios where you need multiple collated terms (see the [discussion](https://docs.google.com/document/d/1aTK-hWdTPbSt3DhB-pW8jDvK5-krJW557Vsw6V3jG-s/edit?pref=2&pli=1#heading=h.w8b0lsmtgmwl), or the "[phenopackets](phenopackets github)" proposed by @mellybelly, @cmungall et al.).
If it is easy to change afterwards, diagnosis (OntologyTerm) is fine; but it is certainly not final. See also metadata branch etc.

src/main/proto/ga4gh/assay_metadata.proto
diekhans: This should be changed to use a list of Program objects, however lets do that in a separate PR.

src/main/proto/ga4gh/assay_metadata.proto
mbaudis: Full agreement with @diekhans here.

diekhans: id should be a server-defined id, not a UUID.

src/main/proto/ga4gh/bio_metadata_service.proto
david4096: For more discussion on how this pattern exists in our API and to spark discussion on how to normalize access patterns see https://github.com/ga4gh/schemas/issues/595 https://github.com/ga4gh/schemas/issues/507 . I think this is a good inclusion to support the "let me search the 1kgenomes samples," a common data interchange use case.

src/main/proto/ga4gh/assay_metadata.proto
david4096: Re: UUID, fixed here and in [master](https://github.com/ga4gh/schemas/pull/638).

diekhans:
@david4096 pointed out that many of the inconsistency that are being noted are already in master.   Given this, we should go ahead with this PR and fix the inconsistencies in separate PRs

that is +1

src/main/proto/ga4gh/bio_metadata_service.proto
diekhans: I believe all objects are should to be in datasets.  It is a
container around ownership.  It is also the only way we can
manage the scope of `name'.

Being in different datasets shouldn't be considered a barrier
using a object. If it is, the API has a serious problem.

jeromekelleher:
I'm happy to +1 this if we remove the ``dataset_id`` argument from the searches. I don't think we've clearly defined the model here. This is definitely different to other objects which unequivocally have a ``dataset_id`` attribute, and can therefore be said to belong to a dataset. Rather than get bogged down in this, I suggest we just remove ``dataset_id`` from the search functions for now, as it's optional anyway.

diekhans:
agreed to move forward and fix dataset inconsistency in another PR.

david4096:
@jeromekelleher I added a line stating the dataset_id is optional. Looking forward to moving ahead with this!

diekhans:
@jeromekelleher the problem here is that dataset was removed from the metadata objects.  All objects should be in datasets (IMHO, it's wrong that  references special cases and there is a discussion about fixing it).

dataset is a ownership bag.  This ownership appears to have the side-effect that it also scopes name and makes it the owners responsibility to ensure name uniqueness.

It is specifically not a barrier to doing cross-dataset analysis.  This is really expected to be the norm.

@calbach, @dglazer  Dataset still causes lots of confusion with people wanting it to be something that it's not.  Could you please work on more detailed documentation.  

I am +1 for putting all metadata in a dataset and leaving it in the query.

pgrosu:
You might remember [my diagram](https://github.com/ga4gh/schemas/pull/74#issuecomment-44753668) from a couple of years ago - you don't really need a dataset id, as all the generated components are tied to one another, and any (manually/auto-generated) additional meta-attributes would be automatically bound and propagated from the previous data component that generated them, all stemming from an initial sample:

https://github.com/ga4gh/schemas/pull/74#issuecomment-44753668

Each specific Project/Study - which encompasses the Experiment and Analysis performed - would thus differentiate similar collections of Samples that were processed in different ways.

diekhans:
The current approach of the GA4GH API uses datasets as the container.

This could be change, we would do this a part of a full API design,
not piece meal.

src/main/proto/ga4gh/bio_metadata_service.proto
cmungall: > Can we add a search by disease also? This would be very useful. I'm not sure what the semantics of searching by ontology term should be though. Would exact matching do, or do we need to be more sophisticated?

I think doing this lexically would be a bad idea. Would it be possible to use an ontology service (either part of GA4GH or distinct) to first retrieve a disease identifier. The sample query could then take that identifier (rather than a label) and implement the appropriate search. A lot more needs to be specified here. What should happen in the case where we have equivalence axioms between identifiers in different disease resources (equivalent diseases can have different IDs in OMIM, Orphanet, NCIT, DO, MESH, ICD9/10/11, SNOMED, ...). What about when a generic disease that subsumes various specific diseases? If a user requests "parkinson disease" samples, then any sample annotate with any of the different forms of PD should presumably be returned, but how this should be implemented is under discussion in a different ticket.

I'm not sure what to recommend as I'm not sure what the general design pattern is for the FooRequest objects. We can imagine a fairly open-ended set of requirements for different ways to search samples. It looks like the goal is not to allow the expressivity of a full-blown query language. Instead common types of request are supported by shadowing the domain objects, allowing the client to plug in values that are interpreted as a conjunctive query.

To be honest I'm a bit confused by:

https://github.com/ga4gh/schemas/blob/2efea413027a820a53f313efa2ed5cf8ee380d31/src/main/proto/ga4gh/bio_metadata_service.proto#L64-L82

The semantics of this aren't clear.

The comment says the dataset id is optional, but it doesn't have the `optional` proto tag.

The name field is not marked optional, the docs say string matching is to be used. We might assume that an empty string would match everything (more docs are required on the type of string matching. Any substring? Lucene type queries?).

I find the choice of `name` as the field confusing here. Why not something like `nameMatch`?

The compositionality of these two fields aren't clear. I'd guess these would be conjunctive, e.g. I can search for samples matching a string AND that are in a particular dataset.

This could rapidly become confusing as the ability to do more expressive search is added (unless the goal is to deliberately restrict to highly unexpressive searches)

david4096: Yes, I believe that at this phase of data modeling we should be focused on the minimum level of search expressivity to carry out data interchange. If other analytical use cases exist, they can be added by implementors and the GA4GH methods become a subset.

Ontology and synonym search is out of scope for the GA4GH, an ontology service will provide you with the set of search terms you may want to use. The goal is to provide the minimum data model required to perform data interchange and if an ontology term is too limited for the disease model, we can address this in future work. I would avoid a kitchen sink approach.

The `optional` tag has been removed from protobuf 3.

You're correct: name matching is in the spirit of the rest of the API https://github.com/ga4gh/schemas/blob/master/src/main/proto/ga4gh/read_service.proto#L59 https://github.com/ga4gh/schemas/blob/master/src/main/proto/ga4gh/variant_service.proto#L143 and these fields are matched directly and all matches have to be present for a given message to be returned.

pgrosu:
So part of the reason why I [wrote up the comprehensive analysis](https://github.com/ga4gh/schemas/issues/264#issuecomment-90081657) of the GA4GH ecosystem - which was the final push to switch us to [Protocol Buffers](https://developers.google.com/protocol-buffers/) - was to take a step back and consider all the moving parts and types of data models for GA4GH, including their associated methods and queries.  When I created the [above-referenced diagram](https://github.com/ga4gh/schemas/pull/74#issuecomment-44753668), it was to initially get us to consider another approach of how data moves along a wire from multiple global repositories, in order to help us put together the schemas and methods which would allow for [complex](https://github.com/ga4gh/schemas/pull/253#issuecomment-97489069) [queries](https://github.com/ga4gh/schemas/pull/636/files#r67718852) that @lh3 and @jeromekelleher are looking for.  Yes, a full API would have additional \<Key,Value> model-based components with several dynamically-updated inverted indices to allow examples like the below-referenced comments, but they are based on having the minimally-required definitions of a GA4GH API wire-protocol that would naturally lend itself to auto-updating all the dependent components, based on the volume of data flowing over-the-wire in the coming years:

https://github.com/ga4gh/schemas/issues/248#issuecomment-104301212

https://github.com/ga4gh/schemas/issues/142#issuecomment-55518571

https://github.com/ga4gh/schemas/issues/154#issuecomment-56716514

I also suspect that digests of such data structures would surely be distributed, where load-balanced analyses would be performed across multiple ever-increasing repositories in order to also pick specific versions of the sets of interest before having the processed results reported back to the user - which is another layer of complexity.  In any case, as I'm a fairly patient person :)

Hope it helps,
Paul

kozbo:
poke. Can we wrap up this discussion with another +1 from someone? @jeromekelleher have your concerns been met with both spawned off issues and some minor adjustments?

src/main/proto/ga4gh/bio_metadata_service.proto
jeromekelleher: Thanks for the comments @cmungall, this is very useful. For what it's worth, I tend to agree with you about the structure of the queries, their lack of flexibility and poorly defined semantics. This isn't a criticism of this PR, it's more an observation on the overall structure of the API which has been generalised outwards from a couple of very specialised cases (searching for reads and variants, essentially).

jeromekelleher:
I'm still not sure we've properly defined the semantics of the ``dataset_id`` arguments here @kozbo, which is why I suggested deleting them for now. However, I guess we can just think of these objects as just inheriting their dataset from the VariantSet or ReadGroupSet or whatever and so the meaning is clear enough.

+1

diekhans:
I don't see how we can have it in the request is it's not part
of the object.  There is nothing to query against.

macieksmuga:
@diekhans Having the `dataset_id` in the search but not in the object models an optional many-to-many relationship between `Individual` and `Dataset`. It is exactly the same formalism as currently found in the API between `Reference` and `ReferenceSet`. I agree that adding documentation to explicitly state this is desirable, but we were hoping to introduce said documentation as part of a larger `Dataset` overhaul.

diekhans:
I don't understand.  The Reference and ReferenceSet searches
don't take dataset_id.

There is no concept of optional dataset in the schemas.

Dataset was removed from the object based on confusion.
Can we just put it back in?

src/main/proto/ga4gh/bio_metadata_service.proto
pgrosu: @david4096 Sounds like the right path to optimize at the start, but with the option to expand down the line.  Ideally this would be performed via simulated complex query tests where the data is distributed and load-balanced.  This way we know the dependent components are possible, and that the data transfer is not the bottleneck.  For example, users will eventually want to query by disease in a genome browser - see my previous post here:

https://github.com/ga4gh/schemas/pull/570#issuecomment-194514979

Very excited about this new approach!
~p

pgrosu:
Mark, it's encapsulated by the intersection of shared information, which connects the two together via `source_accessions` leading to natural many-to-many translation tables.  It basically would allow for tables I wrote here:

https://github.com/ga4gh/schemas/issues/125#issuecomment-52805241

Then the more frequent many-to-many connections can be cached for faster access.

~p

PR #639, opened 2016-06-22
Add name and gene_symbol to feature and featureSearch

david4096:
This PR improves the sequence annotation API by adding two new search fields to the SearchFeaturesRequest, as well as to the feature message. These fields are `name` as defined in the remainder of the API and `gene_symbol`, which should be a common symbol selected by the data preparer for the gene a feature occurs on. Both of these fields are optional in both the request and message.

For example, one can now perform a search for features on the BRCA1 gene with a SearchFeaturesRequest that looks like: `{"geneSymbol": "BRCA1"}`. This will return all features that have been labeled as having that gene symbol. Similarly, strict string matching on name makes it simple to use familiar access patterns from the rest of the API.

These fields were determined to be those necessary to perform the G2P and RnaSeq data ingestion tasks. There is currently no way, aside from downloading the entire feature set and inspecting each feature individually, to carry out this task. This change will enable a clearer interface for populating `feature_id` fields of RNA expression levels and `FeaturePhenotypeAssociations`. @bwalsh @diekhans @saupchurch @macieksmuga


src/main/proto/ga4gh/sequence_annotation_service.proto
```diff
@@ -70,41 +70,47 @@ message SearchFeaturesRequest {
   // The annotation set to search within. Either `feature_set_id` or
   // `parent_id` must be non-empty.
   string feature_set_id = 1;
+  
+  // The name of this feature.
+  string name = 2;
+  
+  // The symbol for the gene that this feature occurs on.
+  string gene_symbol = 3;
```
diekhans: Since this is a highly specific feature to the a generic type, this may not be long-term solution. Please document these issues.  Perhaps a more generic name, such as just "symbol" would be clearer.

diekhans:
+1, other than a comment that stating that pulling out gene_symbol into a generic type is preliminary.

david4096:
@diekhans I've added the comment to the `Feature` and `SearchFeaturesRequest` message: `This field may be replaced with a more generic representation in a future version.`

src/main/proto/ga4gh/sequence_annotation_service.proto
```diff
@@ -70,41 +70,48 @@ message SearchFeaturesRequest {
   // The annotation set to search within. Either `feature_set_id` or
   // `parent_id` must be non-empty.
   string feature_set_id = 1;
+  
+  // The name of this feature.
```
jeromekelleher: What are the semantics of searching on this field? Is it optional or mandatory?

src/main/proto/ga4gh/sequence_annotation_service.proto
```diff
@@ -70,41 +70,48 @@ message SearchFeaturesRequest {
...
+  string name = 2;
+  
+  // The symbol for the gene that this feature occurs on. This field may be
+  // replaced with a more generic representation in a future version.
+  string gene_symbol = 3;
```
jeromekelleher: Same problem here. This is just describing the field (which is already done). This should be documenting the _search_.

david4096:
@jeromekelleher Thanks I've clarified the documentation to specify a case sensitive exact match on those fields.

macieksmuga:
+1

macieksmuga:
Requesting additional +1s on this... @sarahhunt @reece @mbaudis @jeromekelleher @diekhans @bwalsh @benedictpaten

diekhans:
+1!!

sarahhunt:
Name was definitely needed. Not so sure about gene symbol - what about regulatory features? - external identifier would have been cleaner here. 
I assume if names and symbols relating to different genes are supplied in the query an empty list of features is returned?

macieksmuga:
@sarahhunt I can't answer about `external_identifier`, but as for conflicting `name` and `gene_symbol` fields, then yes, I believe that the intention is that either one or the other is supplied. If both are given and conflict with each other then you will get back an empty list.

jhl667:
What is the expected behavior when synonym gene symbols are passed?  I worry about that with something that isn't always well defined such as a gene_symbol.  My thought is that being able to source the origin of symbols is important.

diekhans:
thanks @sarahhunt

gene_symbol is a short term solution until we can come up with a
proper data mode.

It's not clear if ExternalIdentifier is the right solution for
gene_symbol. ExternalIdentifier hasn't real been used so we need
to do some data modeling with it.  It was in add in without
enough investigation; I know, I put it in.

In the case of gencode, sometimes it's an external identifier,
sometimes it's something internal only to ensembl/gencode, or
even another database.

Also ExternalIdentifier is a type, not a field.  It's not clear
how to generically name the field.  Feature is conceptually
polymorphic, however there is not a natural way to do polymorhism
in protobuf (or avro, or thrift).

This is something that the Google/EBI's proposed modifications
to the sequence annotations schema attempts to address.

kozbo:
Thanks for your comments @jhl667, @sarahhunt, @diekhans.  

This PR is about solving some the need to be able to find a feature with the ensemble ids from RNA data or with the gene symbol as used in the G2P data. Which values you put into the Name and gene symbol fields is up to the data preparer and must be set at ingest time. In fact, our first suggestion didn't include the gene symbol field. It was to just overload Name and put the ensemble ID in there for the RNA feature set, and put the gene symbol in there for the G2P feature set. But we felt it was clearer to have separate fields for each, thus the resulting PR. There are many more problems to solve here, like the synonym issue brought up by @jhl667. For now that problem will have to be solved with the more exhaustive search done on the info field values. We can solve that efficiency problem in another PR. We may have to promote other fields out of the info section to the top level as needs arise (like we have in this PR).

I hope that clarifies our approach. We can pull out the gene symbol field and go back to overloading the Name field if so desired but we would prefer to stay the course for now. RNA and G2P modules need us to decide and commit the code one way or the other very soon, as they are both blocked by this.

@jhl667, @sarahhunt : Are you OK with keeping the PR as is?

sarahhunt:
Thanks for the implementation clarification @macieksmuga .
If this is a short term solution essential for progress +1

PR #647, opened 2016-07-06
Simplify list reference bases endpoint

david4096:
The GA4GH genomics API offers a utility for listing the bases of a given reference, by ID. However, the access pattern for getting these bases differs from the remainder of the API. This PR suggests that the API could be made more consistent by enabling the familiar RPC style message passing as opposed to the query string.

For example, when listing variants via the variants API, a search variants request is sent to the variants RPC endpoint. This same problem might have been alternatively solved using the URL string `/variants?start=0&end=123123&referenceName=1`, where the fields of the search variants request would be populated via the query string. Although both approaches are valid, the former is favored throughout the schemas.

This PR does nothing to modify the functionality of the API, it simply changes a GET request to a POST request to improve the consistency of our offering by unifying access patterns.


diekhans:
+1  in should be consistend

While this isn't intended to be traditional `RPC' (it's still
stateless) , post are used due to the difficulty in encoding
complex requests as a flat query string.

There was discussion of providing for both options; simple
requests could use GET, while more complex ones would require
POST of JSON.  There maybe a ticket, but nothing ever came of
this discussion.

david4096:
@jeromekelleher What do you think of this? It's the only endpoint that uses query parameters and would be trivial to change to a POST.

diekhans:
Can anyone for the reads task team comment on the rationale for this inconsistency and then vote.

kozbo:
+1  @cassiedoll do you have any objections to this change?

fnothaft:
IIRC, @delagoya is the person to comment RE reads task team.

jeromekelleher:
I don't think this change helps much, as the request is still an outlier. Why do we have the ID in the URL  instead of the request? Why is it called ListReferenceBases instead of SearchReferenceBases? Making the API consistent is a good goal, but this change doesn't really help that much.

delagoya:
Have to agree with @jeromekelleher here. If you are shifting to using a JSON packet for a POST request, then the URL should change to not have URL parameters at all.

david4096:
Thanks @delagoya and @jeromekelleher! It sounds like there is consensus for a new endpoint that accepts `ListReferenceBasesRequest`. I've unceremoniously named it `/listreferencebases`.

kozbo:
+1 update looks like it meets the most recent requests

kozbo:
bump. Would like to get this completed as it is blocking our ability to add the schema annotations required for going from protobuf to swagger.

delagoya:
+1 looks good to me. 

Sorry for the delay!

delagoya:
@kozbo would you like me to merge?

kozbo:
@delagoya Yes please, thanks!

dcolligan:
Looks like there's consensus... merging...

Issue #623, opened 2016-05-19
ANN to TranscriptEffect field correspondence


GuilhemSempere:
Hi !

My colleague and I are having doubts implementing a REST service meant to return functional annotations in ANN format from a VCF file. This data is typically added by SnpEff. We are actually trying to find out how each of the VCF fields fits in the objects defined via the allele annotation schema.

Our mapping looks like this so far (right-hand field names based on the structure of the TranscriptEffect class):

Allele -> alternateBases
Annotation -> effects[0].term (may we have several terms in this list?)
Annotation_Impact -> ?
Gene_Name -> ?
Gene_ID -> ?
Feature_Type -> ?
Feature_ID -> featureId
Transcript_BioType -> ?
Rank -> ?
HGVS.c / HGVS.p -> hgvsAnnotation.genomic / hgvsAnnotation.protein
cDNA.pos / cDNA.length -> cDNALocation.start / cDNALocation.end
CDS.pos / CDS.length -> CDSLocation.start / CDSLocation.end
AA.pos / AA.length -> proteinLocation.start / proteinLocation.end
Distance -> ?
ERRORS/WARNINGS/INFO -> ?

We are tempted to feed the “info” map in VariantAnnotation with all the fields that we could not find an equivalent for, but we have the feeling that we are missing something.

Can someone please clarify this for us? Thanks in advance,

Guilhem

david4096:
The info field is an effort to capture the other bits of data that don't have named fields in the schema. However, it seems like in your case there are elements of individual transcript effects you would prefer to represent alongside the transcript effect itself. A temporary workaround could be to add to the info field any details about the transcript effect by using the transcript effect ID. For example, in the case of distance you might try `annotation.info[transcriptEffectId + "distance"] = distance`.

An alternative solution would be to propose an `info` field at the level of the transcript effect. This would allow these data to be better represented in the schema.

david4096:
@GuilhemSempere with the addition of the `info` field to transcript effect messages one should be able to interchange arbitrary key:value pairs. Our general approach is to add named fields to messages as needed. Please let me know if you think we can close this issue!

GuilhemSempere:
Yes, this solution suits our needs, you can close the issue. Thanks!

PR #648, opened 2016-07-06
Add an info field to transcript effect record

david4096:
As stated in issue https://github.com/ga4gh/schemas/issues/623, the current transcript effect model loses fidelity for some information generated by SnpEff. This PR adds an `info` field in a spirit similar to the rest of the API which maintains key:[value] pairs for modeling information that do not have named fields.

This change allows a wider variety of data to be modeled and induces minimal changes to existing implementations.

Here is the mapping provided in #623 that shows which fields are not being represented. With this PR, all data presented by a `?` would be appended to the info field.

> Allele -> alternateBases
> Annotation -> effects[0].term (may we have several terms in this list?)
> Annotation_Impact -> ?
> Gene_Name -> ?
> Gene_ID -> ?
> Feature_Type -> ?
> Feature_ID -> featureId
> Transcript_BioType -> ?
> Rank -> ?
> HGVS.c / HGVS.p -> hgvsAnnotation.genomic / hgvsAnnotation.protein
> cDNA.pos / cDNA.length -> cDNALocation.start / cDNALocation.end
> CDS.pos / CDS.length -> CDSLocation.start / CDSLocation.end
> AA.pos / AA.length -> proteinLocation.start / proteinLocation.end
> Distance -> ?
> ERRORS/WARNINGS/INFO -> ?

### DIFF
--- a/src/main/proto/ga4gh/allele_annotations.proto
+++ b/src/main/proto/ga4gh/allele_annotations.proto
@@ -119,6 +119,9 @@ message TranscriptEffect {
 
   // Output from prediction packages such as SIFT.
   repeated AnalysisResult analysis_results = 9;
+  
+  // Additional data in key-value pairs.
+  map<string, google.protobuf.ListValue> info = 10;
 }
 
 // A `VariantAnnotation` record represents the result of comparing a variant to


sarahhunt:
@david4096 - I missed the earlier discussion, but a number of these values are attributes of the gene or transcript which I expect would be retrieved via the sequence annotations API using the transcript id rather than duplicated in TranscriptEffect.
That said, there is no harm in an additional info field, as long as important, commonly used 'info's are identified and promoted to named attributes.

david4096:
@sarahhunt agreed! @GuilhemSempere had provided the above list. I believe the fields that are not currently described in feature messages would be Distance, 
ERRORS/WARNINGS/INFO, and Annotation_Impact. This provides a consistent way to continue to interchange these data without enforcing them as fields.

david4096:
@sarahhunt can I get your +1 ?

sarahhunt:
+1. Any other comments @pcingola, @andrewjesaitis ?

andrewjesaitis:
+1

The only thing that jumped out to me is in the #623 comment. I believe it should be `HGVS.c -> hgvsAnnotation.transcript`, since this is the annotation relative to the transcript and not to the genome. Everything else looks good.

I agree on the data duplication point that @sarahhunt brought up, but since these info fields are just an arbitrary map, it can be seen as an implementation detail (that is adding/checking what is a "sequence annotation" field and adding the rest here).

sarahhunt:
Thanks @andrewjesaitis! Good point- I'd not noticed the HGVS.c ->hgvsAnnotation.genomic misassignment.

PR #651, opened 2016-07-07
Move Program message to metadata.proto

david4096:
The quantification API makes use of the Programs message for its provenance model. This PR makes the message available for reuse by moving it to `metadata.proto`. This clears the way for further work to make the provenance model consistent.

### DIFF
--- a/src/main/proto/ga4gh/metadata.proto
+++ b/src/main/proto/ga4gh/metadata.proto
@@ -41,3 +41,21 @@ message Dataset {
   // Additional, human-readable information on the dataset.
   string description = 3;
 }
+
+// Program can be used to track the provenance of how read data was generated.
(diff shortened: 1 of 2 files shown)


saupchurch:
thanks +1

diekhans:
program is general enough that it belongs in the general metadata.proto file,
not assay.

david4096:
@diekhans updated, and @saupchurch we still have your +1 for metadata.proto I imagine?

saupchurch:
yes having this in metadata.proto is still +1 good

diekhans:
+1

-----Original Message-----

sarahhunt:
This makes sense. +1

mbaudis:
+1

Issue #583, opened 2016-03-21
Can a callset be in multiple variant sets?


david4096:
Callsets can be a member of multiple variant sets according to the schema, yet the reference server is currently underspecified for this case. Is there an example of when a callset is in multiple variant sets?

```
  /** The IDs of the variant sets this call set has calls in. */
  array<string> variantSetIds = [];
```
https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/variants.avdl#L90

`SearchCallSetsRequest` requires a single variant set ID to be specified, making the above semantics even more strange. If a callset can be a member of multiple variant sets, why do we specify a single variant set ID when performing search?

```
record SearchCallSetsRequest {
  /**
  The VariantSet to search.
  */
  string variantSetId;
```

https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/variantmethods.avdl#L158

mbaudis:
@david4096 Logically Callsets should only refer to one Variantset, since they can be thougfht of as an ordered list with the length == no. of variants described. So Variantsets would have to be identical, to be referred from Callsets.

jacmarjorie:
Also, 
```
record SearchCallSetsRequest {
  /**
  The VariantSet to search.
  */
  string variantSetId;
```

If this is agreed then variantSetIds in the CallSetRequest should be a list, not a string.

diekhans:
@jacmarjorie

I believe your use cases is what was imagined.  However there is
a multi-month discussion, that was never resolved,
on if this should be supported:

   https://github.com/ga4gh/schemas/pull/395

We would love contributions to the documentation on variants,
including documenting use cases justifying the design:

   https://github.com/ga4gh/schemas/issues/408
   https://github.com/ga4gh/schemas/issues/379

Variants is suffering from no one who has a deep understanding
of variants and VCF analysis owning finishing the work.

Mark

diekhans:
related to https://github.com/ga4gh/schemas/pull/395 and https://github.com/ga4gh/schemas/pull/412

david4096:
https://github.com/ga4gh/ga4gh-schemas/blob/master/src/main/proto/ga4gh/variants.proto#L75

Callsets are still allowed to be in multiple variant sets. We should remove this. The biosample ID tag on callsets is what allows you to compare calls in multiple variant sets.

PR #655, opened 2016-07-07
Calls search endpoint

david4096:
This PR enables variant call data to be queried directly. Currently, call data are only made available in the variant message itself, which makes large sample sets difficult to work with. By enabling a separate endpoint for querying call data we can greatly improve common access patterns without disturbing the existing method.

The endpoint, `calls/search` receives a SearchCallsRequest, in which one provides a `variant_set_id`, `call_set_id`, or `variant_id`. This allows pages of Calls to be returned matching the search request as opposed to in a variant message. To carry out these requests and to support the GetCallSet request `variant_id` and `id` have been added to the `Call` message.

The existing method of accessing call data is left in place, however, I believe in general this practice of modifying messages at query time to be problematic. It leads to the situation where you have one identifier that could describe two unique returned messages. If this access pattern proves to serve for the same existing use cases we ought to deprecate placing calls in a variant message. For previous discussion on callsets see https://github.com/ga4gh/schemas/issues/583.

### DIFF
--- a/src/main/proto/ga4gh/variant_service.proto
+++ b/src/main/proto/ga4gh/variant_service.proto
@@ -133,6 +133,46 @@ message GetVariantRequest {
   string variant_id = 1;
 }
 
+// ******************  /calls  *********************
+// This request maps to the body of `POST /calls/search` as JSON.
+message SearchCallsRequest {
+  // The VariantSet to search.
+  string variant_set_id = 1;
+
+  // Only return calls with this Call Set ID.
+  string call_set_id = 2;
+  
+  // Only returns calls for this variant ID.
+  string variant_id = 3;
+
+  // Specifies the maximum number of results to return in a single page.
+  // If unspecified, a system default will be used.
+  int32 page_size = 4;
+
+  // The continuation token, which is used to page through large result sets.
+  // To get the next page of results, set this parameter to the value of
+  // `next_page_token` from the previous response.
+  string page_token = 5;
+}
+
+// This is the response from `POST /calls/search` expressed as JSON.
+message SearchCallsResponse {
+  // The list of matching calls.
+  repeated Call calls = 1;
+
+  // The continuation token, which is used to page through large result sets.
+  // Provide this value in a subsequent request to return the next page of
+  // results. This field will be empty if there aren't any additional results.
+  string next_page_token = 2;
+}
+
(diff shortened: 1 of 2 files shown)


kozbo:
+1  this seems like a great improvement after writing some programs that search the calls using the current scheme. 

@mbaudis @jeromekelleher @jacmarjorie @diekhans Any thoughts?

src/main/proto/ga4gh/variant_service.proto
```diff
@@ -133,6 +133,46 @@ message GetVariantRequest {
   string variant_id = 1;
 }
 
+// ******************  /calls  *********************
+// This request maps to the body of `POST /calls/search` as JSON.
+message SearchCallsRequest {
```
diekhans: would be good to document which fields in search request are required vs optional.

Is there any reason to leave the array of call in a variant with this endpoint?

diekhans:
This seems well worth trying out (as in +1).

Is there any reason to leave the array of call in a Variant object with
this endpoint?

david4096:
@diekhans the calls array is left as is to improve the likelihood of acceptance by easing migration to the new access pattern.

sarahhunt:
@david4096  - this is a far more intuitive way to query genotype calls. Is variant_set_id to be mandatory or does this allow non-strandard querying across sets?
If variant_id is not mandatory, is there any expectation that the calls will be ordered by genomic location?
I assume you are aware that @reece and the VMC are looking at this area too.

david4096:
Thanks @sarahhunt. I believe that leaving `variant_set_id` as an optional field allows more expressive queries without changing the data model. I have not made the expectation that they will be returned by genomic position.

I am excited to see what the VMC comes up with. I see this PR as a simple iterative change from our current access patterns. I believe that adding some notion of `Sequence` and/or `Allele` and making the variant model unary will further clarify the domain. We need to provide the simple bridge from VCF, and I believe this is one further step.

kozbo:
@david4096 will demo this approach in a test program before merging

pgrosu:
A very sensible and welcome recommendation!

Thank you and looking forward to the demo,
Paul

david4096:
As per @diekhans comment, I think removing the callSetIds parameter of variants/search is correct.

PR #665, opened 2016-07-12
Search references by name

david4096:
A common use case of the references API is to find `reference_id` for building a SearchReadsRequest. This PR adds the ability to search references by their name.


diekhans:
+1, although should document that reference_set_id is required.

kozbo:
+1

PR #675, opened 2016-07-28
Make rna_quantification_set_id mandatory...

dcolligan:
... for SearchRnaQuantificationsRequest's

See https://github.com/ga4gh/server/issues/1314

PR #676, opened 2016-07-28
Make rna_quantification_set_id mandatory...

david4096:
Pluck @dcolligan's commit onto the RNA branch as opposed to rna-update.


saupchurch:
moving +1 from #675 to here

PR #677, opened 2016-08-02
Remove feature group from docs

david4096:
@dcolligan


dcolligan:
Seems like we're deleting too much here... we only want to remove references to feature groups

david4096:
The first commit removes mention of feature group from hand written docs. The second removes docs created by @saupchurch which can now be generated automatically.

dcolligan:
+1

PR #699, opened 2016-08-24
Replace NCBI taxon ID integer with ontology term

### LINKED ISSUE #612: References replace NCBI Taxon ID with Ontology Term
The current reference model uses the NCBI taxon ID to define the species of a record. Should this be changed to an Ontology Term?

https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/references.avdl#L69

david4096:
In #612 I mentioned the inconsistency in representing the NCBI taxon ID as an integer on a reference when the biometadata objects use an Ontology term. This PR closes #612 by replacing the NCBI taxon ID `int32` with an ontology term.


mbaudis:
-1 on this in the current implementation.

Species is defined in```bio_metadata.Individual.species```

So this should be a reference to the individual (or a copy of the same structure, as a fudge). 

```
  // For a representation of an NCBI Taxon ID as an OntologyTerm, see
  // NCBITaxon Ontology
  //   http://www.obofoundry.org/ontology/ncbitaxon.html
  // For example, 'Homo sapiens' has the ID 9606. The NCBITaxon ontology ID for
  // this is NCBITaxon:9606, which has the URI
  // http://purl.obolibrary.org/obo/NCBITaxon_9606
  OntologyTerm species = 7;
```

david4096:
Thanks @mbaudis, that makes sense I've changed the field to be named species.

mbaudis:
+1 (and removal of the -1 above)

david4096:
Updated to include using an Ontology Term on reference sets as well. Thanks @ejacox !

david4096:
Implemented in the server [here](https://github.com/ga4gh/server/pull/1513), and [client](https://github.com/ga4gh/ga4gh-client/pull/41/) with demonstrated tests in [compliance](https://github.com/ga4gh/compliance/pull/238).

PR #701, opened 2016-08-25
Updates G2P documentation

david4096:
Added a banner
Removed mentions of the extra <datasetId> paths
Removed mentions of genotypes/search
Removed the list of convenience endpoints
Added g2p to index toctree
Add g2p schemas to schema index toctree
Removed generated avro docs

The automated doc generation process requires that comments be placed directly above the field they are describing. Also improved styling. @bwalsh @dcolligan @sarahhunt


src/main/proto/ga4gh/genotype_phenotype_service.proto
```diff
@@ -80,28 +89,34 @@ message EvidenceQuery {
...
     string phenotype_association_set_id = 1;
 
+    // Phenotype ID
     string id = 2;
 
-    string description = 3; /* The system should support regular expression matches */
+    // The system should support regular expression matches
+    string description = 3;
 
+    // type
```
david4096: Details on how the request fields are to be handled should be in this document. I added boilerplate when there was not other description.

src/main/proto/ga4gh/genotype_phenotype_service.proto
```diff
@@ -20,19 +20,28 @@ import "ga4gh/genotype_phenotype.proto";
...
+  // `SearchPhenotypesRequest` as the post body and will return a JSON version
+  // of `SearchPhenotypeAssociationSetsResponse`.
+  rpc SearchPhenotype(SearchPhenotypesRequest)
+      returns (SearchPhenotypesResponse);
+
+  // Gets a list of genotype-phenotype assoications accessible through the API.
+  // `POST /genotypephenotype/search` must accept a JSON version of
+  // `SearchPhenotypesRequest` as the post body and will return a JSON version
+  // of `SearchPhenotypeAssociationSetsResponse`.
+  rpc GetPhenotypeAssociations(SearchGenotypePhenotypeRequest)
```
david4096: `SearchPhenotypeAssociations` would be the RPC method that accepts a `SearchPhenotypeAssociactionRequest`. `Get` is reserved for RPC methods that perform the retrieve by ID task.

The endpoints are `/phenotypes/` `/phenotypeassociationsets/`, and `/genotypephenotype/`, when the naming convention suggests `/phenotypeassociations/`. A common convention is to name the endpoint after the message name. In that case we would name it `/featurephenotypeassociations/`, which is a bit of a mouthful. I'm neutral towards which is more correct, but it is important to choose one and not mix naming schemes. If the endpoint will be named differently from the message it returns, we can at least have internal consistency in the IDL.

src/main/proto/ga4gh/genotype_phenotype_service.proto
```diff
@@ -80,28 +89,34 @@ message EvidenceQuery {
...
 
 message SearchPhenotypesRequest {
     // The `PhenotypeAssociationSet` to search. Mandatory
     string phenotype_association_set_id = 1;
 
+    // Phenotype ID
     string id = 2;
 
-    string description = 3; /* The system should support regular expression matches */
+    // The system should support regular expression matches
```
david4096: Is this required? I'd like to keep the schemas as easy to implement and I don't think regex matching is needed.

bwalsh: @david4096  We have consistently gotten feedback from the g2p task team that some type of wildcard is desirable. The auxiliary verb "should" was used deliberately as opposed to "must".  This should give, I hope, enough room. 
Per RFC 2119 [should] "mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course."

bwalsh:
+1

src/main/proto/ga4gh/genotype_phenotype_service.proto
kozbo: I don't think there is doubt about the need for the ability to do a regex matching, it is just that this would be the only API in the set that uses this approach. So the comment is more about consistency.

kozbo: Suggest that we change the 'should' to 'may' and that we create enough issue to track the desire to remove this type of search from the schema. There is much that we can learn from the users of this API that will inform the evolution of this interface.

david4096: Two other options: remove the regex from the specification entirely as I believe we can satisfy the interchange use cases in its absence, or, allow regex matching on strings through the API. Just stating "regex" matching is probably not clear enough either, perhaps point to a more specific implementation of regular expressions? What features can they trust will be present? Grouping? Quantification?

Issue #684, opened 2016-08-16
Questionable use of snake_case (bio_sample_id instead of biosample_id)


mbaudis:
The use of snake case here is confusing, since `BioSample` is a first class object. This probably comes from the confusing use of camelCase for the `BioSample` object name. Since we generally accept "biosample" as proper noun and use it as such in the documentation (this is also the case in major resources), I propose we change:
- `BioSample` => `Biosample`
- `bio_sample_id` => `biosample_id`

dcolligan:
+1

kozbo:
+1

PR #705, opened 2016-09-01
Add Biosample ID to quantification

david4096:
This PR allows one to filter RNA quantifications by bio_sample_id.

Currently, RNA quantifications are not given a `bio_sample_id` field. RNA quantifications can be optionally associated with read group IDs, which in turn are tagged by biosample. This addition allows one to query RNA quantifications by `bio_sample_id` directly, retaining the redundancy in read groups. This is done by adding it to the quantification message and search request. @saupchurch

### DIFF
--- a/src/main/proto/ga4gh/rna_quantification.proto
+++ b/src/main/proto/ga4gh/rna_quantification.proto
@@ -46,6 +46,9 @@ message RnaQuantification {
  
   // Name
   string name = 2;
+  
+  // Biosample ID
+  string bio_sample_id = 8;
 
   // Description
   string description = 3;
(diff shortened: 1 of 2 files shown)


saupchurch:
Looks reasonable to me and covers the case where a read group ID may not be available.
+1

kozbo:
@dcolligan This will be coming next after the work you are doing on server #1368. Something to keep in mind.

mbaudis:
+1 (but please remember https://github.com/ga4gh/schemas/issues/684)

dcolligan:
+1 with @mbaudis 's proviso

kozbo:
@mbaudis we haven't forgotten #684, coming soon! Merging this first.

Issue #707, opened 2016-09-07
Biosample tissue type


david4096:
Biosamples are normally from a tissue sample that can be named. This data should be represented as a named field of a Biosample message, i.e. `tissue_type`. I suppose one might use something like the [Foundational Model of Anatomy](https://bioportal.bioontology.org/ontologies/FMA) to restrict the vocabulary. @mbaudis @nishill

mbaudis:
I am for solving this through an OntologyTerm object, as part of a general BioFeatures collection object which contains information private to the BioSample. Others would be histology, anatomic location, associated disease (of the ->tissue<- as a somatic phenotypic variation...). We seem to address this in discussions happening at the moment in the MTT, and on documents such as [BioFeature](https://docs.google.com/document/d/1aTK-hWdTPbSt3DhB-pW8jDvK5-krJW557Vsw6V3jG-s/edit#heading=h.w8b0lsmtgmwl)  and [use cases](https://docs.google.com/document/d/18K8TpU6_oxWCTLnmJlxSpv1dl00mFEYHpMfIRo4E6SQ/edit#heading=h.4yzkbieal421) documents.

mcourtot:
+1, adding suggestion to use [Uberon](http://uberon.github.io) for anatomical parts (it does map to FMA amongst others)

mdmiller53:
here's a snippet of how TCGA classifies tissues.  the total number of combinations is 188:
```
mysql> select sampletype, tissue_anatomic_site, tissue_anatomic_site_description, count(*) ct 
from metadata_biospecimen 
where tissue_anatomic_site in ('Head/Neck', 'Eye', 'Colon') 
group by sampletype, tissue_anatomic_site, tissue_anatomic_site_description;
+---------------------+----------------------+----------------------------------+-----+
| sampletype          | tissue_anatomic_site | tissue_anatomic_site_description | ct  |
+---------------------+----------------------+----------------------------------+-----+
| Primary solid Tumor | Colon                | NULL                             |   2 |
| Primary solid Tumor | Colon                | Ascending Colon                  |  98 |
| Primary solid Tumor | Colon                | Cecum                            | 110 |
| Primary solid Tumor | Colon                | Descending Colon                 |  22 |
| Primary solid Tumor | Colon                | Hepatic Flexure                  |  24 |
| Primary solid Tumor | Colon                | Splenic Flexure                  |   7 |
| Primary solid Tumor | Colon                | Transverse Colon                 |  38 |
[...]

PR #710, opened 2016-09-09
Biosample sample characteristics

mbaudis:
This PR addresses the need for a different structure and naming of the `Biosample.disease` attribute:

* there has been consensus in the metadata task team that an attribute name of `disease` is misleading in the context of `Biosample` and should be reserved for the `individual` object level
* a `Biosample` can use a number of ontologies, e.g. for patho-histology, anatomic location, tissue type ...; see also the discussion at https://github.com/ga4gh/schemas/issues/707

The renaming to sample_characteristics is in line with the use e.g. at GEO; however, the `sample_` prefix may not be strictly necessary. Alternatives welcome.

### DIFF
--- a/doc/source/api/biometadata.rst
+++ b/doc/source/api/biometadata.rst
@@ -37,7 +37,7 @@ Attribute             Notes
 *name*                * a human readable object label/identifier
                       * not to be used for referencing
 *description*         * additional, unstructured information about this BioSample
-*disease*             * OntologyTerm annotating the disease of the sample
+*sample_characteristics*             * List of OntologyTerm objects annotating the biological characteristics (e.g. anatomic location, patho-histology) of the sample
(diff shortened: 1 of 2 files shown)


david4096:
This makes sense to have a tag-bag field that strictly represents Ontology Terms, although the semantic context of those terms can be lost if no name is provided.

Please also consider the approach of allowing tagging via Ontology Term (and others) in a generic attributes field by upgrading the `info` field. This would allow a data curator to define named tag bags on a biosample with some more context without adding new named fields to the message. This replaces the info field and both approaches are not exclusive. PR for this feature [here](https://github.com/ga4gh/schemas/pull/700).

I'm +1 for this approach as it solves the immediate issue of specifying tissue type for TCGA data.

mbaudis:
Closed & moved to https://github.com/ga4gh/schemas/issues/711 (clean up).

PR #694, opened 2016-08-22
Change ontology term identifier

david4096:
Ontology terms had an ambigious 'id' field and description. This PR renames the field to term_id to remove the confusion over the role of the field, as well as improves the comment for its proper use. A field 'term_uri' has been added that allows one to specify the full URI for a term if it is available.

Closes #621 #165

PR #711, opened 2016-09-09
Biosample sample characteristics

david4096:
Reopened from #710 to remove extraneous commits. @mbaudis if you're happy with it you can close the other.

> This PR addresses the need for a different structure and naming of the Biosample.disease attribute:

>there has been consensus in the metadata task team that an attribute name of disease is misleading in the context of Biosample and should be reserved for the individual object level
a Biosample can use a number of ontologies, e.g. for patho-histology, anatomic location, tissue type ...; see also the discussion at #707
> The renaming to sample_characteristics is in line with the use e.g. at GEO; however, the sample_ prefix may not be strictly necessary. Alternatives welcome.

### DIFF
--- a/doc/source/api/biometadata.rst
+++ b/doc/source/api/biometadata.rst
@@ -37,7 +37,7 @@ Attribute             Notes
 *name*                * a human readable object label/identifier
                       * not to be used for referencing
 *description*         * additional, unstructured information about this BioSample
-*disease*             * OntologyTerm annotating the disease of the sample
+*sample_characteristics*             * List of OntologyTerm objects annotating the biological characteristics (e.g. anatomic location, patho-histology) of the sample
 *individualId*        * the *id* of the *Individual* this BioSample was derived from
 *created*             * the time the record was created, in ISO8601
 *updated*             * the time the record was updated, in ISO8601
--- a/src/main/proto/ga4gh/bio_metadata.proto
+++ b/src/main/proto/ga4gh/bio_metadata.proto
@@ -76,8 +76,15 @@ message BioSample {
   // The "description" attributes should not contain any structured data.
   string description = 4;
 
-  // OntologyTerm describing the primary disease associated with this BioSample.
-  OntologyTerm disease = 5;
+  // Zero or more OntologyTerm objects describing characteristics of this
+  // Biosample. These should not be used to annotate general characteristics
+  // of the individual this Biosample was derived from.
+  // Example: In a Biosample comprised of leukemic blasts from an individual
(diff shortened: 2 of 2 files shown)


david4096:
Copying response from #710:

This makes sense: to have a tag-bag field that strictly represents Ontology Terms, although the semantic context of those terms can be lost if no name is provided.

Please also consider the approach of allowing tagging via Ontology Term (and others) in a generic attributes field by upgrading the info field. This would allow a data curator to define named tag bags on a biosample with some more context without adding new named fields to the message. This replaces the info field and both approaches are not exclusive. PR for this feature here.

I'm +1 for this approach as it solves the immediate issue of specifying tissue type for TCGA data.

mdmiller53:
+1 for the merge

sarahhunt:
+1 Looks good to me.

david4096:
Seconding my +1. When loading the TCGA biosamples it appears some are labeled with more than one disease, making the singly valued `disease` field not especially helpful!

mbaudis:
So merge depending on the integration team. @david4096 @kozbo?! (If no objections to the attribute's name).

mcourtot:
Hi @david4096,

We talked further with @mbaudis and would like to propose replacing the *sample_characteristics* attribute with the *characteristics* object, with structure:

characteristics: [
      {
      description: “squamous cell carcinoma, base of tongue, stage 2”,
      type: phenotype (could be organism, disease...)
      repeated OntologyTerm ontologyTerms: [
          {
        ontologyId:  “http://purl.obolibrary.org/obo/DOID_0050865”,
        term:  “tongue squamous cell carcinoma”,
        },
        {
        ontologyId: “http://purl.obolibrary.org/obo/UBERON_0006919”,
        term:  “tongue squamous epithelium”,
        },
        {
        ontologyId:  “http://purl.obolibrary.org/obo/UBERON_0010033”
        term:  “posterior part of tongue”,
        },
        ],
      }
]

where each ontologyTerm is simplified in the above, but would in fact be the OntologyTerm structure as we agreed on at https://github.com/ga4gh/schemas/pull/694/files (and include version and source).

david4096:
Neat! That's certainly a more flexible way of describing characteristics! Could you take a look at https://github.com/ga4gh/schemas/pull/700 ? I believe it is attempting to perform a similar facility and would apply across the API. The additions you've made from what I can tell are restricting the characteristics to ontology terms, and providing a description and controlled vocabulary for the `type`.

mbaudis:
Following the discussions at Vancouver: Closing this in favour https://github.com/ga4gh/schemas/pull/725.

PR #698, opened 2016-08-24
G2p

dcolligan:
Merge of G2P into master

@bwalsh

PR #712, opened 2016-09-12
Bio metadata age

mbaudis:
This PR introduces an `Age` object, to address both quantitative and qualitative encoding of age (-like) data points, which has been a recurring topic in G2P and metadata.

Please see also the discussion at https://github.com/ga4gh/schemas/pull/698/files/11402061293da9dbd77327c7439f4ce7dd60c086#r78443366


david4096:
@mbaudis I presume you would like to add this field to the Individual message?

mbaudis:
@david4096 This was thought as a proposal for a prototype `Age` object, addressing the G2P problems as well as the needed (pre-configured in metadata branch, maybe different names etc.) 

* bio_metadata.Biosample.age_at_collection
* bio_metadata.Individual.diseases.age_of_onset

... etc. attributes. The need to do both OT & quantitative is consensus; here is a suggestion to get this finally done.

And yes, this impacts the G2P merge & the associated methods.

mbaudis:
@david4096 ... so, no, it is not part of Individual but general metadata; however, I dropped this into `bio_metadata.proto` since for the bio context. G2P currently does not seem to import this, only `metadata.proto`?!

kozbo:
I suggest that we put this on the list of items for the next round of G2P. It is not critical for basic functionality of a first release so I don't suggest that we hold the merge of #698 for this.

mbaudis:
@kozbo I agree that this should not block G2P; it can be discussed independently. We just have to make progress in the general schema but also to keep things aligned; and this PR also serves as a reminder of it.

From the metadata side, the next step is to use `Age` with various use case driven object implementations.

As a technical note: From the metadata side the attempt to do a "metadata branch" has totally failed to move the schema forward, and this is not due to a lack of structured content from MTT. Therefore now: Identifying required pieces & developing them one-by-one (we have so many thought-out concepts, which basically haven't changed over the last 1-2 years; so it is not an issue of untested ideas).

mbaudis:
@kozbo @david4096 This has now been amended with an implementation of `Age` for the `Biosample` object (`individual_age_at_collection`).

mbaudis:
@kozbo @david4096 Can we pick this up again? Also, not sure if the re-numbering of `info` is o.k.; so probably have to fix this?

david4096:
Closing in favor of https://github.com/ga4gh/ga4gh-schemas/pull/849

PR #713, opened 2016-09-14
update version to master-SNAPSHOT

bwalsh:
* No fixes or functionality changes
* Sets pom `version` tag for downstream activities (compliance)


bwalsh:
Cancelling after requirement clarification

PR #714, opened 2016-09-14
set schema version to 0.6.0-SNAPSHOT

bwalsh:
* No fixes or functionality changes
* Sets pom version tag for downstream activities (compliance)


david4096:
+1 thanks

PR #720, opened 2016-10-03
Added proto file for all services.

dkucsc:
When running swagger-codegen to generate server code, it only accepts 1 input file.

So if you want to a single server to handle multiple services, one method is to put all your services into 1 file.

This means that we'll have to keep this all_services.proto file up-to-date with any changes to any *.proto file, but this shouldn't be too expensive (they don't change frequently).

### DIFF
--- a/src/main/proto/ga4gh/all_service.proto
+++ b/src/main/proto/ga4gh/all_service.proto
@@ -1,0 +1,1231 @@
+syntax = "proto3";
+
+package ga4gh;
+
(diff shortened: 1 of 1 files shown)


david4096:
I've found this useful. Can you update it to include the g2p functionality?

david4096:
Closing, however, the feature is noted in the issue https://github.com/ga4gh/schemas/issues/776

PR #728, opened 2016-10-25
Allow descriptor type 12

david4096:
The doc generation process was broken due to an uncaught exception.

Fixes the currently breaking build. @dcolligan @bwalsh @saupchurch


david4096:
Thanks @mbaudis , it was a comment whitespace problem :/

PR #731, opened 2016-10-26
Put proto files under ga4gh_schemas directory

dcolligan:
This change will allow `protoc` to compile pb2 files in a way that fits our pypi packaging strategy.

Here is the repository that is packaging the `ga4gh_schemas` pypi package (this will be moved from my personal repo to a ga4gh repo soon): https://github.com/dcolligan/ga4gh-schemas

### DIFF
--- a/src/main/proto/ga4gh/schemas/ga4gh/allele_annotation_service.proto
+++ b/src/main/proto/ga4gh/schemas/ga4gh/allele_annotation_service.proto
@@ -1,10 +1,10 @@
 syntax = "proto3";
 
-package ga4gh_schemas.ga4gh;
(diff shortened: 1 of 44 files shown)


david4096:
Could you explain more why this is the case? Can we use our organization `ga4gh` as the package and class off of that. Is there an approach that allows us to make this change without restructuring the directories? Perhaps a script can make the namespace alterations.

dcolligan:
@david4096 the second commit in this PR implements the hierarchical namespace.

We could alternately pursue the script approach...

dcolligan:
Abandoned in favor of #740

PR #732, opened 2016-10-26
Note that reference names must be unique

### LINKED ISSUE #617: API specification allows non-unique reference names in a reference set
The documentation specifically allows for multiple reference names in a reference set.  Variants and reads link via reference names to a single reference.  It's also massively confusing to have two `chr1' chromosomes in a single reference set.

Reference should be document to require name a name to be unique withing a reference set.  Also update the intro documentation to indicate that some collections require unique names.

related to https://github.com/ga4gh/schemas/issues/518

david4096:
Reference names are expected to be unique within a reference set and this PR makes that explicit. Thanks @diekhans https://github.com/ga4gh/schemas/pull/620

Close #617

### DIFF
--- a/src/main/proto/ga4gh/references.proto
+++ b/src/main/proto/ga4gh/references.proto
@@ -21,8 +21,7 @@ message Reference {
   // excluding all whitespace characters (this is equivalent to SQ:M5 in SAM).
   string md5checksum = 3;
 
-  // The name of this reference. (e.g. '22'). This should be unique within the
-  // Reference Set.
+  // The unique name of this reference within the Reference Set (e.g. '22').
   string name = 4;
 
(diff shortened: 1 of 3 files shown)


src/main/proto/ga4gh/references.proto
```diff
@@ -19,7 +19,8 @@ message Reference {
   // excluding all whitespace characters (this is equivalent to SQ:M5 in SAM).
   string md5checksum = 3;
 
-  // The name of this reference. (e.g. '22').
+  // The name of this reference. (e.g. '22'). This should be unique to the
```
saupchurch: :%s/to/within

david4096:
@ejacox want to merge this documentation only PR?

PR #735, opened 2016-10-27
Change genotype to List value

### LINKED ISSUE #727: Call schema not suited for variant calls with no coverage
As recommended by @dcolligan, I am adding this issue here to describe a problem that I found on GA4GH schemas and was first described at ga4gh/server#1389.

When the VCF file has variant calls with no coverage represented by `./.` the GA4H API server cannot convert this data to the appropriated value. The solution I found was change [this line](https://github.com/ga4gh/schemas/blob/master/src/main/proto/ga4gh/variants.proto#L121) to:

```
google.protobuf.ListValue genotype = 3;
```

After compiled protobuf files and updated the server it worked fine.

david4096:
This extends the possible values for the genotypes array to better support when no call is given.

Close #727 Thanks @Welliton309


mbaudis:
+1

kozbo:
+1

kozbo:
tested all the way through compliance, working for me. +1

PR #739, opened 2016-11-04
Coverage Service

david4096:
A common concern prospective API users have raised is that we do well interchanging low level data, but most times an application won't make use of it. A very common case is returning binned counts of elements over a genomic region. Some visualizers will make counts of elements themselves at the client-side, however, because of network transfer costs oftentimes one might just want the counts and not the data themselves.

These coverage bins provide a data model for read depth, variation density, or genomic feature density. With the addition of this endpoint, existing genomic data can be rendered in visualizers much more effectively. Thanks @akmorrow13!

Remove info from bin
Add more comments
Renamed width to bin_width in CoverageRequest

### DIFF
--- a/src/main/proto/ga4gh/coverage_service.proto
+++ b/src/main/proto/ga4gh/coverage_service.proto
@@ -1,0 +1,93 @@
+syntax = "proto3";
+
+package ga4gh;
+
+import "google/api/annotations.proto";
+
+// The coverage service returns binned counts for on requested containers.
+// It is designed to provide intermediate scale views of the data for genome
+// browsers.
+service CoverageService {
+  // Gets a list of `Dataset` matching the search criteria.
+  //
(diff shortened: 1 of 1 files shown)


src/main/proto/ga4gh/coverage_service.proto
```diff
@@ -0,0 +1,93 @@
...
+import "google/api/annotations.proto";
+
+// The coverage service returns binned counts for on requested containers.
+// It is designed to provide intermediate scale views of the data for genome
+// browsers.
+service CoverageService {
+  // Gets a list of `Dataset` matching the search criteria.
+  //
+  // `POST /datasets/search` must accept a JSON version of
+  // `SearchDatasetsRequest` as the post body and will return a JSON
```
achave11: It needs to be fixed!

PR #740, opened 2016-11-15
Add ability to create python package of schemas

dcolligan:


david4096:
This is great!

This addition will make it so that releases of the compiled python protobuf bindings can be released to pypi and easily integrated into other code bases, regardless of whether they implement a HTTP API, or would simply benefit from the GA4GH data model. In order to support this, some modification of the proto is done before packaging. Would you please say a little more about the problems you faced in creating the packaging scripts and document usage?

PR #741, opened 2016-11-17
Move ga4gh proto files up one directory

dcolligan:
Changed 'google' dir to 'elgoog' because of namespace collision

Issue https://github.com/dcolligan/ga4gh-schemas/issues/7


david4096:
Thanks @dcolligan, we're going to take a closer look at this in a future version. The current import line will state `import ga4gh.schemas.ga4gh.variants_pb2`.

dcolligan:
scrapping this after a conversation with @david4096 ... salvageable parts are in #743

PR #744, opened 2016-11-21
Move setup

dcolligan:


dcolligan:
Depends on #743

@david4096 this gives you the functionality you want (in, say, server directory):
```
$ pip uninstall -y ga4gh-schemas && pip install ../schemas && ga4gh_run_tests
```

david4096:
Nice one on figuring out the google imports! @kozbo this will need your approval as it alters the release process. +1 from me, we'll want to follow this with some python specific usage documentation.

PR #746, opened 2016-11-28
Improve installation process via pip

### LINKED ISSUE #745: Unify python install paths
There are currently two python install paths, one for generating documentation and the other for creating the protobuf bindings. The dependencies are in `python/requirements.txt` and `./requirements.txt` respectively. This issue will close when there is a single install path for python that removes the duplicated requirements.

david4096:
In addition to moving to a stable version of protocol buffers, this PR attempts to improve the build process to close #745 

* Generates compiled protocol buffers during `setup`
* Still requires `protoc` be installed locally
* Updates version of protocol buffers in travis
* Merge `requirements.txt` to `python` directory
* Removes process_schemas dependency on `ga4gh.common` to bootstrap setup
* Set protocol version in `setup.py`

Allows the compiled schemas to be installed using the `pip` interface. For example, `pip install git+git://github.com/david4096/schemas.git@protobuf31` will install this branch of the schemas to your local environment.

The goal is to easily develop against development branches of multiple packages using requirements "constraints". These will direct the installer to resolve dependencies using developer provided git branches. @dcolligan @ejacox


ejacox:
+1 
It worked for me and I found it useful.
I did 'pip install -r dev-requirements -c constraints.txt' in the server, which installed the schemas directly from github.

setup.py
```diff
@@ -59,3 +137,5 @@
     # Use setuptools_scm to set the version number automatically from Git
     setup_requires=['setuptools_scm'],
 )
+
+shutil.rmtree('package_python')
```
dcolligan: Check for existence of directory before removing it; remove it only if it exists, otherwise this will throw an error (perhaps not a likely scenario, but ya never know...)

setup.py
```diff
@@ -23,10 +93,19 @@
         pinnedVersion = line.split()[0]
         install_requires.append(pinnedVersion)
 
+tempPath = 'package_python'
+try:
+  shutil.rmtree(tempPath)
```
dcolligan: this and two lines below looks like insufficient indentation...

dcolligan:
I don't like putting things in `python/requirements.txt` that are not dependencies of the `ga4gh_schemas` package.  Your change will install, for instance, `nose` when someone tries to install `ga4gh_schemas`, and that doesn't seem right.

That and a couple nits I left in comments aside, this looks good.

david4096:
Thanks @dcolligan and @andrewjesaitis. I've fixed the indentation and safely deleted the `package_python` directory. I also moved the tests and documentation generation related items to the `dev_requirements` so the base install is smaller. My hope is that this will make it much easier to develop against the schemas from python!

dcolligan:
+1

kozbo:
Looks good to me too
+1

PR #751, opened 2016-12-07
Pip installation fixes

david4096:
The last setup.py worked because of the presence of the generated files in the `python` directory, but didn't transfer over the `_protocol_version` properly. In changing the script to move over the file, I opted to use the `python` directory to include the generated python and just ignore the `_pb2` files. This simplifies the installation by removing the need for a temporary directory, which is what was causing the problem. This also makes development easier. After checking out the tree, to generate the protobuf bindings, do `python setup.py develop`. This will add the generated files so you can run the tests.

This method of using a staging directory was also failing when using setup.py from another package.

You can install this PR and test it out using:

```
pip install git+git://github.com/david4096/schemas@pipfixes
```

I ran into this problem when working on ways of managing dependencies against git branches using setuptools and pip. The problem was that the `setup.py` call to install the schemas from server would execute the setup script for schemas, but the resulting package didn't contain the generated protobuf. When installing from pip this wasn't a problem. I understand pip to use setup.py to do its installation so I found this confusing.

In the end this PR keeps the development pattern within setup.py. Instead of packaging the compiled protobuf in a separate directory before packaging, you run `setup.py develop` (or install) to add it to the working python directory. This makes the contents of the resulting package more clear and development easier.

Compiling protocol buffers schema to python and putting it in a python namespace is a generally solvable problem. Perhaps we can package the process schemas script as a separate module.


ejacox:
I tested it by adding `git+git://github.com/david4096/schemas@pipfixes` to requirements.txt from where I am working, e.g. server. 
+1

PR #762, opened 2016-12-21
278 maven repository

### LINKED ISSUE #278: Release artifacts to Maven Central repository
In order to release artifacts to the Maven Central repository, the following should be done
- Add `<plugin-management>` section, Maven enforcer plugin configuration, and release profile to `pom.xml`
- Add `<developers>` section to `pom.xml`
- Create a new project on [Sonatype OSS Nexus](https://oss.sonatype.org/) for groupId `org.ga4gh`
- Add `KEYS` file with public code signing keys of all developers interested in the role of release manager
- Deploy current version `0.6.0-SNAPSHOT` to Sonatype OSS Nexus snapshots repository
- Manually sign and release artifacts from 0.5 and 0.5.1 releases to Sonatype OSS Nexus releases repository
- When version 0.6.0 is ready, release to Sonatype OSS Nexus releases repository and enable rsync to Maven Central repository

I am willing to take these on.  Would any one else be interested in the role of release manager?  @massie @fnothaft?

ejacox:
Updated maven repository release notes. Close #278


heuermh:
Sorry, I don't follow what is happening here, nor the motivation.  Can someone explain?  The version set in pom.xml (`0.6.0a`) does not match that used elsewhere (`0.6.0-a9`).

fnothaft:
+1 @heuermh's comments. Manually modifying the POM and then deploying is not best practice, and is a step backwards relative to the previous release scripts.

ejacox:
What is wrong with modifying the POM? 

The previous release script didn't work. How is this a step backwards?

heuermh:
> What is wrong with modifying the POM?

The only time the version in `pom.xml` should be the same as a release version is on a commit with a release tag or on a release branch.  Otherwise anyone building locally or depending on git HEAD (which if I understand correctly is how ga4gh server works, i.e. https://github.com/ga4gh/server/blob/master/constraints.txt#L27) can build artifacts with the same version as a release but different code.  Those artifacts would fail checksums and signature validation and could break downstream code.

The Maven build ecosystem (e.g. Maven, Ivy, Gradle, SBT, etc.) uses `-SNAPSHOT` version numbers to indicate work-in-progress towards the next release version number.  Git HEAD should always be set to a `-SNAPSHOT` version number.

> The previous release script didn't work. How is this a step backwards?

I've been volunteering to help with this since Apr 2015 if not earlier, with repeated offers to help since then.  "didn't work" how?  I can be available via irc or gitter or slack or on the phone or whatever it takes.

fnothaft:
+1 @heuermh; as stated many times before, I'm also glad to hop on the phone/etc to sort this out. The prior scripts are derived from the release scripts we have used to cut ~10 releases in the past year across 4 different projects. As such, I can say with full confidence that any bugs that were in the script are minor mistakes that were introduced when migrating the scripts to a new repository.

fnothaft:
Apologies for being a pain, but we've been blocked wrt supporting the GA4GH schemas for more than 6mo due to the lack of proper release artifacts (we've had to cut our own forked release to support 0.5.1, even) and we have repeatedly offered support (the original scripts and docs, hopping on calls, email threads, etc) with the release workflow. Can we properly resolve this release workflow issue so that we can move on with downstream development? Perhaps it'd be most productive to get on a call to circle up?

kozbo:
Thanks @heuermh and @fnothaft for all the help and comments, and you are not "a pain" :-). We now have someone to chase this one down (@ejacox) so progress will be made. Looks like the first approach isn't quite right so yes, let's set up a call this week. Let me check in with Edwin tomorrow and then suggest some times.

fnothaft:
Thanks @kozbo and @ejacox, I look forward to working together to sort this out.

scripts/release/README.md
```diff
@@ -3,20 +3,21 @@ Notes for release managers
...
+Setup:
+1. Add the GPG key to the KEYS file.
+1. Copy (or incorporate) the settings.xml file to ```~/.m2/settings.xml``` and
+   add your GPG key and sonatype login.
 
-Then from the project root directory, run `./scripts/release/release.sh`.
-If you have any problems, run `./scripts/release/rollback.sh`.
+Release:
+1. Update the version in pom.xml.
+2. mvn deploy
```
heuermh: In order to produce -source.jar, -javadoc.jar, and gpg signatures, this should be `mvn -Prelease deploy`

PR #765, opened 2017-01-03
added timestamps to metadata

### LINKED ISSUE #626: Extend dataset record structure to include fields used by Beacon
Beacon has some fields that might be good to add to the generic dataset record. I suggest adding the following:

@namespace("org.ga4gh.consentcode") form [here](https://github.com/ga4gh/beacon-team/blob/develop/src/main/resources/avro/consentcode.avdl)
- this seems very useful

created and modified date stamps from [here](https://github.com/ga4gh/beacon-team/blob/develop/src/main/resources/avro/beacon.avdl) - see BeaconDataset
"""
  /** The time the dataset was created (ISO 8601 format). */
  string createDateTime;

  /** The time the dataset was updated in (ISO 8601 format). */
  string updateDateTime;
"""
- note the lack of default values

Asking a broader question, how about making the Dataset record be the same as the BeaconDataset record? I think that some will complain about adding the assembly ID to the Dataset structure as there could be multiple references used in the data that is collected on this server. We might solve this by stating that there can be multiple Datasets on a server (I believe that this is already true) and then we could require that each Dataset be for only one reference assembly id. 

I am still a bit new to the schemas so I apologize if I am off base here.

ejacox:
Create and update timestamps were added to the dataset message in the metadata proto.
Closes #626.

### DIFF
--- a/src/main/proto/ga4gh/metadata.proto
+++ b/src/main/proto/ga4gh/metadata.proto
@@ -43,6 +43,12 @@ message Dataset {
 
   // A map of additional dataset information.
   map<string, google.protobuf.ListValue> info = 4;
+
+  // The time the dataset was created (ISO 8601 format).
+  string createDateTime = 5;
+
+  // The time the dataset was updated in (ISO 8601 format).
+  string updateDateTime = 6;
 }
 
 // A Program describes software used in data processing or analysis.


david4096:
Can we close in favor of https://github.com/ga4gh/schemas/pull/690 ? It implements the same timestamps across the board and uses the much less painful `created` and `updated` field names. Sorry for the confusion!

ejacox:
I agree. Close this and the pull requests in compliance and server.

mbaudis:
+1 for https://github.com/ga4gh/schemas/pull/690

ejacox:
Also close 682, 568, and 685.

mbaudis:
Closing this in favour of #690.

PR #772, opened 2017-01-09
Extending the variant model for DUP, DEL structural variants

mbaudis:
This is a first attempt to address the need for representation of
structural variants. It is chiefly aimed to encode sequence
duplications/deletions, and follows examples from VCFv4.2.
The first commit does not yet implement the services part or documentation.

For some of the discussions, please see issue https://github.com/ga4gh/schemas/issues/752 .


david4096:
Awesome @mbaudis! Looking forward to working with this in practice. Being able to easily interpret the type of the variant always seemed a useful canonical field to me! What do you think of making the type an enumeration as opposed to a string?

mbaudis:
@david4096 I am neutral regarding enum vs. string; both need some way to document permitted values. This is mostly a schema design decision (we've been going back and forth, with the first version having enums, then those being discouraged ...).

For the code integration team to edit/decide (so go ahead ...).

src/main/proto/ga4gh/variants.proto
```diff
@@ -191,4 +191,29 @@ message Variant {
   // determination of genotype with respect to this variant. `Call`s in this
   // array are implicitly associated with this `Variant`.
   repeated Call calls = 12;
+
+  // The "variant_type" is used to denote e.g. structural variants.
+  // Examples:
+  //   DUP  : duplication of sequence following "start"; not necessarily in situ
+  //   DEL  : deletion of sequence following "start"
+  string variant_type = 13;
```
sarahhunt: Can we use SO terms (eg SO:0000159, deletion) here?

src/main/proto/ga4gh/variants.proto
```diff
@@ -191,4 +191,29 @@ message Variant {
...
+
+  // Length of the - if labeled as such in variant_type - somatic variation.
```
sarahhunt: Should that be 'structural' rather than 'somatic'?

mbaudis: Correct - shows my subconscious bias...

src/main/proto/ga4gh/variants.proto
mbaudis: In that case, i.e. when not working with a limited (by doc, enumeration) list of terms =>
```
ontologyTerm variant_type = 13;
```
... would be the appropriate way to do it. Alas: overhead? @kozbo, @david4096 ?

david4096: +1 don't write an enum if someone else already has :).

Offering in the proto the variety of variants would be very helpful to other implementors.

```
duplication    (CURRENT_SVN)
SO Accession: 	SO:1000035 (SOWiki)

structural_variant    (CURRENT_SVN)
SO Accession: 	SO:0001537 (SOWiki)

insertion    (CURRENT_SVN)
SO Accession: 	SO:0000667 (SOWiki)
```

mbaudis: @david4096 But not all of the variant types may (well, will) be supported by the schema (yet). So:

* you need either A) a positive (i.e. defining a set of) or B) a selective (i.e. which of all defined in external reference) variants would constitute supported values
* case A would naturally be an enumeration, B a documentation case with ontologyTerm (+ service)
Both have conceptual pros and cons, with in A ease/stringency, in B beauty/future speaking for them.

???

david4096: You mean you want to use a controlled vocabulary when it's available, and an open string otherwise? In that case we can make the field a oneof a string or ontology term. This makes constructing the message a little more baroque, but captures the cases nicely.

mbaudis: @david4096 Thought about that, but also the "baroque" :-)

I am open to any solution, as long as it is accepted & forward looking, and gets implemented. The implementation of other variant types is extremely important - SNVs are just a small subset (regarding genome coverage, functional impact; don't get blinded by their count number...).

mbaudis: @david4096 I actually had offered this as 2 ways to treat the problem, not as additive solution. You'll still have to document the allowed values in the OT object...

david4096: @mbaudis Great, sounds good to me! Variant callers want to be able to describe their own variant types and they use their own plain strings to do so. Giving the option of upgrading to a controlled vocabulary is something we might consider doing in other places in the API as well.

It is actually a very useful way to understand how the science moves forward: as new string values for variant type become popular, the sequence ontology has a strong case for reviewing adding it to their vocabulary. 

VCF spec has `SVTYPE` https://samtools.github.io/hts-specs/VCFv4.2.pdf using BND in the type. I believe these need to be interchanged as well. This post is very helpful on the topic of SV calls: https://simpsonlab.github.io/2015/06/15/merging-sv-calls/ .

david4096:
Closed in favor of https://github.com/ga4gh/schemas/pull/827

PR #656, opened 2016-07-08
Add quality and filters close #219

david4096:
Updates the work presented by @heuermh in https://github.com/ga4gh/schemas/pull/304 and @johntbates in https://github.com/ga4gh/schemas/pull/220 to protobuf. Includes the filters_passed field as suggested by @fnothaft 

Closes #219

PR #777, opened 2017-01-13
Add filters_applied, filters_passed, and filters_failed to variant

heuermh:
See https://github.com/ga4gh/schemas/pull/656#issuecomment-272273999


david4096:
Thank you kindly! We try to avoid renumbering tags or reusing tags. It's not important that the tag numbers are in order, but if a field hasn't changed it's tag number shouldn't change either. This helps guarantee that clients at future versions of the protocol can still handle messages, and field names can be reused safely.

heuermh:
Got it, I'll fix and rebase

heuermh:
Addressed review comments and squashed commits.

david4096:
This PR is awaiting implementation.

heuermh:
Sorry, what do you mean by implementation?

Note that if any code implementation relies on htsjdk, it will run into unresolved issue:
Genotype requires a method filtersWereApplied (https://github.com/samtools/htsjdk/issues/741)

Let me know if I should rebase and force push to address the necessary merge.

david4096:
Hi @heuermh! We look forward to merging this one but need to include it in our existing implementation before we can move forward. Leaving that comment as a reminder to the group that we should implement this sooner, rather than later. That implementation relies on pysam https://github.com/ga4gh/server/issues/1539 .

heuermh:
I see, thanks!

david4096:
Closed in favor of https://github.com/ga4gh/schemas/pull/827

heuermh:
Thanks!

Issue #774, opened 2017-01-09
pip install ga4gh-schemas fails


kozbo:
Steps to reproduce:
```
virtualenv schema-test; source schema-test/bin/activate
pip install ga4gh-schemas --ignore-installed --no-cache-dir
```
then you will get this error:
```
(jan17) KO-Mac:server kozbo$ pip install ga4gh-schemas --ignore-installed --no-cache-dir
Collecting ga4gh-schemas
  Downloading ga4gh_schemas-0.0.9.tar.gz (2.1MB)
    100% |████████████████████████████████| 2.1MB 2.4MB/s 
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/xn/1bh5_q793113z57s4g9kr0wr0000gn/T/pip-build-vU6vnu/ga4gh-schemas/setup.py", line 37, in <module>
        process_schemas.main([PROTOCOL_VERSION, 'python'])
      File "scripts/process_schemas.py", line 222, in main
        pb.run(parsedArgs)
      File "scripts/process_schemas.py", line 207, in run
        protoc = self._getProtoc(destination_path)
      File "scripts/process_schemas.py", line 169, in _getProtoc
        protocs))
    Exception: Can't find a good protoc. Tried [u'/private/var/folders/xn/1bh5_q793113z57s4g9kr0wr0000gn/T/pip-build-vU6vnu/ga4gh-schemas/python/protobuf/src/protoc']
```
Note that the same error occurs for schemas version 0.6.0a9, but not version 0.0.8, which depended on the older version of protobuf. So the error seems to be connected with the dependency on the new protobuf package.

@dcolligan I am hoping that you can sort out this dependency problem. The release is blocked.

david4096:
You do need to have the protocol buffers compiler installed if you want to generate your own bindings, but we shouldn't place that requirement on the release. When you do the release you can run protoc locally on your system and comment out the lines in `setup.py`.

kozbo:
It makes sense to require developers who want to use the pip installable schemas to have protoc pre-installed (though we should inform them of this), but there are other pathways that try to install the schemas that do not require the protoc compiler. Document generation in the virtual environment provided through Read The Docs is one example (fails in the same way). Someone installing the ga4gh-client (which also fails now) library is another, perhaps more important example.

dcolligan:
[...]

PR #778, opened 2017-01-13
Fix schemas install by removing protoc call

dcolligan:
Issue #774


kozbo:
A tentative  +1 on this. My testing got beyond the problem this fix was targeted to address (protoc complier dependency) but then I ran into another sphinx build error from RTD: 
```
Exception occurred:
  File "/home/docs/checkouts/readthedocs.org/user_builds/ko-ga4gh-server/envs/latest/lib/python2.7/site-packages/ga4gh/schemas/ga4gh/metadata_service_pb2.py", line 16, in <module>
    from ga4gh import metadata_pb2 as ga4gh_dot_metadata__pb2
ImportError: cannot import name metadata_pb2
```
That pb2 file is in place in the tree so I am a bit mystified.  This needs to be worked out before I can merge this PR. 
@dcolligan does a `make` in the server docs directory work for you after this change?

dcolligan:
It doesn't work, but I get a different error:
```
Exception occurred:
  File "/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/ga4gh/schemas/ga4gh/metadata_service_pb2.py", line 17, in <module>
    from google.api import annotations_pb2 as google_dot_api_dot_annotations__pb2
ImportError: No module named api
```
Thought it might be because our "google" dir and the top-level google package might be getting confused, so I tried changing everything to elgoog, but that wasn't the issue... still investigating...

By the way, this is why having a cross-repo testing strategy is a good idea!  https://github.com/BD2KGenomics/ga4gh-integration/issues/41

dcolligan:
I think something's wrong with the packaging.  In the server after an install of the `ga4gh-schemas` package, I can do:
```
>>> import ga4gh.variants_pb2
>>> import ga4gh.schemas.ga4gh.variants_pb2
>>> ga4gh.variants_pb2.__file__
'/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/ga4gh/variants_pb2.pyc'
>>> ga4gh.schemas.ga4gh.variants_pb2.__file__
'/usr/local/lib/python2.7/site-packages/ga4gh/schemas/ga4gh/variants_pb2.pyc'
```
Whereas it seems like only the later import should work; I'm not sure why anything gets installed in the `Cellar` directory...

All the schemas tests pass, so the culprit is probably in `setup.py`, I would wager... still investigating...

kozbo:
@david4096 Can you have a look at this? You have worked with he sphinx build more than anyone else at this point :-)

david4096:
We're working on it!

dcolligan:
Closed in favor of #781

PR #781, opened 2017-01-18
Remove protoc call from install path

dcolligan:
Remove protoc call from install path

Issue #774


dcolligan:
@kozbo @david4096 this branch works with client and server

dcolligan:
@david4096 if you want to test locally you can just mv the `protoc` executable to some other name temporarily and try a fresh install of the schemas

dcolligan:
Huh, you can push commits to my branch, @david4096 ?  Didn't know that was possible

david4096:
@dcolligan Yeah, unless you protect them specifically, I believe. I think if you are OK with it, I can think of a couple cases when a pushed commit makes more sense than waiting for a review: typos, changing dependencies in constraints, general documentation. Let's work together faster!

david4096:
As suggested by @dcolligan I tried running the installation locally when protoc was not available and observed a successful build.

We may like to add some more scaffolding to give better control over when one intends to rebuild the compiled protobuf. Currently, my best practice is running `python setup.py develop`.
```
  Running setup.py install for ga4gh-schemas ...     Running command /home/david/git/server-1/env/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-9G1i7P-build/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-yBy24X-record/install-record.txt --single-version-externally-managed --compile --install-headers /home/david/git/server-1/env/include/site/python2.7/ga4gh-schemas
    Couldn't find a good protoc, using precompiled protobuf.
```

Issue #780, opened 2017-01-17
Eliminate warnings in sphinx build


dcolligan:
Lots of lines are spewed out in the doc build like `/home/travis/build/dcolligan/schemas/doc/source/api/genotypephenotype.rst:340: WARNING: Title underline too short.` ... it would be nice to get rid of these.

dcolligan:
Fixed in #784

PR #784, opened 2017-01-19
Get rid of warnings/errors spam in build

dcolligan:
Issue #780

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@ build
 _version.py
 # Ignore proto included in the generated python package
 python/**/*.proto
+# sphinx error log
+err.log
 
 #********** windows template**********
 
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,10 @@
 # enable more robust bash
 SHELL:=/bin/bash -beEu -o pipefail
 BUILD_DIR:=target/doc
+ERR_LOG:=2> err.log
+ifdef verbose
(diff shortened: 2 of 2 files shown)


tools/sphinx/protobuf-uml.py
```diff
@@ -181,10 +181,8 @@ def write_graph(fields, containments, nests, matched_references, clusters):
 
     # Define edge properties for containments
     graph += "edge [\n"
-    graph += "\tdir=both\n"
```
david4096: What's happening here? Are they ignored values?

dcolligan: yeah, I deleted everything that the warning output said it was ignoring

dcolligan:
@david4096  If one wants to see errors / warnings from the sphinx build, can't one just turn them back on?  Errors and warnings are no good if everyone habitually ignores them.

david4096:
Add a flag that makes it easy for the current logging to happen to change the default. I think straight passing down the arguments to `make docs` might be fastest/easiest, but if you want to add a `verbose` key, no argument here.

dcolligan:
@david4096 flag added.  Get the warning output by doing: `make docs verbose=true`

david4096:
Thanks, we don't deserve to take credit for all those warnings when we're trying to parse other people's comments :).

PR #793, opened 2017-01-23
Use common

dcolligan:
Depends on #792


scripts/process_schemas.py
```diff
@@ -14,7 +14,8 @@
 import fnmatch
 import re
 import argparse
-import shlex
+
+import ga4gh.common.utils as utils
```
david4096: My bad, I had removed this specifically for the reason in #797 and should have mentioned it on this PR.

Issue #797, opened 2017-01-24
Remove process_schemas from setup.py


dcolligan:
Now that we are depending on `ga4gh-common` for `process_schemas.py` which is used in `setup.py`, the install can fail if the common package is not available before install happens (see below).  The solution is to remove any references to `process_schemas.py` from `setup.py`.

(This problem is post-0.6.0a9-release, so it doesn't affect our released packages, only development.)

```
(server-env) $ pip install -r dev-requirements.txt -c constraints.txt --upgrade
Collecting ga4gh_common from git+git://github.com/ga4gh/ga4gh-common.git@master#egg=ga4gh_common (from -c constraints.txt (line 27))
  Cloning git://github.com/ga4gh/ga4gh-common.git (to master) to /private/var/folders/62/16flnsrx3dj9qxqtf4djg16m0000gn/T/pip-build-AJEjgG/ga4gh-common
Collecting ga4gh_schemas from git+git://github.com/ga4gh/schemas.git@master#egg=ga4gh_schemas (from -c constraints.txt (line 28))
  Cloning git://github.com/ga4gh/schemas.git (to master) to /private/var/folders/62/16flnsrx3dj9qxqtf4djg16m0000gn/T/pip-build-AJEjgG/ga4gh-schemas
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/62/16flnsrx3dj9qxqtf4djg16m0000gn/T/pip-build-AJEjgG/ga4gh-schemas/setup.py", line 10, in <module>
        import scripts.process_schemas as process_schemas
      File "scripts/process_schemas.py", line 18, in <module>
        import ga4gh.common.utils as utils
    ImportError: No module named ga4gh.common.utils
```

dcolligan:
Example of this failure: https://travis-ci.org/ga4gh/server/builds/194838639

david4096:
Yeah, I specifically deprecated the common elements of process schemas to avoid this. I think it's OK to duplicate code if it's needed to perform installation like this.

I would keep process schemas as something simple that can run inside setup.py with no dependencies.

dcolligan:
Ok, this issue has now changed to: fix the install path by removing references to ga4gh-common from the install path.

dcolligan:
Fixed (as it were) in #798

PR #798, opened 2017-01-24
Remove common dependency from install path

dcolligan:
Issue #797

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+exclude = *_pb2.py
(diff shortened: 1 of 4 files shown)


david4096:
I like having process schemas run with setup. This makes schema development sensible and much easier.

dcolligan:
Ok! this PR has changed then.  Now it preserves the current installation mechanism, but removes the common library from being referenced on the install path.  I put a big fat comment at the top of `process_schemas.py` so nobody is tempted to do this again. @david4096

david4096:
Great!

Issue #782, opened 2017-01-18
Fix test_compile.py


dcolligan:
After #781 is merged, fix `test_compile.py` and incorporate other relevant changes from #778

dcolligan:
Fixed in #799

PR #799, opened 2017-01-25
782 test compile

dcolligan:
Fixes test_compile.py

Issue #782

Depends on #798


dcolligan:
@david4096 can you review?

PR #802, opened 2017-01-27
Bigwig

### LINKED ISSUE #769: Continuous Data
This issue was created in order to define a continuous data message type.

Continuous, or signal data, associates a value with sequence base pairs. This can be raw experimental data, such as ChIP-Seq data, or calculated values, such as conservation scores. There can be gaps (unsampled bases) in the values, so more than a simple array of values with a start point is needed. Such data is commonly stored in wiggle, bigwig, and bedgraph formats. In addition to just listing values, the formats can do some simple compression: either 'step', which can more compactly represent regularly sampled data, or 'span', which can more compactly represent windowed data.

There are at least a couple of points to discuss:

1. Where does it go?

From PR #591
"The name was chosen because SequenceAnnotations will include more than Features; The second data type is continuous (wiggle) was not include in this PR to speed implementation."

From this discussion, it seems that it should be a part of the sequence_annotations. However, most reference to "sequence annotations" refer to genomic annotations, i.e. associated with genes, which does not apply here. The continous data could be its own service in that case.

2. What should it be called?

Wiggle doesn't seem appropriate. 'Continuous' has been used as a place holder.

A draft schema follows:

`// This protocol defines a format for exchanging continous valued signal data,
// such as those produced experimentally (e.g. ChIP-Seq data) or through 
// calculations (e.g. conservation scores). It can be used, for example,
// to share data stored in Wiggle, BigWig, and BedGraph formats.
//
// Only bases with a signal are represented. Gaps in the values indicate bases
// with no signal.
//
// Step and span from the wiggle format are used for a simple compression.

// A chunk of continuous data. Due to gaps in the signal, the values
// cannot be represented by a single array, but require a set of arrays.
message Continuous {
  // The start position at which this signal occurs (0-based). This
  // corresponds to the first base of the string of reference bases. Genomic
  // positions are non-negative integers less than reference length.
  int64 start = 1;
 
  // If not one, values are defined for every 'step' base, leaving
  // gaps of undefined regions.
  int32 step = 2;

  // The number of bases each value spans. For example if span is 5,
  // then the first value is defined over start plus the next 4 bases, the
  // second value over the following 5 bases, ...
  int32 span = 3;

  // The data values.
  repeated double values = 4;
}

// A set of continuous data.
message ContinuousSet {
  // The ID of this continuous data set.
  string id = 1;

  // The ID of the dataset this set belongs to.
  string dataset_id = 2;

  // The ID of the reference set which defines the coordinate-space for this
  // set of data.
  string reference_set_id = 3;

  // The display name for this dataset.
  string name = 4;

  // The source URI describing the file from which this set was
  // generated, if any.
  string source_uri = 5;

  // Remaining structured metadata key-value pairs.
  map<string, google.protobuf.ListValue> info = 6;
}`

ejacox:
This PR adds the Continuous message to SequenceAnnotations in order to represent continuous valued data. It closes #769.


david4096:
Will you please add the import of the Continuous Sets to protocol.py?

https://github.com/ga4gh/schemas/blob/master/python/ga4gh/schemas/protocol.py

ejacox:
Continuous Sets is part of Sequence Annotations, which is in protocol.py.

david4096:
An excellent addition to the protocol! 

Looking forward to seeing applications and example code! For now, here is a simple example using this protocol: https://gist.github.com/david4096/ac3813cbcbc913074e95b7c9519a38d3

We expect to continue development to provide resolution and perhaps summary statistics for Continuous data https://github.com/ga4gh/schemas/issues/820 . Thanks @ejacox !

PR #807, opened 2017-02-06
Integrating external identifiers

mbaudis:


src/main/proto/ga4gh/bio_metadata.proto
```diff
@@ -48,8 +48,15 @@ message Individual {
...
+  // External identifiers representing this individual.
+  // TODO: Define if those are a different representation of the same
+  // record, or also can refer to instances containing this record, e.g.:
+  //  - publication identifiers with this record
+  //  - experimental series, datasets containing this record
+  repeated ExternalIdentifier external_identifiers = 10;
+
   // A map of additional information regarding the Individual.
-  Attributes attributes = 10;
+  Attributes attributes = 11;
```
kozbo: We cannot renumber existing fields. New fields should get the next higher unused number. The numbering will appear out of order but renumbering a field breaks compatibility with earlier implementations of the API.

mbaudis: Thanks @kozbo. Was confused between re-numbering and re-using. Pls. check.

david4096:
Thanks @mbaudis we can definitely offer some light federation this way! Would you be so kind as to look over https://github.com/ga4gh/schemas/pull/761 ? It applies this concept across the API.

Since we have proposed adding the `metadata` field in the [Timestamps PR](https://github.com/ga4gh/schemas/pull/690), we might use it for this `external_identifiers` message.

mbaudis:
@david4096 #761 is probably fine? However, and I actually don't want to push this, but: What is an `OntologyTerm` other than an external identifier? For which you need a service, if you want to get it resolved. By CURIE.

Just saying: ExternalIdentifier & OntologyTerm seem basically the same - or?

david4096:
@mbaudis Yep! However, I don't expect all these other databases to present themselves as ontologies. For that reason we give a message that better suits the use case.

mbaudis:
Merging into the metadata integration branch

PR #816, opened 2017-02-14
Adds deep set/get attr to protocol module

andrewjesaitis:
Adds a pair of deep get/set attr methods to protocol to support PR #1562 on the server.


david4096:
Great! https://github.com/ga4gh/server/pull/1562 And thanks @dcolligan for explaining how to carry this out.

PR #817, opened 2017-02-15
Prevent erroneous constraints.txt file checkins

dcolligan:


python/constraints.txt.default
```diff
@@ -0,0 +1,26 @@
...
+#
+# When one has proposed a common modification, they can change this
+# constraint to point at the branch of their proposed change:
+#
+# git+git://github.com/david4096/ga4gh-common.git@some-branch#egg=ga4gh_common
+#
+# Then, when the change has been accepted at the common repository
+# you can change this to point back at the master branch!
+#
+# When making a release, these constraints are changed to point at
```
david4096: Noting that we specifically opted not to do this at the release. Instead the constraints line is deleted and the pinned version is set in the requirements.txt.

dcolligan: Something to circle back to when we change our documentation (all of the constraints.txt files should have similar instructions... I just copy and pasted them)

david4096:
Thanks @dcolligan

PR #818, opened 2017-02-17
remove feature_id from ExpressionLevel

saupchurch:
removes feature_id from ExpressionLevel and changes search to filter by name


ejacox:
Thanks @saupchurch.
+1

david4096:
Thanks @saupchurch +1. This change satisfies the same access patterns of resolving a feature and then searching the RNA expression database, except that one does not need to have a fully formed `Feature` record present to access RNA quantification data properly.

Searching and indexing expression levels by name is a more tractable and lower ETL cost than resolving specific feature IDs in a sequence annotation database.

kozbo:
closes #829

PR #826, opened 2017-02-24
1491 vcf filters

ejacox:
Updated #777


david4096:
Thanks @ejacox carrying over my +1 from there.

Issue #833, opened 2017-02-28
Rename repository


dcolligan:
For consistency with other repos, rename to `ga4gh-schemas`

Spawned from https://github.com/ga4gh/server/issues/1585

david4096:
+1

dcolligan:
Fixed in #834 (although somehow most changes got merged earlier)

PR #834, opened 2017-02-28
Prepare for repository rename to ga4gh-schemas

dcolligan:
Issue #833


david4096:
+1 Thanks for doing this. You'll note that the maven artifact using the name `ga4gh-schemas`, so this makes our offering more sensible overall.

https://github.com/ga4gh/schemas/blob/master/pom.xml#L5

PR #836, opened 2017-03-02
refactoring BioCharacteristics

mbaudis:
Following discussions during a meeting at UCSC, this PR changes the BioCharacteristics object structure by removing one object level and changing assignment to phenotypes, diseases ... to a label.


mbaudis:
@david4096 @kozbo Please check - this addresses the structural simplification for the ontology wrappers.

src/main/proto/ga4gh/bio_metadata.proto
```diff
@@ -134,7 +140,7 @@ message BioCharacteristic {
   //     termId: "UBERON:0010033",
   //     term: "posterior part of tongue",
   //
-  repeated OntologyTerm ontologyTerms = 2;
+  repeated OntologyTerm ontologyTerms = 3;
```
kozbo: We shouldn't renumber pre-existing fields in protobuf. Add a new number instead.

src/main/proto/ga4gh/bio_metadata.proto
```diff
@@ -155,19 +161,6 @@ message BioCharacteristic {
...
-// with a first class BioMetadata object (Individual, BioSample).
-message Characteristics {
-  // Diseases associated with the parent BioMetadata object (Individual,
-  // BioSample).
-  repeated BioCharacteristic diseases = 1;
-
-  // Phenotypes associated with the parent BioMetadata object (Individual,
-  // BioSample).
-  repeated BioCharacteristic phenotypes = 2;
+  repeated OntologyTerm negatedOntologyTerms = 4;
```
kozbo: this looks like a simple re-organization of existing fields, so we can keep the original protobuf number of 3

mbaudis:
@kozbo I know; but this is still in a branch & has never seen master => ?!

src/main/proto/ga4gh/bio_metadata.proto
mbaudis: Can you just adjust it to what you see appropriate? Thanks!

mcourtot: Small comment on 'characteristics_class': I would avoid 'class' and use maybe 'scope' to match the documentation. I would also suggest to replace 'typical examples could be' with a stronger requirement, such as a list of acceptable values (maybe with a definition, for example in the above I'm not sure of the difference between a phenotype and an observation - does it matter?)

mbaudis:
@mcourtot Thanks for the comment; I did change the attribute name before, now to `scope`.

mbaudis:
Consolidating this into metadata-integration as the staging branch.

PR #848, opened 2017-03-08
Peer service improvements

david4096:
~~Move the `Peer` message to its own file.~~

Add a `remote_address` field to the GetInfoResponse which can be used to reflect on the origin of a request. This makes it easier to make well formed announce messages in mixed network environments.


kozbo:
I discussed this with @ejacox and I now think you were right to keep the messages together with the data. The peer service has nothing to do with the representation of genomic data. It is associated with serving the data and as such should be represented as a protocol only. 

@david4096 : you responded to my questions very quickly with this PR. Thanks for your quick reaction :-) Let's go with your first design and close this PR without merging.  So, -1 with apologies, since I suggested it in the first place!

david4096:
No problem, updated!

david4096:
This is intended for a future release, we haven't implemented the remote_address functionality yet.

kozbo:
thanks @david4096

PR #858, opened 2017-03-10
Update INSTALL.rst

jpdna:
file name of `python/dev-requirements.txt` is currently incorrect, fixed here


david4096:
Thanks very much!

PR #864, opened 2017-03-14
geodata object and implementations

mbaudis:
This is a go at the previously discussed geodata object. See e.g.
* Geodata object issue https://github.com/ga4gh/ga4gh-schemas/issues/771
* Google doc with current notes https://docs.google.com/document/d/1PBIlNtxJAXe2K1wwTlPMNO4N0e03uejl8Da14RtEjJg/edit#
* "old" metadata dev notes https://github.com/ga4gh/ga4gh-schemas/blob/metadata/src/main/resources/avro/metadata.avdl#L47

### DIFF
--- a/src/main/proto/ga4gh/bio_metadata.proto
+++ b/src/main/proto/ga4gh/bio_metadata.proto
@@ -48,6 +48,9 @@ message Individual {
   // Recommended: PATO http://purl.obolibrary.org/obo/PATO_0020001; PATO_0020002
   OntologyTerm sex = 9;
 
+  // The address coded as geolocationplace where the biosample was collected.
+  GeoLocation location = 12;
+
   // A map of additional information regarding the Individual.
   Attributes attributes = 10;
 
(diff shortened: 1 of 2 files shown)


helenp:
Summarising here from the notes document: 
"As a pragmatic approach, the combination of lat,long + a (geonames.org mapped) location name seems the best compromise, with the addition of a “precision level” providing additional features (e.g. possibility to randomize point locations in a given boundary)."

helenp:
+1

mbaudis:
@ejacox Thanks - great! Now, (also @kozbo, @david4096), how/when can we get this as a pick -> push_to_master? IMHO any time ...

ejacox:
I didn't realize you were working on a branch of master. I thought we were only forking from master (is that right @kozbo, @david4096?). Anyways, I will let @kozbo merge the branch since that will signify that we are no longer keeping the reference server in sync with the schemas.

mbaudis:
The `metadata-integration` is thought as a staging branch for master. We had discussed this in Vancouver as a practical way to collect major "metadata" code updates, which however all are intended to be merged to master at some point. It is up to the schema integration team to select what & when is merged ... 

So for me, changes in the `metadata-integration` are "forward looking" and can be used as basis for to add features to our test implementations.

ejacox:
Thank you for the clarification @mbauds.

PR #879, opened 2017-06-29
Metadata geoobjects

mbaudis:


mbaudis:
Small changes to the Geoobject (attributes ...).

PR #884, opened 2017-10-03
Deprecation notice added

rishidev:
It was decided at a GA4GH meeting to deprecate future work in this repository, and make it clear to users that this is the case.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,6 +2,10 @@
 	    :alt: Image
 
 
+THIS REPOSITORY HAS BEEN DEPRECATED FOR NEW WORK
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+Due to a GA4GH move to produce smaller, tightly defined, standards in smaller bites, this repository is no longer suitable as a place for new standards to be added. If further work is required on an existing schema defined within this repository the recommendation is to extract that schema from within this repo and work on it in a separate space. Please see the GA4GH website <https://ga4gh.org> for further details on the GA4GH Connect operational structure (which includes details on Work Streams), lists of existing standards and contact details should you have further queries. 
+
 Schemas for the Data Working Group
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 


heuermh:
More explanation needed, please.

mcourtot:
Hi @heuermh; following the GA4GH plenary meeting in Orlando the project has been reorganised into work streams and driver projects, which supersede & replace the task teams. Being part of the 'old' metadata task team I know that we will still produce a data model, but I don't think the reference implementation will be maintained for example. Also as you can see above the current build in this repository is failing. The new GA4GH website is at https://www.ga4gh.org and hopefully provides more info.

akmorrow13:
I find it interesting the htsget also implements its own schemas based on BAM fields, when it seems this project would have been a great example for the use of ga4gh-schemas. Any thoughts on why they developed new schemas? This is proof that there is still a need for these.

lh3:
Github has recently implemented a new feature to [archive a repo and make it read-only](https://github.com/blog/2460-archiving-repositories). Perhaps it is time to use the new feature here?

heuermh:
> Perhaps it is time to use the new feature here?

No, I don't believe that would be appropriate.  Even though GA4GH the organization has deprecated future work in this repository, that should not prevent the greater community from doing so (Github reports 117 forks currently).

lh3:
Archiving only makes this repo read-only, not the forks. In addition, you can still fork archived repos.

heuermh:
I don't see anything wrong with adding a deprecation notice and referencing new projects or efforts.

Archiving the repository however is an act of aggression against the community -- what is wrong with continuing to allow issues to be created and discussed?  pull requests to be considered and merged?

If the current build is in a sorry state, perhaps GA4GH members who did not attend the plenary meeting and weren't part of the deprecation conversation or even non-GA4GH contributors might be willing to help make it right?

https://www.youtube.com/watch?v=OT5mqWKirhI

mbaudis:
@heuermh I'm neutral on locking vs. plastering deprecation notes all over the repo.

However, regarding the "pick up & make it work": The problem wasn't the ability/willingness of people to contribute, but rather the existence of independent fiefdoms with different levels of activity (i.e. `~0` for variants after first definitions...), but then interdependencies. I cannot easily convey my frustrating experiences, seeing code after-monthlong-discussions-and-acceptance getting reverted later, due to nobody moving "essential" server code forward.

I had made it explicit that I see the schema fulfilling the role of "glue" keeping different projects/parts of GA4GH together.

However, without an actively guiding hand managing code advancements, I have learned to appreciate the modular approach, where different areas of the schema are managed quasi independently, but with cross-participation. So if there are time stamps in Un*x time somewhere - whatever.

Therefore: I wish for a consistent GA4GH schema _in spirit_, but see the current "disassemble but talk to each other" approach as a way forward. And sure, with some, limited, effort, one can easily build a coherent + working + better schema than the remnants you see; but it is about uptake through participation (which makes this option a no-go).

Just IMO...

